PR #5, opened 2018-12-16
Minimal VCS backend

Jbalkind:
Not sure that all the arguments work but this backend worked for my simple cases with openpiton.

Happy to make edits versus what's here. Also it's basically a direct copy of one of the other backends (I forget which) so I may have left something in there accidentally.

### DIFF
--- a/edalize/vcs.py
+++ b/edalize/vcs.py
@@ -1,0 +1,86 @@
+import os
+import logging
+
+from edalize.edatool import Edatool
+
+logger = logging.getLogger(__name__)
+
+class Vcs(Edatool):
+
+    tool_options = {
+        'lists' : {'vcs_options' : 'String'}
+    }
+
+    argtypes = ['plusarg', 'vlogdefine', 'vlogparam']
+
+    MAKEFILE_TEMPLATE = """
+all: $(TARGET)
+
+$(TARGET):
+	vcs -full64 -t $(TOPLEVEL) -f $(TARGET).scr -o $@ $(VCS_OPTIONS)
+
+run: $(TARGET)
+	$(TARGET) -l vcs.log $(EXTRA_OPTIONS)
+
+clean:
+	$(RM) $(TARGET)
+"""
+
+    def configure_main(self):
+        f = open(os.path.join(self.work_root, self.name+'.scr'),'w')
+
+        (src_files, incdirs) = self._get_fileset_files()
+        for key, value in self.vlogdefine.items():
+            f.write('+define+{}={}\n'.format(key, self._param_value_str(value, '')))
+
+        for key, value in self.vlogparam.items():
+            f.write('+parameter+{}.{}={}\n'.format(self.toplevel, key, self._param_value_str(value, '"')))
+        for id in incdirs:
+            f.write("+incdir+" + id+'\n')
+        timescale = self.tool_options.get('timescale')
+        if timescale:
+            with open(os.path.join(self.work_root, 'timescale.v'), 'w') as tsfile:
+                tsfile.write("`timescale {}\n".format(timescale))
+            f.write('timescale.v\n')
+        for src_file in src_files:
+            if src_file.file_type in ["verilogSource",
+		                      "verilogSource-95",
+		                      "verilogSource-2001",
+		                      "verilogSource-2005",
+                                      "systemVerilogSource",
+			              "systemVerilogSource-3.0",
+			              "systemVerilogSource-3.1",
+			              "systemVerilogSource-3.1a"]:
+                f.write(src_file.name+'\n')
+            elif src_file.file_type == 'user':
+                pass
+            else:
+                _s = "{} has unknown file type '{}'"
+                logger.warning(_s.format(src_file.name, src_file.file_type))
+
+        f.close()
+
+        with open(os.path.join(self.work_root, 'Makefile'), 'w') as f:
+
+            f.write("TARGET           := {}\n".format(self.name))
+            f.write("TOPLEVEL         := {}\n".format(self.toplevel))
+            f.write("VCS_OPTIONS := {}\n".format(' '.join(self.tool_options.get('vcs_options', []))))
+            if self.plusarg:
+                plusargs = []
+                for key, value in self.plusarg.items():
+                    plusargs += ['+{}={}'.format(key, self._param_value_str(value))]
+                f.write("EXTRA_OPTIONS    ?= {}\n".format(' '.join(plusargs)))
+
+            f.write(self.MAKEFILE_TEMPLATE)
+
+    def run_main(self):
+        args = ['run']
+
+        # Set plusargs
+        if self.plusarg:
+            plusargs = []
+            for key, value in self.plusarg.items():
+                plusargs += ['+{}={}'.format(key, self._param_value_str(value))]
+            args.append('EXTRA_OPTIONS='+' '.join(plusargs))
+
+        self._run_tool('make', args)
--- a/edalize/vcs.py
+++ b/edalize/vcs.py
@@ -13,19 +13,6 @@ class Vcs(Edatool):
 
     argtypes = ['plusarg', 'vlogdefine', 'vlogparam']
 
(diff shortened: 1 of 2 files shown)


edalize/vcs.py
```diff
@@ -0,0 +1,86 @@
...
+        'lists' : {'vcs_options' : 'String'}
+    }
+
+    argtypes = ['plusarg', 'vlogdefine', 'vlogparam']
+
+    MAKEFILE_TEMPLATE = """
+all: $(TARGET)
+
+$(TARGET):
+	vcs -full64 -t $(TOPLEVEL) -f $(TARGET).scr -o $@ $(VCS_OPTIONS)
```
olofk: I don't know myself around vcs. Is full64 something we would always like to have set, or should that go into VCS_OPTIONS? Or is it something that has to do with 32/64-bit hosts and should be autodetected?

edalize/vcs.py
```diff
@@ -0,0 +1,86 @@
...
+
+        (src_files, incdirs) = self._get_fileset_files()
+        for key, value in self.vlogdefine.items():
+            f.write('+define+{}={}\n'.format(key, self._param_value_str(value, '')))
+
+        for key, value in self.vlogparam.items():
+            f.write('+parameter+{}.{}={}\n'.format(self.toplevel, key, self._param_value_str(value, '"')))
+        for id in incdirs:
+            f.write("+incdir+" + id+'\n')
+        timescale = self.tool_options.get('timescale')
```
olofk: I presume this was copied from the icarus backend. The `timescale` option was only added here as Icarus doesn't have a way to set the default timescale. If vcs has a way, I'd prefer to use that instead and let users put it in vcs_options

edalize/vcs.py
```diff
@@ -0,0 +1,86 @@
...
+        if timescale:
+            with open(os.path.join(self.work_root, 'timescale.v'), 'w') as tsfile:
+                tsfile.write("`timescale {}\n".format(timescale))
+            f.write('timescale.v\n')
+        for src_file in src_files:
+            if src_file.file_type in ["verilogSource",
```
olofk: I guess vcs support VHDL as well, but if you don't have any VHDL code to test with, this can be added later upon request

edalize/vcs.py
```diff
@@ -0,0 +1,86 @@
+import os
+import logging
+
+from edalize.edatool import Edatool
+
+logger = logging.getLogger(__name__)
+
+class Vcs(Edatool):
```
imphil: A bit of API doc would be nice, ideally outlining the supported/tested version(s) as well.

edalize/vcs.py
```diff
@@ -0,0 +1,86 @@
...
+
+$(TARGET):
+	vcs -full64 -t $(TOPLEVEL) -f $(TARGET).scr -o $@ $(VCS_OPTIONS)
+
+run: $(TARGET)
+	$(TARGET) -l vcs.log $(EXTRA_OPTIONS)
+
+clean:
+	$(RM) $(TARGET)
+"""
```
imphil: I would love if no more inline Makefiles are introduced and templates would be used instead. The Spyglass backend is one of the most recent ones and could be a good starting point.

edalize/vcs.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def configure_main(self):
+        f = open(os.path.join(self.work_root, self.name+'.scr'),'w')
+
+        (src_files, incdirs) = self._get_fileset_files()
+        for key, value in self.vlogdefine.items():
+            f.write('+define+{}={}\n'.format(key, self._param_value_str(value, '')))
+
+        for key, value in self.vlogparam.items():
+            f.write('+parameter+{}.{}={}\n'.format(self.toplevel, key, self._param_value_str(value, '"')))
```
imphil: isn't `self.toplevel` a list? At least `Edatool::__init__()` uses it that way. (Yes, the name is and was confusingly singular.) [Speaking about that, I see that I've used it as string in the spyglass backend as well... I guess we need a couple more tests here.]

edalize/vcs.py
```diff
@@ -0,0 +1,86 @@
...
+            f.write('timescale.v\n')
+        for src_file in src_files:
+            if src_file.file_type in ["verilogSource",
+		                      "verilogSource-95",
+		                      "verilogSource-2001",
+		                      "verilogSource-2005",
+                                      "systemVerilogSource",
+			              "systemVerilogSource-3.0",
+			              "systemVerilogSource-3.1",
+			              "systemVerilogSource-3.1a"]:
```
imphil: how about using `startsWith()` here?

edalize/vcs.py
imphil:
```
-full64
   Compiles the design in 64 bit mode and creates a 64 bit executable
   for simulating in 64 bit mode.
```

So it's required on 64 bit platforms (otherwise VCS won't start). Given that I cannot imagine anybody running VCS on a 32 bit machine these days it's probably safe to unconditionally set it.

Detecting 32 vs 64 bit would probably be sufficiently easy, however testing for that is probably annoying ...

edalize/vcs.py
```diff
@@ -0,0 +1,86 @@
...
+    tool_options = {
+        'lists' : {'vcs_options' : 'String'}
+    }
+
+    argtypes = ['plusarg', 'vlogdefine', 'vlogparam']
+
+    MAKEFILE_TEMPLATE = """
+all: $(TARGET)
+
+$(TARGET):
```
imphil: This make target should depend on the scr file.

edalize/vcs.py
Jbalkind: The rub is that we've always seen vcs fail on 64 bit machines if you don't give the -full64 flag. We could check something but I'm not sure what the best thing to check is?

Jbalkind:
Several of these comments reference details about the code that I don't know about. I just wrote this as a copy of icarus.py and removed the parts I knew didn't work the same for vcs (hence why I didn't add vhdl - we don't write/use any vhdl). If this backend should do something different/have different comments etc, then icarus should too, right? Are you wanting me to provide those changes for icarus.py too?

edalize/vcs.py
Jbalkind: VCS has a `-timescale=unit/precision` flag - just made commit 59e4159 to kill this bit

edalize/vcs.py
Jbalkind: Switched to using a template instead! Would be good if you could give it a look over as I've never used jinja2.

edalize/vcs.py
Jbalkind: Now I look at it, I'm not even sure how you tell vcs which your top-level module is. In my other use I've just seen it tell you it has 9 (true) top-modules and then simulates all of them. I can't even find an argument in the help page to indicate and it seems to just silently accept -t (other people on the internet have mentioned -top, which also is not in the manual)

edalize/vcs.py
Jbalkind: Hoping that 3e1cfe1 did this the way you wanted

edalize/vcs.py
olofk: Cool. Thanks

edalize/vcs.py
olofk: My patience with jinja2 is growing thin, and also, it's likely that the makefiles will be replaced by platform-specific .bat/.sh files in the near future, so I don't think we should spend too much time on this. If the template works, let's keep it. Otherwise revert to the inline makefile

olofk:
I'm happy with the changes. Just need to register the new template in setup.py. Otherwise it won't be available when installing with pip. @Jbalkind, I can amend that locally if you want to save yourself from adding another patch

Future improvements would be
- VHDL support
- Test case for the backend
- VPI/DPI support
- 32/64-bit detection

But I'm fine with bringing this in later. I'm about to do a new release and would like to have this backend included

Jbalkind:
Please feel free, if you're happy with it!

olofk:
Thanks! I added the template and squashed into a single commit. Picked and pushed!

PR #9, opened 2019-01-05
Add VHDL generics support for quartus and vivado.

benreynwar:


olofk:
Thanks! Picked and pushed
I did a small local addition to change some of the newlines to make the testa pass in the references files since jinja2 is as always completely incapable of making anything sensible when it comes to newlines

PR #10, opened 2019-01-06
ise: Make Project generator TCL file idempotent.

cr1901:
Migen in general creates a build tree that's idempotent (you can rerun the top level script- in this case, the Makefile _when `-B` is specified_- multiple times on the same set of files, and the behavior between runs doesn't change). It's possible/common for Migen to regenerate Verilog or build files without the Verilog contents actually changing (you added a new Verilog file) or build files actually changing (you updated the Verilog but left the build files alone).

The TCL API of ISE has a few quirks such that if you try to create an already existing project or add an existing UCF, `xtclsh` will fail and refuse to continue. 

Therefore I created two TCL functions to work around this; if the files already exist, the build will continue. _Any file (property) that the TCL API can tolerate adding multiple times (setting multiple times, to different values even) is left alone._


olofk:
Thanks! This is a welcome addition. Picked and pushed

FYI, I took the liberty to update the ISE TCL reference files so that tests passes again

PR #13, opened 2019-01-28
Document Spyglass backend

imphil:
Add documentation for the Spyglass backend.

Side remark: You get much nicer tables in the source code with the rstFlatTable extension. 
- https://github.com/opensocdebug/osd-doc/blob/master/src/02_spec/03_dataformats.rst#debug-packet-structure
- https://github.com/opensocdebug/osd-doc/blob/master/sphinx/rstFlatTable.py (it's from https://github.com/torvalds/linux/blob/master/Documentation/sphinx/rstFlatTable.py)


olofk:
Thanks! Picked and pushed

PR #15, opened 2019-03-10
Actually pass vcs_options to vcs

Jbalkind:
Turns out I wasn't passing the right thing to VCS. If you added vcs_options, the vcs compile string would just include the string "vcs_options" instead of what you actually wanted to pass.


olofk:
Thanks! Picked and pushed!

PR #17, opened 2019-03-14
Correct ModelSim VHDL generic handling

GCHQDeveloper560:
Currently attempts to set the value of a generic for ModelSim give me the following:
```
# vsim -do "run -all" -c boolean_generic -gbits=11 -gaddtwo=1
...
# ** Warning: (vsim-3351) Invalid value "1" for generic "addtwo".  Not using this value.
```
It seems ModelSim wants the strings `true` and `false` for VHDL generics. After the changes in this PR I get the following with no warning:
```
# vsim -do "run -all" -c boolean_generic -gbits=11 -gaddtwo=True
```

I believe Quartus 16 and older have the same behaviour, so a similar solution is likely required there.

### DIFF
--- a/edalize/edatool.py
+++ b/edalize/edatool.py
@@ -76,7 +76,8 @@ class Edatool(object):
             trim_blocks = True,
             lstrip_blocks = True,
         )
-        self.jinja_env.filters['param_value_str'] = jinja_filter_param_value_str
+        self.jinja_env.filters['param_value_str']   = jinja_filter_param_value_str
+        self.jinja_env.filters['generic_value_str'] = jinja_filter_param_value_str
 
     @classmethod
     def get_doc(cls, api_ver):
--- a/edalize/edatool.py
+++ b/edalize/edatool.py
@@ -9,22 +9,23 @@ from jinja2 import Environment, PackageLoader
 logger = logging.getLogger(__name__)
 
 # Jinja2 tests and filters, available in all templates
-def jinja_filter_param_value_str(value, str_quote_style=""):
+def jinja_filter_param_value_str(value, str_quote_style="", bool_is_str=False):
     """ Convert a parameter value to string suitable to be passed to an EDA tool
 
     Rules:
-    - Booleans are represented as 0/1
+    - Booleans are represented as 0/1 or "true"/"false" depending on the
+      bool_is_str argument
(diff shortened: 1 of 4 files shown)


GCHQDeveloper560:
I've added commits to handle the similar Quartus issue.

GCHQDeveloper560:
The Quartus check may need to be a little more complex. I need to do more testing, but it seems Quartus 18.1 Standard wants boolean generics to be strings. So the check may be older versions of Pro and all versions of Standard. I'll try and check the Lite edition as well, though I assume it behaves like Standard.

GCHQDeveloper560:
This works for the Quartus versions I can access. It seems newer versions of Quartus Pro are the only ones that support 1/0 for true/false VHDL boolean generics. Other editions and Quartus Pro < 17 need the strings "true" and "false"

olofk:
Thanks for the patches and for the extensive testing. I will hopefully be able to review this work soon

olofk:
ok, this looks great even though I haven't actually tested it yet. Just a couple of questions.
1. Does the code still pass py.test ?
2. Do we really need the version check, or could we just use true/false for all quartus versions to make it simpler? (I'm fine with keeping it the way it is, but would be good to know if there are any problem reports later on)
3. Ok to squash `Correct Quartus versions needing strings for boolean generics` into the previous patch? (I can do that locally)

Thanks!

GCHQDeveloper560:
1. The branch passed pytest, but it looks like the latest two commits to master are causing failures.
2. Hopefully I understand your question. We need the version check as Quartus Pro 17 and later require 1/0, while the older versions need the string. They completely changed the behaviour.
3. Feel free to squash anything you like!

olofk:
>     1. The branch passed pytest, but it looks like the latest two commits to master are causing failures.
> 

Great. The failures are probably unrelated and comes from the amazingly poor dependency handling of Python packages. I will try to fix that independently

>     2. Hopefully I understand your question. We need the version check as Quartus Pro 17 and later require 1/0, while the older versions need the string. They completely changed the behaviour.

Wow! That just sounds like a terribly bad decision from Intel's side. Looks like you did the right thing here then

> 
>     3. Feel free to squash anything you like!

Great. Thanks

olofk:
All good. Picked and pushed. Thanks!

PR #22, opened 2019-04-08
Fix Qsys filename in quartus-std-makefile template

wgoossens:
When trying to build the sockit design I encountered an error in the Makefile template. This should fix it.

Signed-off-by: Walter Goossens <waltergoossens@creative-embedded.com>


olofk:
I'll admit that I didn't do a ton of testing for this patch but it does seem like the right thing to do.

I also took the liberty to add in the corresponding change to the reference makefile so that the test suite passes. Hope that's fine with you (remember to run py.test after making changes)

Picked and pushed. Thanks!

PR #29, opened 2019-05-12
First attempt at an incisive backend

Jbalkind:
I mainly copy and pasted from the VCS backend. Happy to modify however is best. It seems to run my simple tests at least.

### DIFF
--- a/edalize/incisive.py
+++ b/edalize/incisive.py
@@ -59,6 +59,5 @@ Example snippet of a CAPI2 description file for Incisive:
             plusargs = []
             for key, value in self.plusarg.items():
                 plusargs += ['+{}={}'.format(key, self._param_value_str(value))]
-            args.append('EXTRA_OPTIONS='+' '.join(plusargs))
 
         self._run_tool('make', args)
--- a/edalize/incisive.py
+++ b/edalize/incisive.py
@@ -1,0 +1,64 @@
+import os
+import logging
+
+from edalize.edatool import Edatool
+
+logger = logging.getLogger(__name__)
+
+
+class Incisive(Edatool):
+
+    _description = """ Cadence Incisive Verilog Backend
+
+Example snippet of a CAPI2 description file for Incisive:
+
+.. code:: yaml
+
+   incisive:
+     incisive_options:
+       # Compile-time options passed to the incisive command
+       - -some_buildarg
+     run_options:
+       # Run-time options passed to the simulation itself
+       - +some+plusarg
+"""
+
+    tool_options = {
+        'lists' : {
+            'incisive_options' : 'String', # compile-time options (passed to Incisive)
+            'run_options' : 'String', # runtime options (passed to simulation)
+        }
+    }
+
+    argtypes = ['plusarg', 'vlogdefine', 'vlogparam']
+
+    def configure_main(self):
+        self._write_fileset_to_f_file(os.path.join(self.work_root, self.name + '.scr'),
+                                      include_vlogparams = True)
+
+        plusargs = []
+        if self.plusarg:
+            for key, value in self.plusarg.items():
+                plusargs += ['+{}={}'.format(key, self._param_value_str(value))]
+
+        template_vars = {
+            'name'              : self.name,
+            'incisive_options'  : self.tool_options.get('incisive_options', []),
+            'run_options'       : self.tool_options.get('run_options', []),
+            'toplevel'          : self.toplevel,
+            'plusargs'          : plusargs
+        }
+
+        self.render_template('Makefile.j2', os.path.join(self.work_root, 'Makefile'), template_vars)
+
+    def run_main(self):
+        args = ['run']
+
+        # Set plusargs
+        if self.plusarg:
+            plusargs = []
+            for key, value in self.plusarg.items():
+                plusargs += ['+{}={}'.format(key, self._param_value_str(value))]
+            args.append('EXTRA_OPTIONS='+' '.join(plusargs))
+
+        self._run_tool('make', args)
--- a/edalize/templates/incisive/Makefile.j2
+++ b/edalize/templates/incisive/Makefile.j2
@@ -1,0 +1,10 @@
+all: {{ name }}
+
+{{ name }}: {{ name }}.scr
+	ncverilog -C -f {{ name }}.scr +nctop+{{ toplevel }} {% for option in incisive_options %} {{ option }}{% endfor %}
+
+run: {{ name }}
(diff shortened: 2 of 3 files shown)


edalize/incisive.py
```diff
@@ -0,0 +1,64 @@
...
+
+.. code:: yaml
+
+   incisive:
+     incisive_options:
+       # Compile-time options passed to the incisive command
+       - -some_buildarg
+     run_options:
+       # Run-time options passed to the simulation itself
+       - +some+plusarg
```
olofk: Is plusarg really a good example here? Shouldn't they rather be handled with parameters of type plusarg instead? Not a dealbreaker but could avoid some confusion

edalize/incisive.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+
+    tool_options = {
```
olofk: tool_options and _description should be replaced with a get_doc function. I have missed that in my review of the spyglass, vcs and ascentlint backends but you could see any of the other backends. It should be pretty self-explanatory.

edalize/templates/incisive/Makefile.j2
```diff
@@ -0,0 +1,10 @@
+all: {{ name }}
+
+{{ name }}: {{ name }}.scr
+	ncverilog -C -f {{ name }}.scr +nctop+{{ toplevel }} {% for option in incisive_options %} {{ option }}{% endfor %}
+
+run: {{ name }}
+	ncverilog +nclibdirname+INCA_libs -r worklib.{{ toplevel }}:v {% for plusarg in plusargs %} {{ plusarg }} {% endfor %}{% for option in run_options %} {{ option }}{% endfor %}
```
olofk: I'm always sceptic towards hard-coded options as they tend to be the backend author's preference in 99% of the cases rather than something really required. Is this INCA_libs thingamob a mandatory option? What does it do?

setup.py
```diff
@@ -14,6 +14,7 @@ def read(fname):
         'templates/spyglass/spyglass-project.prj.j2',
         'templates/spyglass/spyglass-run-goal.tcl.j2',
         'templates/vcs/Makefile.j2',
+        'templates/incisive/Makefile.j2',
```
olofk: Bonus points for remembering this. No one ever does :)

olofk (changes requested):
I'm being a bit more harsh than I have been before to save myself from rewriting a bunch of code later on. I also very much would like to see tests added. But other than that I say it looks very good and I'm looking forward to merge it with the review points fixed

edalize/incisive.py
```diff
@@ -0,0 +1,64 @@
...
+import logging
+
+from edalize.edatool import Edatool
+
+logger = logging.getLogger(__name__)
+
+
+class Incisive(Edatool):
+
+    _description = """ Cadence Incisive Verilog Backend
```
imphil: The "Verilog" isn't part of the product name any more I think, it's "Incisive Enterprise Simulator" these days.

Jbalkind: Ah this was meant to be read as "(Cadence Incisive) (Verilog Backend)" to indicate it's only been tested with verilog and not SV or VHDL

edalize/templates/incisive/Makefile.j2
imphil: What's `:v` after the toplevel name? Is `worklib` required? Do you need to prefix parameters with the library name as well?

Jbalkind: My understanding is that this is required in the same way modelsim has that work directory?

imphil (commented):
Thanks @Jbalkind! As Olof said, please add tests as well, every time we merged something without tests, it was broken in one way or the other. (And every time I tried to fix something without running the tests, I broke things as well.)

edalize/incisive.py
Jbalkind: Yeah I'll read some more of the docs here to figure out an actual valid runarg

edalize/incisive.py
```diff
@@ -0,0 +1,64 @@
...
+       # Compile-time options passed to the incisive command
+       - -some_buildarg
+     run_options:
+       # Run-time options passed to the simulation itself
+       - +some+plusarg
+"""
+
+    tool_options = {
+        'lists' : {
+            'incisive_options' : 'String', # compile-time options (passed to Incisive)
```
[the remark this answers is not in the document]
Jbalkind: Should I change the name of the backend too or are you ok with incisive?

edalize/incisive.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def run_main(self):
+        args = ['run']
+
+        # Set plusargs
+        if self.plusarg:
+            plusargs = []
+            for key, value in self.plusarg.items():
+                plusargs += ['+{}={}'.format(key, self._param_value_str(value))]
+            args.append('EXTRA_OPTIONS='+' '.join(plusargs))
```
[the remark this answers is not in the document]
Jbalkind: Got it!

edalize/incisive.py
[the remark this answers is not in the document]
imphil: "ncsim" is the most commonly used name I've heard, and Wikipedia agrees (https://en.wikipedia.org/wiki/NCSim). So I'd use that as name.

edalize/templates/incisive/Makefile.j2
cmarqu: I think the ``nclibdirname`` is not required, and ``INCA_libs`` is the default anyway. Also, it's usually written like ``-nclibdirname INCA_libs``. I basically never use plusargs with Incisive.

cmarqu:
Maybe https://github.com/VUnit/vunit/blob/master/vunit/incisive_interface.py can be a source of inspiration as well. It uses the ``irun`` executable and can e.g. hand defines/generics to it (via ``-define`` and ``-gpg`` respectively).

olofk:
Should we drop this one now that there's a xcelium backend?

Jbalkind:
I thought the xcelium backend didn't support incisive though?

olofk:
I would say that's a pretty good argument

From my (somewhat limited) understanding, these tools have a lot in common. Can we benefit from this somehow or are we just as well to have two independent backends?

Jbalkind:
To be clear, I'm not tied at all to this PR. I don't know the tool that well and just need the ability to run the simulations I can already run independently of fusesoc in order to port openpiton to fusesoc. This PR may not even work today and it could be that sharing the xcelium backend would be a better idea. But I remember that you previously commented that keeping them separate was fine (https://github.com/olofk/fusesoc/issues/272#issuecomment-494933700) so maybe we could just copy and adapt it? I don't know.

PR #32, opened 2019-05-22
VCS: Set SystemVerilog parsing flag when needed

imphil:
VCS needs a "-sverilog" flag when SystemVerilog sources are encountered.


olofk:
Makes sense. Have no way of testing but I definitely trust you on this

PR #36, opened 2019-06-27
Verilator: Support configuration files

imphil:
Verilator has support for configuration files, typically ending in
`*.vlt`. Read more about them at
https://www.veripool.org/projects/verilator/wiki/Manual-verilator#CONFIGURATION-FILES

This commit adds support to edalize to pass files with type 'waiver' as
configuration files (which is the most common use case for them), or
files with type 'user' and a file extension of '.vlt'.

Note that the current behavior of ignoring files with type 'user' is
preserved for all file extensions other than '.vlt'.

### DIFF
--- a/edalize/verilator.py
+++ b/edalize/verilator.py
@@ -114,8 +114,11 @@ class Verilator(Edatool):
                     f.write(src_file.name + '\n')
                 elif src_file.file_type in ['cppSource', 'systemCSource', 'cSource']:
                     opt_c_files.append(src_file.name)
-                elif src_file.file_type in ['user']:
+                elif src_file.file_type == 'vlt':
+                    f.write(src_file.name + '\n')
+                elif src_file.file_type == 'user':
                     pass
+
             f.write('--top-module {}\n'.format(self.toplevel))
             f.write('--exe\n')
             f.write('\n'.join(opt_c_files))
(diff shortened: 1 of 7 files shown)


olofk:
Supporting these files sounds like a good idea but I'm not happy with this implementation for two reasons

1. I think waiver is a poor choice of name. Even if that is the most common use, it also supports tracing and coverage. I would prefer to just call them `vlt`. Doesn't look like any other related file types claim this extension, so it should be ok.
2. I don't want to infringe on the user type. This type is used for files that the backends never should mess with. Instead, I'm increasingly warming up to the idea of suffix autodetection. `.v` files are in 99% of cases `verilogSource` and on the few occasions where they actually are `systemVerilogSource` we can set that explicitly. Not sure how hard it is, but I'm thinking that if the file is passed through EDAM without a file type then the backend will set one based on extension. We could start with vlt files.

So to sum it up, change file type from `waiver` to `vlt` and drop the stuff for the `user` type. Then I'm happy to accept

imphil:
1. `waiver` is the type used currently by the spyglass and ascentlint backends, that's why I chose to support that. It's more a convenience functionality for symmetry with other lint backends, which can be dropped if you don't like it. 

2. I choose `user` to avoid playing whack-a-mole with new file types. (There's no special processing involved here, it just gets passed on to Verilator.)

So I'd go for:
- Remove the `waiver` convenience functionality.
- Add `vlt` as file type. (Easiest solution for now.)
- Open a new issue to think about auto-detection of file types. (We probably need a bit more time to get that right.)

imphil:
Updated based on discussion.

imphil:
@olofk Could you have a look at this PR?

olofk:
Sorry for the delay. Picked, pushed and soon to be released now. Thanks!

PR #38, opened 2019-07-03
Add tclConstraint file_type for Vivado backend.

m-kru:
Vivado supports .tcl files as constraint files.
It is called unmanaged constraint file.

### DIFF
--- a/edalize/vivado.py
+++ b/edalize/vivado.py
@@ -100,6 +100,7 @@ class Vivado(Edatool):
             'xci'                 : 'read_ip',
             'xdc'                 : 'read_xdc',
             'tclSource'           : 'source',
+            'SDC'                 : 'read_xdc -unmanaged',
         }
         _file_type = f.file_type.split('-')[0]
         if _file_type in file_types:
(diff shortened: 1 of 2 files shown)


m-kru:
I know that it could be handled in different way by creating 2 tcl files, the first one as tclSource that would include tcl with constraints, but I think add tclConstraint file_type is better approach.

m-kru:
@olofk ping

olofk:
Sorry for the wait. Thinking about this, I'm leaning towards calling this `SDC` instead of `tclConstraint`, because that's basically what it is, and we already have SDC support in other backends so there's no need to introduce another file type. If you're ok with that I'm happy to accept a patch

m-kru:
That's fine for me, lets use SDC. All file types for Vivado backend are in lower case, is it ok to add SDC? Do you want me to update the patch or you will update it by yourself?

olofk:
Yeah, it's unfortunate that there is a mix between upper and lower case. When possible, I have based the list of file types on the ones found in the IP-XACT standard, and that one uses uppercase `SDC`. With that reasoning, the lowercase xdc and xci are "wrong", but I don't think it makes any sense to change now

I'm fine with updating the patch or let you do it. Whoever gets there first :)

m-kru:
I have updated. However, take a loot at that `No newline at end of file` change. I don't know how to prevent this, but it makes pytest to fail.

olofk:
Thanks. I fixed the newline issue and pushed now. Not sure what is going on with those newlines tbh. Think it's some kind of jinja2 thing but I find it impossible to understand how it works. A tip is to run with `GOLDEN_RUN=1 py.test`. That way, the test suite will update any differing files, and you can manually review the changes to make sure they are sane

Issue #39, opened 2019-07-17
GHDL support for generic argtype


GCHQDeveloper560:
It appears the GHDL backend expects VHDL generics to be supplied as `vlogparam` arguments from the days before support for the generic argument type was added (#7). Since GHDL doesn't support Verilog, should `vlogparam` be replaced with `generic`, or should both be supported for backward compatibility?

olofk:
Oh, that's a good question. My first thought would be to add support for generic and still handle vlogparam but print a warning that it will be removed eventually

PR #40, opened 2019-07-19
Add GHDL support for generic parameters and deprecate vlogparam support

GCHQDeveloper560:
Hopefully fixes olofk/edalize#39


olofk:
That looks fine to me. Thanks for fixing this!

PR #45, opened 2019-08-02
Remove use of undefined variable

GCHQDeveloper560:
The argument to this function was removed in commit 31eb06f, but this use of the
argument was left behind.


olofk:
Oops. Thanks for fixing. I'm amazed no one (including me) didn't notice before. Looks like the test suite could need some more love

PR #46, opened 2019-08-09
vivado: Do not exit at target open error

towoe:
Prevent the forwarding of an error generated by `open_hw_target` and
continue with possible further targets.

This can happen if multiple hardware targets exist but no devices are detected for a specific target.

### DIFF
--- a/edalize/templates/vivado/vivado-program.tcl.j2
+++ b/edalize/templates/vivado/vivado-program.tcl.j2
@@ -10,16 +10,19 @@ set found 0
 
 foreach { hw_target } [get_hw_targets] {
     current_hw_target $hw_target
-    open_hw_target
-    foreach { hw_device } [get_hw_devices] {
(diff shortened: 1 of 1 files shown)


towoe:
I tested this with Vivado 2018.3 and the Arty A7-100T and Nexys Video boards.

olofk:
Cool. Thanks.

Related to this I really wish someone could create a stand-alone library of FPGA programming tools so that I could extract this code from FuseSoC and collaborate with all other projects that also need to program FPGAs

1138-4EB:
> Related to this I really wish someone could create a stand-alone library of FPGA programming tools so that I could extract this code from FuseSoC and collaborate with all other projects that also need to program FPGAs

It might useful to explain this in the README. I.e., that edalize is focused on running synthesis, P&R and simulation, not on actually programming the boards.

PR #47, opened 2019-08-27
FileNotFoundError doesn't exist in Python 2

imphil:
Work around it in the same way we do in fusesoc: map FileNotFoundError
to OSError for Python 2.


imphil:
@tomroberts-lowrisc that should fix your edalize problem. You can try it out with

```
pip install -U git+https://github.com/imphil/edalize@fix-file-not-found-error-py2
```

olofk:
Thanks @imphil ! Picked and pushed

PR #52, opened 2019-09-21
Feature/add tool vunit

julian-becker:
I love both VUnit and FuseSoC/Edalize, and this is an early, opinionated attempt to integrate them more easily, by adding VUnit as a (minimally) supported tool in Edalize.

All feedback welcome! This is not yet very well tested, but I was thinking it might be a good idea to get feedback early, since I'm not yet very familiar with the codebase.

If you are aware of a better/alternative way to integrate vunit with FuseSoC, please let me know.

Regarding the tests: I added a corresponding test_vunit which passes on my machine (on Win10), however, the majority of the other tests are failing on my machine... Do they require some special environment setup?


olofk:
Fantastic! This looks very much like how I would have chosen to implement it. I would love to give this a proper review, but unfortunately with a deadline coming up and ORConf this weekend I will not have the time to look at it until next week.

julian-becker:
Great! Thanks for the initial feedback! There is no urgency, so take your time reviewing it when it is most convenient for you. In the meantime, I will probably do some more testing over the next couple of days and add a few more options etc when needed/convenient.

yottahawk:
Hey @julian-becker 
I've been playing with my own vunit tool for fusesoc/edalize - you beat me to the PR! haha
Your solution is very similar to mine, if you're interested : [yottahawk/edalize](https://github.com/yottahawk/edalize/commit/001f946f84a29e1b99c0ba1e7e0415727e1fb7f6)

One thing I'm still pondering is best how to allow the user full access to the vunit functionality in terms of the [python interface](https://vunit.github.io/python_interface.html), without duplicating the interface in terms of edalize tool options (obviously bad from a maintainence perspective). 
In my fork I experimented with tool options of pre / post flow python fragment files, which are inserted into the template when it's rendered. 'pre' would be for adding libraries/packages, and 'post' would be any of the options marked "Only affects \<x\> added before the option is set.". The implementation feels slightly clunky, but it does allow the user to access the full vunit interface. 

Any thoughts on that aspect?

julian-becker:
Hi @yottahawk 
Great to see there is more interest in this feature! And an alternative implementation is also great to have, so we can now pick and choose the best parts from both and merge them into one awesome new feature.

The question you bring up is one that I have been struggeling with as well. Indeed, there is a certain mismatch between the Edalize options and the VUnit-python interface.

I have not been using VUnit for very long, so I'm not very familiar with its advanced use-cases that require a more complicated Python setup or interfacing. My approach so far was to pragmatically add options as I encounter their necessity.

I would imagine that for a normal user, it is most convenient if no additional python files are necessary and everything is just generated from the Edalize-config. But I do agree, that we should not bar advanced users from using the full VUnit interface. 

The Python-Fragments for pre-/post-configuration as in your implementation could be a reasonable solution to support more advanced scenarios.
However, I'm a little bit concerned about the integration and development of those fragment files: If they are just copy-pasted as-is into the run.py, the fragment files themselves will probably be problematic for usual python-tooling like linters etc. which will e.g. be unable to find the definition for `vu` or to do any autocomplete.

Maybe instead of just textual replacement into the run.py file, the Python-VUnit-extensions could be properly imported, and the imported hooks could then be used to pre- and post-configure the global `vu`, as well as the libraries.
I'm thinking of something along these lines (in the `run.py`)

    from vunit_hooks import VUnitHooks

    hooks = VUnitHooks()

    # This hook allows creation and customization of the VUnit instance.
    # Allows access to full vunit.ui.VUnit interface:
    vu = hooks.create()
    
    for logical_name in libraries:
        lib = vu.add_library(logical_name)
        for file in src_files:
            lib.add_source_files(...)
        
        # This hook allows per-library customization of e.g. compile-flags
        # Allows full access to vunit.ui.Library interface
        hooks.handle_library(logical_name, lib)
    
    # Allow custom invokation of the VUnit instance, equipped with all libraries:
    hooks.main(vu)

yottahawk:
Yes I like your ideas here. The linting aspect is definitely something to consider. And probably plenty more QOL/workflow things that I can't even imagine! 
I definitely agree that a minimal run.py should be generated with no extra config. Certainly from my experience, that would cover a large number of implementations. Keep-It-Simple-Stupid is definitely worth us keeping in mind, as a low barrier to entry would be a massive asset.

If I understand your suggestion correctly, the user would populate eg. a VunitHooks template file/class ,which the run.py script would call into? It's definitely cleaner that the fragments idea, more pythonic for sure.
I was thinking it would be really nice to generate a run.py file that could be read and run on its own, possibly after halting at the edalize configure() stage. If you came to fusesoc knowing Vunit quite well beforehand, it would be much easier to understand if you could read a run.py file just like all the others out there. Plus it would be easier to debug any generation issues. Just from a readability/familiarity standpoint. 
It makes me think, one option could be taking the hooks idea and basically reverse it, allowing the user to provide the template file. Then some pre-defined metacomment e.g. <ADD_SRC_FILES_INSERTED_HERE> could be replaced with the jinja2 stuff, and then all the file/library adding stuff. It would make the file valid python on it's own, and very familiar to a user. Potentially even formulaic to modify existing run.py files if converting to fusesoc?

There's potentially a lot to consider the more I think about it. How would .core dependencies be included into a top-level vunit run system? Would all tests/tb's/configurations get joined into one large test suite? Could fusesoc generators do cool things here? Ok i'm probably getting ahead of myself :)

julian-becker:
Regarding the debuggability: Both of our current implementation already produce a run.py file which the user can just open with their favorite Python IDE/editor and run or debug to their liking.

As for reversing the hooks, I'm not sure this needs explicit support in Edalize. You should already be able to write a Python script which leverages Python-APIs of FuseSoC/Edalize for core- and dependency management and then runs the tests via Python-APIs of VUnit. This would be something along the lines of the implementation by @olofk  in his [fusesoc_vunit_demo](https://github.com/olofk/fusesoc_vunit_demo/blob/master/run_vunit.py) repo.

What I was more aiming at with this PR was to enable treatment of VUnit as just another EDA-tool which can be controlled via the same (or at least similar) YAML-description as all the other tools, and keep the special treatment to a minimum.

olofk:
Hi @yottahawk and @julian-becker . Many thanks for working on this. I like the conversation you're having and the points you bring up are exactly the ones that I would like to make. It should be easy for the common case, but still possible to support more advanced cases. Unfortunately I still haven't had time to look at the exact implementations you're proposing as things dragged out a bit more than expected, but you have my full attention and I will look at this as soon as I can

julian-becker:
@olofk @yottahawk I have implemented my above suggestion for a VUnit hooks. You can have a look at it [here](https://github.com/julian-becker/edalize/pull/1/files). If you agree with this direction, I will merge this change into this feature-branch / pull-request.

In order to get a picture of how this works, feel free to have a look at the examples from the extended documentation in `api.rst` as well as some of the VUnit examples, which I have migrated to fusesoc/Edalize [here](https://github.com/julian-becker/vunit/pull/1/files#diff-96b466b71c0aea05efd8dd432fe89d38). In particular, take a look at `examples/vhdl/generate_tests/generate_tests.core` and `examples/vhdl/generate_tests/vunit_runner.py`, which demonstrates how to run the same testbench using a multitude of parameter configurations.

julian-becker:
@olofk Sure, will probably look it over again this weekend, and iron out some of the remaining kinks.

julian-becker:
@olofk I have applied some extensions and consolidations to this PR to the point that I'm -- now -- mostly happy with it (with the exception of `Vunit.get_vunit_runner_path()`... maybe you have an idea of how to improve it?).
I have extended the `test_vunit.py` script to also cover the newly introduced hooks. On my machine, all pytest test-cases run successfully now.

As always, feedback welcome, and -- if you have no objections -- feel free to merge.

I'm sure some shortcomings of this extension will present themselves at some point, but I think it is a reasonable foundation to build on.

olofk:
Excellent! I will try to take a proper look this week. Thanks for working on it

doc/edam/api.rst
```diff
@@ -254,6 +255,67 @@ Field Name       Type                  Description
...
+
+================ ===================== ===========
+Field Name       Type                  Description
+================ ===================== ===========
+vunit_options    List of String        Extra options for the VUnit test runner
+add_libraries    List of String        A list of framework libraries to add. Allowed values include "array_util", "com", "json4hdl", "osvvm", "random", "verification_components"
+vunit_runner     String                Name of the Python file exporting a ``VUnitRunner`` class (must derive from ``edalize.vunit_hooks.VUnitHooks``) that is used to configure and execute test. This allows very customized test control via VUnit's Python-interfaces.
+================ ===================== ===========
+
+In case a more advanced VUnit configuration or execution of the testbench is necessary, the option ``vunit_runner`` can be used to specify the filename of a Python-script which can hook into the construction, parametrization, and execution of the test runner. For this to work, the Python-skript must export a ``class VUnitRunner(vunit_hooks.VUnitHooks)`` which derives from (and optionally overrides) the behavior of ``vunit_hooks.VUnitHooks``.
```
cmarqu: Suggesting some small typofixes, and a linebreak in the source that won't show up in the rendered output.
```suggestion
In case a more advanced VUnit configuration or execution of the testbench is necessary, the option ``vunit_runner`` can be used to specify the filename of a Python script which can hook into the construction, parametrization, and execution of the test runner.
For this to work, the Python script must export a ``class VUnitRunner(vunit_hooks.VUnitHooks)`` which derives from (and optionally overrides) the behavior of ``vunit_hooks.VUnitHooks``.
```

doc/edam/api.rst
```diff
@@ -254,6 +255,67 @@ Field Name       Type                  Description
...
+.. code-block:: python
+
+    from vunit_hooks import VUnitHooks as VH
+    from vunit.ui import Library, Results
+    from typing import Mapping, Collection
+
+    class VUnitRunner(VUnitHooks):
+        '''
+        Example of custom VUnit instrumentation
+        '''
```
cmarqu: Nit: https://www.python.org/dev/peps/pep-0257/#id16 suggests formatting such docstrings as
``"""Example of custom VUnit instrumentation."""``
(If you haven't tried https://pypi.org/project/black/ - it's awesome!)

doc/edam/api.rst
```diff
@@ -254,6 +255,67 @@ Field Name       Type                  Description
...
+            '''
+            vu = VUnit.from_argv()
+            vu.enable_check_preprocessing()
+            return vu
+
+        def handle_library(self, logical_name: str, vu_lib: Library):
+            '''
+            Override this to customize each library, e.g. with additional simulator options.
+            This hook will be invoked for each library, after all source files have been added.
+            :param logical_name: The logical name of the library
```
cmarqu: Maybe edalize should use https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html to be able to write nicer docstrings?

doc/edam/api.rst
```diff
@@ -254,6 +255,67 @@ Field Name       Type                  Description
...
+                entity.add_config(
+                        name     = "TestConfig1",
+                        generics = dict(CLK_FREQ=10000000))
+                entity.add_config(
+                        name     = "TestConfig2",
+                        generics = dict(CLK_FREQ=54687500))
+
+        def main(self, vu: VUnit):
+            '''
+            Override this for final parametrization of the VUnit instance (after all libraries have been added), or for custom invokation of VUnit
```
cmarqu:
```suggestion
            Override this for final parametrization of the VUnit instance (after all libraries have been added), or for custom invocation of VUnit.
```

doc/edam/api.rst
```diff
@@ -254,6 +255,67 @@ Field Name       Type                  Description
...
+            '''
+            Customized creation of the test runner
+            '''
+            vu = VUnit.from_argv()
+            vu.enable_check_preprocessing()
+            return vu
+
+        def handle_library(self, logical_name: str, vu_lib: Library):
+            '''
+            Override this to customize each library, e.g. with additional simulator options.
```
cmarqu: There should be a line break after the first sentence.

edalize/vunit_hooks.py
```diff
@@ -0,0 +1,27 @@
+'''This module exports `VUnitHooks` which can be used to implement advanced VUnit test cases'''
+
+from vunit.ui import Library
+from vunit import VUnit
+from typing import Mapping, Collection
+
+class VUnitHooks(object):
+    '''Derive the VUnitRunner instance from this class and override its member functions if necessary'''
```
cmarqu:
```suggestion
    '''Derive the :class:`VUnitRunner` instance from this class and override its member functions if necessary.'''
```
This will hyperlink the identifier.

cmarqu:
Hi @julian-becker, you are using "direct" type annotations, making this code Python-3-only
(https://mypy.readthedocs.io/en/latest/python2.html). @olofk is this okay with you? I didn't immediately find a statement about supported versions (and there is no ``python_requires`` in ``setup.py``).

olofk:
Thanks for the review @cmarqu . I'm a bit torn about the py2 support. So far I have been trying to keep py2.7 compatibility for fusesoc and edalize. Even if py2 is going away there are still so many systems that either default to py2 or doesn't even have py3 easily installed. But there are a few places already where minor functionality is disabled on py2 because I couldn't be bothered to write a lot of extra code to support py2. In this case it's a new backend, so nothing will break for existing users. So to conclude my thoughts, if it's easy to add py2 support, then let's do that. Otherwise I think it's ok and we can add py2.7 compatibility later if users request it

@julian-becker With the changes that @cmarqu commented fixed, I'm ready to merge this

doc/edam/api.rst
julian-becker: Thanks for pointing out the black tool! Did not come across this yet. Although it did not help me in formatting the docstring. Still had to do it manually... or maybe I'm just using it wrong.

doc/edam/api.rst
julian-becker: This looks indeed very useful for improving the readability.

doc/edam/api.rst
julian-becker: OK, I did a little reformatting of this example code.

edalize/vunit_hooks.py
julian-becker: Thanks for pointing this out to me. I have adjusted this instance and some other places as well.

julian-becker:
Thanks again, @cmarqu for reviewing this! I have applied your suggested changes except for the Python2-rework, since I have presently no idea how to make the mocking in tests/test_vunit.py backwards-compatible with Py27, and do not presently have the time to read-up on the old mocking APIs. I can of course just remove all of the type annotations, but am not sure it is worth it without having a proper test for it which runs on Py27.

olofk:
Alright! Let's forget about py2 for now and get this in. I had to do some shady git things to get this squashed properly. Think I got it right in the end, but it would be great if @julian-becker could just do a sanity check of https://github.com/olofk/edalize/tree/vunit to see that I didn't mess up anything before I push to master

julian-becker:
@olofk I looked it over and verified that the tests are still running... All good 👍

olofk:
Excellent! Picked and pushed. Many thanks for your contributions!

PR #53, opened 2019-09-29
Quartus: Support specification of the FPGA's device number in the JTAG chain

julian-becker:
I have just done some experiments on my DE1-SoC board and ran into issues with the quartus_pgm invokation to download the image to the FPGA. In particular, there are 2 devices in the JTAG-chain on this board, and without specifying the device index explicitly in the `quartus_pgm` invokation, the download fails

    Error (209015): Can't configure device. Expected JTAG ID code 0x02D120DD for device 1, but found JTAG ID code 0x4BA00477. Make sure the location of the target device on the circuit board matches the device's location in the device chain in the Chain Description File (.cdf).
    Error (209012): Operation failed

To accomodate for the requirement to explicitly specify the device index, this PR adds a new option `board_device_index`.

### DIFF
--- a/doc/edam/api.rst
+++ b/doc/edam/api.rst
@@ -182,13 +182,14 @@ vsim_options     List of String        Extra options for running the simulation 
 quartus
 ~~~~~~~
 
-================ ===================== ===========
-Field Name       Type                  Description
-================ ===================== ===========
-family           String                FPGA family e.g. *Cyclone IV E*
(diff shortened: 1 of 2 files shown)


olofk:
Thanks. FYI, I'm always reluctant to add new features to the device programming side of FuseSoC as I ideally want to split that out to a separate library or tool that many project can reuse. But this was pretty straight-forward so I'm ok with adding it

julian-becker:
@olofk OK, that is good to know. Having a dedicated tool for the purpose of programming makes sense and would be really handy

PR #54, opened 2019-10-09
Vivado tweaks

kgugala:
This PR adds a few tweaks to the Vivado backend:

* get_version function
* force flag to project and bitstream are generation


olofk:
Many thanks. Code looks good but breaks test suite. I can amend the patch here and push if you're fine with that

kgugala:
thank for pointing that. I have pushed the update that should fix that.

olofk:
Cool. I squashed the tests commit into the first one and pushed

PR #61, opened 2019-10-26
Vivado: Retry opening a device if it fails

imphil:
Vivado isn't reliably opening a hardware device in v2018.3 (at least). Retry up to three times before giving up.

The first commit updates the tests to always include the `part` key, and check the programming TCL file as well.
The second commit then does the actual rewrite of the programming script.
Please see the individual commit messages for more details.

Both commits in one PR for easier merging, since they modify the same test files.


olofk:
Fantastic. Many thanks. Picked and pushed

PR #64, opened 2019-10-30
Vivado programming: Use correct variable in error message

imphil:
hw_device is not available outside of the loop, I should have used
hw_target instead.

Also improve the error message when programming fails.
Experience shows that a very common error is a wrong jumper setting.
Give users a hint about that.


olofk:
All good. Many thanks. Picked and pushed

Issue #63, opened 2019-10-30
Synthesis only option for Quartus


a-gibson:
I would really like to be able to synthesise my RTL without being forced to do a full P&R. I see this option is available for Vivado as a tool option.

Could we have this option for Quartus as well?

a-gibson:
I have a local branch with proposed changes for this but can't seem to push it. Are there repository restrictions preventing me from creating branches? I can't seem to create branches through the web interface either.

olofk:
Fixed by #66

PR #66, opened 2019-10-31
Quartus synth only option

a-gibson:
As described in #63, allow Quartus to support synthesis-only runs by using the same `pnr` tool option that is available for Vivado.

### DIFF
--- a/edalize/quartus.py
+++ b/edalize/quartus.py
@@ -32,7 +32,10 @@ class Quartus(Edatool):
                          'desc' : 'FPGA device (e.g. 5CSXFC6D6F31C8ES)'},
                         {'name' : 'board_device_index',
                          'type' : 'String',
-                         'desc' : "Specifies the FPGA's device number in the JTAG chain. The device index specifies the device where the flash programmer looks for the Nios® II JTAG debug module. JTAG devices are numbered relative to the JTAG chain, starting at 1. Use the tool `jtagconfig` to determine the index."}],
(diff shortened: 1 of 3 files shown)


a-gibson:
I've run the tests as you've suggested but I get an error. I get the same error when running the tests on the `master` branch as well. Is there a configuration stage I've missed? I've installed PyTest and then run `GOLDEN_RUN=1 py.test`, which produces the following collection error:
```
ImportError while importing test module '/Users/a-gibson/development/github/olofk/edalize/tests/test_verilator.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_verilator.py:6: in <module>
    from edalize_common import compare_files
tests/edalize_common.py:4: in <module>
    from edalize import get_edatool
E   ModuleNotFoundError: No module named 'edalize'
```

Any suggestions would be appreciated.

olofk:
I hadn't seen this error before, but I understand where it comes from now.

It seems you must have edalize installed before the test suite works. I tried uninstalling edalize and got the same error. Running `pip install --user -e .` from the edalize root fixed it for me, so you could try that as well

olofk:
All that needs fixing now is the reference Makefile. I can amend that locally if you prefer

a-gibson:
Thanks for the pointers! I admit I didn't expect to have to install Edalize before being able to test it.

olofk:
That surprised me as well. Hadn't noticed before since I tend to have it installed all the time. Not sure how to get around it, but good that it worked out.

All in all, mayn thanks for the fixes. Picked, squashed and pushed now!

GCHQDeveloper992:
Happened to notice this PR whilst running some tests - when I've been wanting to run tests without installing edalize first, I've been doing the following:

PYTHONPATH=/path/to/root/of/edalize/checkout py.test

This way python finds the edalize files without having to have them installed first.

olofk:
Cool. Thanks. That's good to know. I'll add it when I find the time to write some contribution guidelines

PR #68, opened 2019-11-04
Update Vivado test files to reflect templates

GCHQDeveloper992:
Whilst running some tests I noticed that I was getting errors on the Vivado tests. It appears that the templates have changed, but the reference files didn't. This updates the reference files to match the current templates.


olofk:
Thanks! I hadn't noticed that. Picked and pushed

PR #69, opened 2019-11-09
Create directories for libraries with GHDL backend.

m-kru:
When there are 2 entities with the same name in different libraries they need to be compiled to different directories: [Having 2 similar entities in different libraries results in compilation error](https://github.com/ghdl/ghdl/issues/1013). This patch make edalize creates directories for libraries for GHDL backend.


1138-4EB:
Ref ghdl/ghdl#1013, in order to understand the context.

m-kru:
@olofk ping!

olofk:
Sorry about that. Haven't had much time lately.

This is definitely something we need to fix. I'm a bit worried about pushing mkdir into the makefile as I believe that won't work on Windows. OTOH, I have honestly no idea if it worked on Windows before that either. I'm willing to pull this in so that you aren't blocked by this, but it would be great if we could solve it in a more platform-agnostic way. What's blocking now is that the test suite doesn't pass after this commit

Make sure to run `py.test` in the edalize root. If there are any changes to the reference files, you can in most cases run `GOLDEN_RUN=1 py.test` to have the reference files automatically updated. Check that they look good, add them and make a new patch

m-kru:
What is the test that does not pass? If I run py.test locally all test pass:
`================================= 30 passed in 1.42s =================================`

olofk:
Weird. test_ghdl fails here, which makes sense since the Makefile template is updated.

olofk:
Ok, I might have an idea what's going on here. It seems like it runs the tests against the installed Edalize version even when you run it from the edalize root. Try `PYTHONPATH=. py.test`. I will need to look into this at some point

m-kru:
Tests are fixed. Yeah, it would be nice to fix how the tests are run, as user really does not expect this :).

olofk:
Thanks! Picked and pushed

And yes, I agree. Had no idea myself until another user pointed it out a couple of weeks ago

PR #71, opened 2019-11-12
VCS: Pass cSource/cppSource files to VCS

### LINKED ISSUE #70: VCS: Include cSource files in filelist
VCS can compile C files into DPI modules. Until we have real DPI support, we should at least pass C and header files to VCS.

imphil:
Fixes #70

### DIFF
--- a/edalize/edatool.py
+++ b/edalize/edatool.py
@@ -305,7 +305,11 @@ class Edatool(object):
             _s = "'{}' exited with an error code"
             raise RuntimeError(_s.format(cmd))
 
-    def _write_fileset_to_f_file(self, output_file, include_vlogparams = True):
+    def _filter_verilog_files(src_file):
+        ft = src_file.file_type
+        return ft.startswith("verilogSource") or ft.startswith("systemVerilogSource")
+
+    def _write_fileset_to_f_file(self, output_file, include_vlogparams = True, filter_func = _filter_verilog_files):
         """ Write a file list (*.f) file
 
         Returns a list of all files which were not added to the *.f file
@@ -328,7 +332,7 @@ class Edatool(object):
                 f.write("+incdir+" + id + '\n')
 
             for src_file in src_files:
-                if (src_file.file_type.startswith("verilogSource") or src_file.file_type.startswith("systemVerilogSource")):
+                if (filter_func is None or filter_func(src_file)):
                     f.write(src_file.name + '\n')
                 else:
                     unused_files.append(src_file)
--- a/edalize/vcs.py
+++ b/edalize/vcs.py
@@ -44,10 +44,19 @@ Example snippet of a CAPI2 description file for VCS:
                 return True
         return False
 
-
     def configure_main(self):
+
+        def _vcs_filelist_filter(src_file):
+            ft = src_file.file_type
+            # XXX: C source files can be passed to VCS to be compiled into DPI
+            # libraries; passing C sources together with RTL sources is a
+            # workaround until we have proper DPI support
+            # (https://github.com/olofk/fusesoc/issues/311).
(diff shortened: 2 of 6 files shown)


vogelpi:
Thanks for your fix @imphil. I could test it and the C files are now indeed compiled.

What is missing is the include path for dependencies. For example, I have aes_model_dpi depending on `aes_model` in a different directory. When compiling `aes_model_dpi` the include path should also comprise `aes_model`. If I set the include path manually, it works fine.

imphil:
@vogelpi Some questions here ...
- How do you pass the include path to VCS? 
-W hich include path are you missing? A path for third-party includes (e.g. /usr/include/something), or one to find .h files which are part of a core?

vogelpi:
To compile `aes_model_dpi` we need to include a .h-file from `aes_model` (a different core). The directory structure is as follows:
```
hw/ip/aes
       |-dv
       |  |-aes_model_dpi/
       |  \-scratch/aes/default/build/   // this is our VCS build folder
       \-model/
```
For Verilator this is fine, but other simulators require us to pass
`BUILD_OPTS  += -CFLAGS "-I../../../../../model"` for VCS or
`BUILD_OPTS  += -I./../../../../../../../../model` for Xcelium.

olofk:
@vogelpi Are you setting the `is_include_file` attribute on the h file in `aes_model`? Perhaps I'm missing something but that should add the include path

vogelpi:
Thanks for your input @olofk , the `is_include_file` attribute on the h file in `aes_model` is indeed set and the file type is set to `cSource`.

imphil:
A possible workaround: Set `vcs_options` in the tool section of the core file to something like `-CFLAGS "-I../../../../../model"` for VCS. (There is on xcelium support in edalize today.)

Beyond that, things get tricky without full DPI support (https://github.com/olofk/fusesoc/issues/311).
- You're using `--no-export`, so using `copyto` to put all files in a well-defined directory won't work well.
- We can play with the C include path and pass that to the simulator's compiler, but this will be very simulator-specific.

So for now I'd stay with the vcs_options workaround, piling workarounds on top of that will be more pain than gain.

imphil:
@olofk I talked with @vogelpi and the best approach for now is to merge this as-is, and I'll make sure we have a complete solution once we have DPI support in fusesoc.

vogelpi:
Thanks @imphil for looking into that further.

For now, we will keep passing the include paths manually as we are relying on a manually created Makefile (not generated through fusesoc/edalize). We only use the *.f file output from edalize, i.e., C include paths and other flags passed through `vcs_options` in fusesoc don't make it to VCS.

olofk:
@imphil Ideally I would want the DPI support in place first, since I guess we need to remove this once we have proper DPI support (or not? Don't know very much about VCS+DPI). But I'm ok with pulling this in so that we at least have something for now. Feel free to merge

imphil:
Yes, this can be removed once we have real DPI support, but we still need it in the mean time unfortunately.

I cannot merge myself.

sriyerg:
`is_include_file` attribute adds a `+incdir+` option to the filelist, which is used for the SV compilation. For compilation of C code for DPI-C constructs (from within VCS/Xcelium itself), we need the addition of the options indicated by @vogelpi, to tell VCS's invocation of GCC where to look for additional C headers.  

Filelist is tool-agnostic - it only supports +define+, +incdir+ and list of files. It does not support custom tool options. What we need here can only be achieved with either `vcs_options` in the fusesoc core file or in the external script that is invoking it. 

For Opentitan DV, since we are using fusesoc to only generate the filelist, the `vcs_options` is not an option - we have fixed it within our DV regression flow (which invokes VCS). 

This fix works and allows C sources to be added to the generated filelist. I am reasonably happy with this setup and would like to see this PR merged, unless someone has other objections.

olofk:
Thanks for the patches and the good discussion on the topic

PR #74, opened 2019-11-17
Improve documentation a bit .

cmarqu:


olofk:
Thanks. Guess I should have noticed those FIXME before if I had read the docs myself :)

PR #75, opened 2019-11-17
Add sphinx.ext.napoleon to allow for nicer docs.

cmarqu:
See https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html


olofk:
Many thanks! I have read about this napoleon thing before but never got around to figure out what was needed

PR #79, opened 2019-11-25
Improve documentation; render module API docs.

cmarqu:
This will render the missing content from here: https://edalize.readthedocs.io/en/latest/source/edalize.html

@julian-becker The type annotations said it should be ``class VUnit(Edatool):``, with an uppercase U, so I fixed that - correct?

### DIFF
--- a/edalize/vunit_hooks.py
+++ b/edalize/vunit_hooks.py
@@ -20,7 +20,7 @@ class VUnitHooks(object):
         pass
 
     def main(self, vu: VUnit):
-        """Override this for final parametrization of the :class:`~edalize.vunit.Vunit` instance, or for custom invocation of VUnit."""
+        """Override this for final parametrization of the :class:`~vunit.ui.VUnit` instance, or for custom invocation of VUnit."""
         vu.main()
 
 
--- a/edalize/vunit_hooks.py
+++ b/edalize/vunit_hooks.py
@@ -1,4 +1,4 @@
-"""This module exports :class:`VUnitHooks` which can be used to implement advanced VUnit test cases"""
+"""This module exports :class:`VUnitHooks` which can be used to implement advanced VUnit test cases."""
 
 from vunit.ui import Library
 from vunit import VUnit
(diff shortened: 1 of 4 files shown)


olofk:
IIRC, there is some part of the code that assume that only the first letter of a the name of an Edalize backend is capitalized. I could be wrong though

cmarqu:
Oh, okay. Fixed in the forced push.
This also adds an intersphinx setup for Python and VUnit so these respective Sphinx docs can be cross-referenced (not used yet, but becomes handy with more detailed docs).

edalize/vunit_hooks.py
```diff
@@ -1,30 +1,29 @@
...
         return VUnit.from_argv()
 
     def handle_library(self, logical_name: str, vu_lib: Library):
-        """override this to customize each library, e.g. with additional simulator options"""
+        """Override this to customize each library, e.g. with additional simulator options."""
         pass
 
     def main(self, vu: VUnit):
-        """override this for final parametrization of the :class:`VUnit` instance, or for custom invokation of VUnit"""
+        """Override this for final parametrization of the :class:`~edalize.vunit.Vunit` instance, or for custom invocation of VUnit."""
```
cmarqu: Actually, I'm not sure if this shouldn't instead be
```suggestion
        """Override this for final parametrization of the :class:`~vunit.ui.VUnit` instance, or for custom invocation of VUnit."""
```

olofk:
I'm not very good with sphinx so I'm not sure how to check what's right here. Should I just pull it in or do we need to decide or test something?

edalize/vunit_hooks.py
julian-becker: Yes @cmarqu, this override is intended to allow for custom configuration and invocation of the `vunit.ui.VUnit` instance.

cmarqu:
With the clarifying comment of @julian-becker this is ready to go in in my opinion.

olofk:
Excellent. Thanks for working on this

cmarqu:
Unfortunately, what worked locally doesn't when running at ReadTheDocs. I'm looking into it.

PR #80, opened 2019-11-29
Icestorm tweaks

kgugala:
this PR adds two additional targets to Makefiles generated for icestorm toolchain:

* `timing` - timing report generation with the `icetime` tool
* `stats` - logic usage report generation using `icestat` tool

Those targets are not called when `build()` is called.
Also, the PR adds a simple mechanism for calling the additional targets - user may now pass a target name to `build_main` function e.g:

```
self.backend = edalize.get_edatool('icestorm')(edam=self.edam, work_root=self.out_dir)
self.backend.configure("")
self.backend.build()
self.backend.build_main('timing')
``` 

If this mechanism is OK I can work on adding similar targets to other toolchains

### DIFF
--- a/edalize/edatool.py
+++ b/edalize/edatool.py
@@ -124,9 +124,9 @@ class Edatool(object):
         if 'pre_build' in self.hooks:
             self._run_scripts(self.hooks['pre_build'], 'pre_build')
 
-    def build_main(self):
-        logger.info("Building");
-        self._run_tool('make')
+    def build_main(self, target=None):
+        logger.info("Building{}".format("" if target is None else "target " + " ".join(target)))
+        self._run_tool('make', [] if target is None else [target])
 
     def build_post(self):
         if 'post_build' in self.hooks:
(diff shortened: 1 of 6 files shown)


olofk:
This is a really good idea! Defining targets in the build section allows us to have more fine-grained backend-specific control. We should definitely add this to the other backends as well and try to standardize the target names. I think we can expose this as a backend parameter too so it can be set on the fusesoc command line, e.g. fusesoc run $core --target=timing (not sure if target is the best name for it, as there is already the FuseSoC core file target. Have to think about that). This would also probably be a nicer solution for only running synthesis (like `fusesoc run $core --target=synth` instead of `fusesoc run $core --pnr=none` that is used by some backends already)

olofk:
The only thing I don't like is that $DEVICE gets set to `None` if it's not specified, but in practice that doesn't really matter. The targets that needs $DEVICE would need a real value, and any other value would break it anyway

PR #85, opened 2019-12-09
Vivado: create a .bin file for programming flash

antonblanchard:
A .bin file appears to just be a .bit file with the header stripped, but
Vivado requires it to program flash devices.

### DIFF
--- a/edalize/templates/vivado/vivado-run.tcl.j2
+++ b/edalize/templates/vivado/vivado-run.tcl.j2
@@ -1,3 +1,6 @@
+# Create a bin file which can be used to program the flash on the FPGA
+set_property STEPS.WRITE_BITSTREAM.ARGS.BIN_FILE true [get_runs impl_1]
+
 # Vivado will raise an error if impl_1 is launched when it is already done. So
 # check the progress first and only launch if its not complete.
 if { [get_property PROGRESS [get_runs impl_1]] != "100%"} {
--- a/tests/test_vivado/minimal/test_vivado_minimal_0_run.tcl
+++ b/tests/test_vivado/minimal/test_vivado_minimal_0_run.tcl
@@ -1,3 +1,6 @@
+# Create a bin file which can be used to program the flash on the FPGA
+set_property STEPS.WRITE_BITSTREAM.ARGS.BIN_FILE true [get_runs impl_1]
+
 # Vivado will raise an error if impl_1 is launched when it is already done. So
 # check the progress first and only launch if its not complete.
 if { [get_property PROGRESS [get_runs impl_1]] != "100%"} {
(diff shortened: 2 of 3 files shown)


antonblanchard:
@imphil @olofk I'm trying to add programming of flash devices to fusesoc (or at least reduce the number of manual steps required outside of it). For some reason Vivado wants the header stripped from the file and at the moment I have to click a check box in the gui. Does this look ok?

edalize/templates/vivado/vivado-run.tcl.j2
```diff
@@ -24,3 +27,7 @@ if { [get_property PROGRESS [get_runs impl_1]] != "100%"} {
 # be selected from a file picker first.
 set vivadoDefaultBitstreamFile [ get_property DIRECTORY [current_run] ]/[ get_property top [current_fileset] ].bit
 file copy -force $vivadoDefaultBitstreamFile [pwd]/[current_project].bit
+
+# Copy the .bin file too
+set vivadoDefaultBitstreamBinFile [ get_property DIRECTORY [current_run] ]/[ get_property top [current_fileset] ].bin
+file copy -force $vivadoDefaultBitstreamBinFile [pwd]/[current_project].bin
```
imphil: That copying is only for backwards-compat. I wouldn't do that for newly introduced functionality.

antonblanchard: Ok, will remove

antonblanchard:
I averaged 4 runs of `fusesoc run` of microwatt but the times bounce around a fair bit. The baseline runs were 2% slower than the patched version that created both files, which just shows the run to run variability is greater than the time it takes to create the bin file. Can we just do the bitstream generation with fusesoc for a more isolated test?

olofk:
I agree with Philipp here. If it takes a long time, introduces configuration options or is a niche use I would steer people to implement this in a tcl script and use `file_type : tclSource` to pull it in. In this case however it's very straight-forward and a common thing so I'm happy to just pull it in....as long as the test suite still passes :)

run `PYTHONPATH=. py.test` in the edalize root to see if tests still pass. If not, you can run `PYTHONPATH=. GOLDEN_RUN=1 py.test` to have the test suite automatically update the reference files. Add these to the commit and I'll pull it in

antonblanchard:
Thanks @olofk, I've updated the checks and re-pushed.

mithro:
The .bin takes 10ish seconds in my experience.

imphil:
Sounds like a small enough time in a normal synthesis run to have it enabled by default. LGTM then.

olofk:
Many thanks for the contribution. Picked and pushed

PR #86, opened 2019-12-10
Vivado file handling improvements

imphil:
- Explicitly handle `mem` files (memory initialization, e.g. for `$readmemh()`.
- Pass all unknown file types thorugh to Vivado.

The second change might be slightly controversial, but it makes users of edalize much happier: it is annoying not to be able functionality in Vivado without patching edalize first, and no easy workaround available.


olofk:
I trust you on this. I'm all for explicit file types but don't want it to be a bottleneck. We could possibly do it the other way instead and remove known incompatible file types (e.g. qip) if this turns out to be problematic

PR #87, opened 2019-12-12
Fix sys.path in Sphinx config.

cmarqu:
PR #79 did not render in ReadTheDocs despite working locally on my machine, probably because the relative ``sys.path`` in the Sphinx config tried to go up two directory levels which apparently doesn't exist in the RTD file system.
This PR fixes the problem as can be seen in https://cmarqu-edalize.readthedocs.io/en/latest/source/edalize.html#edalize.vunit_hooks.VUnitHooks

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -15,8 +15,8 @@
 from datetime import datetime
 import os
 import sys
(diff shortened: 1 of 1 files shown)


olofk:
Many thanks

Issue #84, opened 2019-12-09
VUnit support: vhdl_standard should be of type 'str'


krishnan-gopal:
Can you please clarify from which commit of VUnit this can be used ?

Currently, the generated run.py uses the type 'VHDL.standard', and I am getting the error:

> Traceback (most recent call last):
>   File "*****/run.py", line 25, in <module>
>     lib.add_source_files("*****/sourcefile.vhd", vhdl_standard=VHDL.standard("2008"))
>   File "*****/vunit/vunit/ui/library.py", line 216, in add_source_files
>     for file_name in file_names
>   File "*****/vunit/vunit/ui/library.py", line 216, in <listcomp>
>     for file_name in file_names
>   File "*****/vunit/vunit/ui/library.py", line 270, in add_source_file
>     vhdl_standard=self._which_vhdl_standard(vhdl_standard),
>   File "*****/vunit/vunit/ui/library.py", line 368, in _which_vhdl_standard
>     return VHDL.standard(vhdl_standard)
>   File "*****/vunit/vunit/vhdl_standard.py", line 80, in standard
>     return VHDLStandard(name)
>   File "*****/vunit/vunit/vhdl_standard.py", line 27, in __init__
>     if standard.endswith(standard_name) and len(standard_name) == 2:
> TypeError: endswith first arg must be str or a tuple of str, not VHDLStandard

Does this need to be corrected or I am missing something here ?

cmarqu:
Pinging @julian-becker who added the VUnit support.

julian-becker:
When I did the original implementation, I was developing against vunit 4.2.0. Since then, it seems there have been some subtle changes to this API. The changes in #88 should fix this, but unfortunately break compatibility for vunit 4.2.0...

olofk:
@kraigher @larsasplund Do you have any comments on this? Did the API change in a backwards-incompatible way or is there a way to support both older and newer VUnit?

kraigher:
The Public API has always been str. Internally we switched to an object to get a better implementation of some rules based on the standard. In 4.2.0 there was a bug where the Public API accepted the VHDLStandard object which it should not have. In 4.3.0 we fixed this http://vunit.github.io/release_notes.html

So in summay VHDLStandard is a private class that should thus never appear on the public api

olofk:
[...]

PR #88, opened 2019-12-15
VUnit: Pass vhdl_standard argument to `add_source_files()` as string in order to support vunit 4.3.0

julian-becker:
This fix addresses issue #84 with vunit_hdl==4.3.0.

However, it breaks compatibility with vunit_hdl<=4.2.0. If anybody has ideas on how to support both (without too much of an effort) that would be awesome.


olofk:
Thanks. Seems like this is the right way to do it. Picked and pushed

Issue #91, opened 2019-12-23
Allow DSE to be used for Quartus P&R


DHWinterWolf:
I use DSE quite a bit to handle multiple P&R runs in parallel for some complex designs targeting Altera FPGAs.

It looked like it should be reasonably straightforward to use the pnr tool option for Quartus to allow DSE to be used instead of the standard Quartus P&R flow, so I had a go at implementing it for my requirements (a seed sweep with limits on parallel runs). It worked a treat, taking options from a FuseSoC core and producing exactly the result I was expecting.

I'm hoping to tidy this up over the next few days and make it completely configurable for other DSE flow types (hopefully with sensible defaults), then submit a PR in case it is of interest to anyone else. Any comments or suggestions are welcome, of course!

DHWinterWolf:
I've now added tests for this to the PR as well - feel free to shout if anything else is missing to bring this up to standards.

a-gibson:
I've used DSE in the past to speed up CI runs which require a few attempts to find a solution that achieves timing closure.

If FuseSoC could support DSE this will be very helpful in future!

olofk:
Of course FuseSoC supports DSE :)

PR #93, opened 2019-12-30
Add DSE as a Quartus P&R tool option

DHWinterWolf:
Implement support for DSE as a P&R tool for Quartus, issue #91

Augments the current pnr options of "none" and "quartus" to also allow "dse". If chosen, DSE is called after synthesis instead of manually calling the Quartus P&R tools. A DSE configuration file is created with the options provided.

By default, a four seed sweep on the local machine is created, with a maximum of four jobs run in parallel and no CPU limitations in place, but I've tried to make all options that make sense in the context of Edalize be customisable.


DHWinterWolf:
Having read through comments on other merged PRs and then looked into py.test, I've updated the Quratus tests to include tests for DSE and well as Quartus as the P&R engine for both Standard and Pro. Hopefully this is useful.

DHWinterWolf:
And I've just realised that I'd forgotten to set up git properly in a new location. I'll get that fixed now!

olofk:
Sorry. I thought I had already responded to this but apparently that must have been a dream. It's extremely annoying that you can't specify the DSE stuff in a TCL file instead and just load it with `file_type : tclSource`, but after looking at the DSE docs, it really looks like your solution is the only way. My main issue is that it introduces a ton of new backend parameters, but perhaps we will have to live with that. The only other solution I could come up with would be to specify a new file type for DSE options, but that's not pretty either. The patches look solid and seems to add useful functionality so I'll pull it in. Thank you for working on tihs

PR #98, opened 2020-01-29
Behavior change in lstrip_blocks in Jinja 2.11.0

imphil:
Jinja 2.11.0 got released a couple days ago, and it changed the behavior
of lstrip_blocks. This breaks some templates.

It's not yet clear if we misunderstood the lstrip_blocks documentation
(and things were never expected to work that way), or if Jinja2
introduced a breaking change.

Upstream bug report https://github.com/pallets/jinja/issues/1138

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,9 @@ setup(
     ],
     install_requires=[
         'pytest>=3.3.0',
-        'Jinja2>=2.8',
+        # Avoid Jinja 2.11 until https://github.com/pallets/jinja/issues/1138
+        # is sorted.
+        'Jinja2 >=2.8, <2.11',
     ],
     tests_require=[
         'pyyaml',


olofk:
I must say I'm still not super happy with jinja in general, but anyway, let's fix this first at least.

sharkcz:
Olof, is there some workaround possible for the jinja2 issue? Fedora 32 will be released with jinja 2.11.1 and thus making edalize/fusesoc uninstalable.

imphil:
This is scheduled to be fixed in an upcoming Jinja2 release, which hasn't happened yet unfortunately.

In the meantime, I see a couple of options:
- Install edalize and Jinja2 with `pip3 install --user`, avoiding system packages.
- Use a venv.
- Work around it in edalize. I'm ready to accept a PR for that, but I won't have time to work on it myself unfortunately. There are workarounds proposed in https://github.com/pallets/jinja/issues/1138.

olofk:
Or just get rid of that stupid jinja thing. I'm getting increasingly tired of it and it doesn't really add much value

imphil:
Jinja2 has now released 2.11.2 which fixed this regression, I'd expect Fedora can update to this point release in an update.

sharkcz:
already built for Rawhide and F-32 (https://bodhi.fedoraproject.org/updates/FEDORA-2020-3937030d54), F<=31 stay on 2.10

Issue #97, opened 2020-01-27
Riviera: Don't compile every file into a separate compilation unit


imphil:
Currently, the Riviera backend calls `vlog` for each Verilog file individually, which results in each file being a separate compilation unit. This behavior is inconsistent with other tools supported by edalize, which pass the whole file list to the tool, which results in all files being in the same compilation unit.

The main end-user-observable difference is the scope of defines: defines are visible within a compilation unit; if each file is a separate compilation unit, defines are not visible across files.

olofk:
This is the way the ModelSim backend works too. I hadn't noticed until I started working on the SweRV core where I had to introduce a [hack](https://github.com/chipsalliance/Cores-SweRVolf/blob/master/cores/swerv.core#L102) to get around this. I think it would be good to offer support for multi compilation units, but I'm not sure about all the details. One particular example is how to handle VHDL and mixed-language designs. Perhaps that is the only issue and can be easily solved but I don't know enough at this point. A compromise could perhaps be to introduce a backend option for riviera and modelsim.

I wonder a bit also about the examples you brought up with defines. I haven't used Riviera all that much but from memory it would pick up defines from `include files just fine. Is what you describe only an issue when you define things directly in the source files?

imphil:
This came up in https://github.com/lowRISC/ibex/pull/566. Effectively, we did the following:
- A file prim_assert.sv contains multiple defines.
- This file is passed to the tool as a normal SV file (*not* using `is_include_file`)

Now: 

- If we compile everything into one compilation unit: Since all cores depend on the core providing prim_assert.sv, this file is compiled first and defines are available without any `include prim_assert.sv` statements in the source code.
- If we don't compile everything into one compilation unit, all of that doesn't work.

I'm not saying that what we had was "correct" or "solid", so we're fixing that in our source code. But we should still try to make edalize behave the same for all simulators, and that means: decide for a behavior (separate compilation units, or a single one), and implement that behavior consistently.

danielmlynek:
LRM (IEEE1800-2017) defines that 2 types of compilation unit scope (CUS) need to be supported:
- one CUS per compiled file
- one CUS per compilation command (that is vlog/alog for RVR)
In edalize for RVR now each file is compiled separately - bunch of vlog command are generated.
Each vlog creates separate CUS - no chance too see defines from previous vlog (unless you include it in all files)
[...]

PR #99, opened 2020-01-30
Added support for common compilation for Riviera-PRO

dawidzim:
fix for issue: #97

### DIFF
--- a/edalize/rivierapro.py
+++ b/edalize/rivierapro.py
@@ -1,6 +1,6 @@
 import os
 import logging
-
+import sys
 from edalize.edatool import Edatool
 
 logger = logging.getLogger(__name__)
@@ -24,6 +24,9 @@ class Rivierapro(Edatool):
                         {'name' : 'vsim_options',
                          'type' : 'String',
                          'desc' : 'Additional run options for vsim'},
+                        {'name' : 'compilation_mode',
+                         'type' : 'String',
+                         'desc' : 'Common or separate compilation, sep - for separate compilation, common - for common compilation'},
                         ]}
 
     def _write_build_rtl_tcl_file(self, tcl_main):
@@ -33,6 +36,12 @@ class Rivierapro(Edatool):
         vlog_include_dirs = ['+incdir+'+d.replace('\\','/') for d in incdirs]
 
         libs = []
+        common_compilation = []
+        common_compilation += ['vlog']
+        common_compilation += self.tool_options.get('vlog_options', [])
+        common_compilation += vlog_include_dirs
+        for k, v in self.vlogdefine.items():
+            common_compilation += ['+define+{}={}'.format(k,self._param_value_str(v))]
         for f in src_files:
             if not f.logical_name:
                 f.logical_name = 'work'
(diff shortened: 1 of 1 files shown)


edalize/rivierapro.py
```diff
@@ -84,7 +90,20 @@ def _write_build_rtl_tcl_file(self, tcl_main):
                 args += ['-quiet']
                 args += ['-work', f.logical_name]
                 args += [f.name.replace('\\','/')]
-                tcl_build_rtl.write("{} {}\n".format(cmd, ' '.join(args)))
+                common_compilation += [f.name.replace('\\','/'),'\\\n']
```
olofk: If Riviera uses forward slashes on all platforms, it's better to call the `get_fileset_files()` function with `force_slash=True`

edalize/rivierapro.py
```diff
@@ -24,6 +24,9 @@ def get_doc(cls, api_ver):
                         {'name' : 'vsim_options',
                          'type' : 'String',
                          'desc' : 'Additional run options for vsim'},
+                        {'name' : 'compilation_mode',
+                         'type' : 'String',
+                         'desc' : 'Common or separate compilation, sep - for separate compilation, common - for common compilation'},
                         ]}
```
olofk: This parameter should not be in the `lists` section, but in a section called `members`. See e.g. [here](https://github.com/olofk/edalize/blob/master/edalize/verilator.py#L43) for a backend that uses both lists and members.

Also, when this change is made, the check for them will be `if (self.tool_options.get('compilation_mode'))=='sep'` instead of `if (self.tool_options.get('compilation_mode'))==['sep']`

edalize/rivierapro.py
```diff
@@ -84,7 +90,20 @@ def _write_build_rtl_tcl_file(self, tcl_main):
...
+                if (self.tool_options.get('compilation_mode'))==['sep']:
+                    tcl_build_rtl.write("{} {}\n".format(cmd, ' '.join(args)))
+
+        if (self.tool_options.get('compilation_mode')==['common']) or (self.tool_options.get('compilation_mode')==None):
+            tcl_build_rtl.write("{} \n".format(' '.join(common_compilation)))
+			
+        if not (self.tool_options.get('compilation_mode')==['common'] or self.tool_options.get('compilation_mode')==None or self.tool_options.get('compilation_mode')==['sep']):
+            print('wrong compilation mode, use --compilation_mode=common for common compilation or --compilation_mode=sep for separate compilation')
+            sys.exit()
+
```
olofk: Instead of print and exit we should raise a `RuntimeError` when we encounter illegal options. You can check how it's done in the verilator backend for handling the `mode` option for reference https://github.com/olofk/edalize/blob/master/edalize/verilator.py#L86-L97

edalize/rivierapro.py
```diff
@@ -84,7 +90,20 @@ def _write_build_rtl_tcl_file(self, tcl_main):
...
                 args += [f.name.replace('\\','/')]
-                tcl_build_rtl.write("{} {}\n".format(cmd, ' '.join(args)))
+                common_compilation += [f.name.replace('\\','/'),'\\\n']
+
+                if (self.tool_options.get('compilation_mode'))==['sep']:
+                    tcl_build_rtl.write("{} {}\n".format(cmd, ' '.join(args)))
+
+        if (self.tool_options.get('compilation_mode')==['common']) or (self.tool_options.get('compilation_mode')==None):
+            tcl_build_rtl.write("{} \n".format(' '.join(common_compilation)))
+
```
olofk: I would prefer the default mode to be `sep` to avoid risk breaking people's code if they rely on the old behaviour.

dawidzim:
Thanks for review, on this commit I added all changes what you propose

dawidzim:
@olofk What you think about this?

olofk:
Thanks for this. Looks fine now. Picked and pushed

PR #100, opened 2020-02-01
Add Verible lint backend

### LINKED ISSUE #95: Support for verible style lint and formatter
Just putting down a note here that it would be great if we can eventually add support for Verible, an open-source style linter/formatter (see: https://github.com/google/verible).

I will add some examples of how this tool can be invoked later.

@imphil @fangism for visibility.

imphil:
This backend can do verible lint only for now (no verible format).

I initially tried to do verilog_lint and verilog_format in one backend, but that turns out to be too messy, as we want actually different tool options for lint and format (e.g. the "rules" or "ruleset" options don't make sense for verilog_format).

So going with lint-only for now, looking into Verible format later.

Fixes #95

### DIFF
--- a/edalize/veriblelint.py
+++ b/edalize/veriblelint.py
@@ -1,0 +1,76 @@
+import logging
+import re
+import os
+import subprocess
+
+from edalize.edatool import Edatool
+
+logger = logging.getLogger(__name__)
+
+class Veriblelint(Edatool):
+
+    argtypes = ['vlogdefine', 'vlogparam']
+
+    @classmethod
+    def get_doc(cls, api_ver):
+        if api_ver == 0:
+            return {'description' : "Verible lint backend (verilog_lint)",
+                    'members': [
+                         {'name': 'ruleset',
+                          'type': 'String',
+                          'desc': 'Ruleset: [default|all|none]'},
+                    ],
+                    'lists': [
+                         {'name' : 'verible_lint_args',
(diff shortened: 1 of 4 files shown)


imphil:
This needs one more update to return an non-zero exit code if lint fails. I'll update this PR with that soon.

mithro:
FYI - @hzeller

edalize/veriblelint.py
```diff
@@ -0,0 +1,76 @@
...
+                          'type': 'String',
+                          'desc': 'What rules to use. Prefix a rule name with "-" to disable it.'},
+                    ]}
+
+
+    def build_main(self):
+        pass
+
+    def _get_tool_args(self):
+        args = [ '-lint_fatal', '-parse_fatal' ]
```
fangism: Yes, these are the options you want.  (These weren't default because the infrastructure this plugged into wanted exit-status 0.)

olofk:
So, are we ready to go with this one. It looks beautiful. Just want to make sure it works too :)

imphil:
Yes it works, I have tried it on Ibex and opentitan. We will need to extend it once we have waiver support, but that's for a follow up.

olofk:
Many thanks for this. Looking forward to try it out myself

msfschaffner:
Thanks for taking care if this @imphil !

Issue #21, opened 2019-03-29
Getting stdout from tool


benreynwar:
It'd be useful to be able to call a tool and get easy access to stdout and stderr from that tool.

Often I'll have a script that processes these to check that everything went OK.

I'd like to modify `Edatool._run_tool` so that it returns a string containing the output from the tool (combined stdout and stderr).  `_run_tool` would also contain optional argument to save stdout and stderr to files, and/or to output them to the same stdout and stderr as the calling process.

olofk:
This would be great. I have a couple of use cases I would like to have supported here
1. As a developer, creating a python tool that uses Edalize, I would like to access stdout/stderr independently of each other
2. As a user of FuseSoC I would like to run FuseSoC (and thereby Edalize) with minimal output in the normal case, but have access to the full output if something goes wrong
3. As a user of FuseSoC I would like to set a verbose mode and a silent mode

2 or 3 is probably most relevant for FuseSoC, but I'm mentioning it here so that the design in Edalize enables this

qarlosalberto:
It would be a really nice feature!!

qarlosalberto:
Is it on your roadmap? We would like use this feature in TerosHDL.

a-gibson:
In the past I've used tool output to help automated CI flows make decisions on build and test outcomes and sometimes it has been necessary to use stdout or stderr. I could definitely make use of this feature if it were available.

olofk:
It sounds like there are multiple uses for this. Just started looking into it hoping it would be a simple hack but after reading up a bit I'm not so sure anymore. The documentaion warns against capturing stdout when using subprocess.check_call and recommends using subprocess.run instead, which would break Python 2. Perhaps this is the time that Python2 support is dropped, or maybe it works fine with check_call. Does anyone have experience with this?

a-gibson:
Sadly this is stretching my knowledge quite far. I'd be wary of using `subprocess.check_call` because of the warning that the child process will block if it produces too much for stdout. These tools can produce an awful lot of stdout.

I know we've used `subprocess.Popen()` in the past. Briefly looking at the documentation it looks like `Popen` is a more flexible option but I don't know what the implications/drawbacks are.

mithro:
There is a `subprocess.check_output()` -- **however** the entire stdout / stderr will be stored in memory.

olofk:
[...]

PR #101, opened 2020-02-02
Common yosys module

kgugala:
This PR extracts Yosys wrapper from icestorm and trellis wrappers. Yosys is now defined as standalone edatool.

Also, yosys flow has been reimplemented to use tcl flow.


mithro:
@kgugala / @olofk - Any idea how to deal with Yosys plugin support?

kgugala:
> @kgugala / @olofk - Any idea how to deal with Yosys plugin support?

@mithro this can be done as a tool option to Yosys edatool wrapper. If this option will be set plugin loading will be injected to yosys tcl script. 

@olofk what do you think?

olofk:
Thanks for this! It will simplify things a lot for the increasing numbers of backends that can use yosys.

I found one issue though. It looks like passing backend arguments on the command-line is broken. Tried it with serv by running `fusesoc run --tool=icestorm serv --pnr=none` and get `error: unrecognized arguments: --pnr=none` even though the pnr option is visible with `fusesoc run --tool=icestorm serv --help`. I guess it has to do with the options parsing. (As a side note, I'm planning to move the options parsing code from edalize to fusesoc at some point because I think it makes more sense to have it there)

kgugala:
oh, I know what happens here. I pass all the provided args to subtool (yosys in this case). Yosys does not have `--pnr` option so it fails. I have to filter the opts before passing it further.

kgugala:
I think the latest changes should fix the arguments issue. @olofk PTAL

olofk:
This allows me to set the backend arguments, but it also allows any arguments to slip through. I would prefer to not allow this as it makes it harder to detect typos or invalid arguments that are specified on the command line.

I also notice that yosys is no longer called in quiet mode. It doesn't matter all that much, just mentioning it in case it was unintended. (As a side note, we should soon fix https://github.com/olofk/edalize/issues/21 so that it's possible to route output to files or an external tool using edalize)

kgugala:
I have added a mechanism for checking the arg. Now the args are passed correctly and we can detect (and report) incorect params. @olofk PTAL one more time

olofk:
Hi @kgugala 

I have finally had time to look at the code now. First of all, I took the liberty to massage these into four patches to make it easier for me to review. It's in the https://github.com/olofk/edalize/tree/cy branch. Could you please take a look there to see that I didn't mess anything up

I did some tests also and found an issue. It seems that string arguments passed through yosys with tcl needs to be wrapped in `{"value"}` . The current code just does `"value"` which makes yosys angry.

You can test it yourself with

    fusesoc library add fusesoc-cores https://github.com/fusesoc/fusesoc-cores
    fusesoc library add serv https://github.com/olofk/serv
    fusesoc run --target=tinyfpga_bx servant

I would also like to exchange the raised Exception with with RuntimeError, as FuseSoC catches those but not Exception.

I can change amend the patches here if that's easier for you

olofk:
I have another update here. Seeing that the most effort was spent on battling the argument parsing, I decided to look at moving that part from Edalize into FuseSoC. I've been meaning to do that anyway but never prioritized it.

Turns out it was actually easier than I had expected. A bit more work needs to be done for runtime parameters, but the backends you have been working on only support compile-time parameters anyway.

The result of this is that I think we can simplify things on the Edalize side. This however requires some coordination, so I will push out a new release of FuseSoC first and then doing the plumbing in edalize

kgugala:
HI @olofk the split you did seems to be OK. @piotr-binkowski found another issue with this PR - the templates were not installed. I have pushed his fix to this PR, please cherry-pick it to your branch.

I'll take a look at string arguments

kgugala:
I have also fixed tcl quoting

olofk:
ok, just a quick update. I am almost at the point where argument parsing is dropped from edalize. I want to get the argparse changes done first as this will drop some lines from your PR. Sorry for taking so long

olofk:
Ok, it's done now. Many thanks for the patience. I will release a new Edalize version now and then we can look at the follow-up patches.

PR #109, opened 2020-02-23
Added option to chose jtag cable

joshuisken:
Hi,

I have multiple Cyclone V boards connected, so I wanted to be able to select the programming jtag cable. Tried with blinky successfully.

Regards,
Jos


olofk:
Thanks for this. Patch applied.

As with all contributions that affect the FPGA programming side of Edalize I also want to say that I really really would like to see this part broken out to a separate library :)

Issue #112, opened 2020-02-25
Specify a custom quality of fit algorithm for Quartus DSE


a-gibson:
Quartus DSE allows users to specify their own algorithm for determining a 'quality of fit' score. The algorithm is expressed in a Python file, to path to which must be included in the `.dse` file (or provided through the DSE GUI, which isn't appropriate for FuseSoC).

Support an additional tool option so that this information can be written to the `.dse` file. Note that if no path is provided DSE will use a default algorithm (the path to which does not need to exist in the `.dse` file).

PR #113, opened 2020-02-25
DSE quality of fit

a-gibson:
See #112

### DIFF
--- a/edalize/quartus.py
+++ b/edalize/quartus.py
@@ -60,7 +60,10 @@ class Quartus(Edatool):
                          'desc' : 'Maximum number of P&R jobs DSE may run at once, default is 4'},
                         {'name' : 'seeds',
                          'type' : 'String',
-                         'desc' : 'Comma seperated list of seeds for DSE to try, default is "1,2,3,4"'}],
+                         'desc' : 'Comma seperated list of seeds for DSE to try, default is "1,2,3,4"'},
+                        {'name' : 'quality_of_fit',
+                         'type' : 'String',
+                         'desc' : 'Path to a Python file containing the algorithm that determines the quality of fit score'}],
                     'lists' : [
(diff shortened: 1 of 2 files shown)


imphil:
Can you please run `pytest` in the repository root to ensure that all tests still pass? If they don't (which I assume), you can update the golden references by running `GOLDEN_RUN=1 pytest`

a-gibson:
Sorry, that took a little longer than anticipated. It seems that we need to use a version of `Jinja2 < 2.11.0`. Because I had set up a new environment and didn't install FuseSoC or Edalize (instead setting PYTHONPATH to my local copy of the Edalize repository) I installed the latest version of `Jinja2` and subsequently got lots of warnings and several test failures that didn't appear to make much sense.

Now that `pytest` successfully runs on `master` branch I ran it again against my branch and all tests still pass. However, one file (`tests/test_vunit/run.py`) was re-written and git shows as changed but the diff doesn't show any changes so haven't pushed it. Does that sound about right?

a-gibson:
Is there anything else I need to do for this PR?

olofk:
Patch looks fine and I'll bring it in, but I have some additonal comments below

I'm not super happy by the amount of new parameters needed for the quartus backend to support DSE. Whenever Intel decides to add or remove parameters to dse it will affect Edalize. Looking at it with fresh eyes I think we should do like in the spyglass backend and just have a `dse_options`parameter and leave it up to the user to submit the correct arguments (We could still do some checking inside the backend but not when parsing the arguments).

If so, I think we should do it sooner rather than later. We can introduce dse_options and keep the existing arguments but mark them deprecated for a while to make sure no one gets too surprised by this.

a-gibson:
Thanks for the additional comments. I see your point of view, having a neater/cleaner interface would be nice.

I'll create another issue to discuss options.

PR #114, opened 2020-02-26
ghdl backend: include and run instead of analyze and elab-run

csantosb:
Concerning this [fusesoc issue](https://github.com/olofk/fusesoc/issues/356) this commit replaces the analyze and elab-run strategy by the include and compile mentioned in there. This makes the order of rtl sources in core file irrelevant.

### DIFF
--- a/edalize/ghdl.py
+++ b/edalize/ghdl.py
@@ -58,24 +58,29 @@ TOPLEVEL = {toplevel}
 ANALYZE_OPTIONS = {analyze_options}
 RUN_OPTIONS = {run_options}
 
-all: analyze
+VHDL_SOURCES = {vhdl_sources}
 
-run:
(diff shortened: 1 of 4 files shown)


olofk:
Thanks! Patch looks fine from a quick look, but please make sure the test suite still passes. Run `pytest` from the repo root.  If there are errors, update the reference files. You can in most cases update them automatically with `GOLDEN_RUN=1 pytest`. Add them and resubmit the patch

csantosb:
Done. I've managed to fix the reference Makefile, I'm afraid I still get errors in .cmd files, sorry.

olofk:
Yeah. That turned out to be a bit tricky. I managed to get it fixed now though, so I took the liberty to add the needed changes to your patch, squash it together and push. Thanks again for the patch

PR #116, opened 2020-03-01
Add back missing dependency and clarify import and make phases

csantosb:
I forgot a dependency on `make libraries directories` when replacing the analyze target.

Then, I think importing and making executable is more appropriate than including (as for ghdl documentation).


olofk:
Did you check that the test suite still passes? There are no updates to the reference files

csantosb:
Sorry, I forgot. This is done now, pytest shows no errors.

olofk:
Thanks! Merged now

Issue #107, opened 2020-02-20
ModelSim exit codes on failure are always 0


GCHQDeveloper992:
As part of integrating FuseSoC & Edalize with our continuous integration setup, I've noticed that ModelSim always has an exit code of 0 when tests fail with either:

- A REPORT with severity ERROR
- A call to std.env.finish, passing a non zero error count
- A call to std.env.stop, passing a non zero error count

It appears that ModelSim treats the completion of a simulation, regardless of errors/warnings, as a success. 

In a previous CI setup, I remember using the 'onbreak' macro to set a flag that could be passed to the 'quit' functions '-code' argument which allowed CI to detect a non zero exit code successfully. I've made an example patch [on my fork](https://github.com/GCHQDeveloper992/edalize/tree/patch/return-error-codes-from-modelsim) which implements something similar. 

Do you think this is a reasonable way to support returning error codes for easier CI integration? I'll raise a pull request if this seems sensible.

a-gibson:
I've been integrating FuseSoC/ModelSim flows into my CI system and hadn't noticed this, which is kinda worrying as I may not be detecting errors! Will need to go have a look.

I presume a `REPORT` with `SEVERITY FAILURE` also causes ModelSim to exit with a code of 0? I know you can change some attributes around this in the `modelsim.ini` file.

It's seems a shame to have to write a wrapper around this but if ModelSim doesn't provide a convenient way to exist with a non-zero status then I think your option is as good as any.

a-gibson:
As I suspected, increasing the severity made no difference.

I have stumbled across this:
```
quit -code [coverage attribute -name TESTSTATUS -concise]
```

So the existing line in the `Makefile` could be modified from:
```
$(VSIM) -do "run -all; exit" -c $(addprefix -pli ,$(VPI_MODULES)) $(EXTRA_OPTIONS) $(TOPLEVEL)
```

to:
```
$(VSIM) -do "run -all; quit -code [coverage attribute -name TESTSTATUS -concise]" -c $(addprefix -pli ,$(VPI_MODULES)) $(EXTRA_OPTIONS) $(TOPLEVEL)
```

Would that simplify things? I've only done a very quick test by hacking a Makefile so this maybe doesn't account for all scenarios.

olofk:
Agree. This is awkward. I scanned the vsim options for any relevant settings but came up blank as well.

[...]

PR #117, opened 2020-03-02
Allow severity ERROR and FAILURE to propagate error codes from ModelSim

GCHQDeveloper992:
As discussed in issue #107, ModelSim doesn't propagate error/failure information up to the shell. This makes integration with CI quite difficult. Based on a suggestion in that issue ticket, I've added some code that can check for the use of severity FAILURE or ERROR messages and pass that up to the shell as an exit code. Using NOTE or WARNING does not trigger an exit code change.

The change calls the coverage function twice as I was having real issues trying to store the returned value in a variable (using set, variable or global never seemed to work). If anyone has a suggestion, that'd be great.


DHWinterWolf:
This also brings things somewhat in line with PR #122 - the consistency certainly feels nice to me here!

olofk:
Alright. Let's go with this. If any complaints comes up we'll deal with that later

PR #120, opened 2020-03-12
Add backend for Verible's verilog_format

imphil:
A very simple backend for Verible's verilog_format tool. Tool parameters
are passed through, no further magic happens.


mithro:
FYI - @hzeller

Issue #121, opened 2020-03-12
Vivado's exit status doesn't reflect synthesis result


a-gibson:
I've been trying to automate Vivado synthesis builds into my CI flow and noticed that Vivado returns with status `0` so long as synthesis ran, regardless of result.

From my perspective it is particularly helpful if the exit status reflects the synthesis result because that will allow tools like Jenkins to more easily determine and report accurate results.

I won't have thought over all use cases but I can't think why this would be a bad idea. It would also make behaviour consistent with other tools, such as Quartus.

olofk:
Fixed with #122

PR #122, opened 2020-03-12
Make Vivado exit with a non-zero exit status if synthesis fails.

a-gibson:
I found this article on a Xilinx forum:
[https://forums.xilinx.com/t5/Vivado-TCL-Community/Using-to-TCL-to-determine-if-synthesis-completed-ok-or/td-p/536155](https://forums.xilinx.com/t5/Vivado-TCL-Community/Using-to-TCL-to-determine-if-synthesis-completed-ok-or/td-p/536155)

It describes how you can query the status of a synthesis run. I've used this to influence the exit status value from a Vivado synthesis run, which is passed by Make to FuseSoC.

Please let me know if you agree/disagree with this propagation of exit values. I think this makes it more consistent with other tools.

I've run `pytest` and all tests pass.

#121

### DIFF
--- a/edalize/templates/vivado/vivado-synth.tcl.j2
+++ b/edalize/templates/vivado/vivado-synth.tcl.j2
@@ -1,2 +1,3 @@
 launch_runs synth_1
 wait_on_run synth_1
+exit [regexp -nocase -- {synth_design (error|failed)} [get_property STATUS [get_runs synth_1]] match]


DHWinterWolf:
That definitely makes it more consistent with Quartus, and would help with automating runs of FuseSoC in test harnesses.

olofk:
I think this makes sense. Picked and pushed. What happens when synthesis passes but P&R fails? We can take that as a follow-up patch if there are things required for that as well

a-gibson:
Thanks for merging my PR. :)

That is a good question. Preliminary testing has shown that the *_run.tcl script can already exit with a non-zero exit status if it fails to complete. I will do some further testing with other failure modes to check that it is consistent.

The exit status for builds that fail timing closure is something else to consider. At the moment, Quartus / DSE will exit 0 so long as PAR completed, irrespective of timing closure results. It is up to users to interpret the results to determine whether or not the build passed or failed. I will test that Vivado is consistent with this. If you think FuseSoC should be part of this analysis, I’d be interested to hear your views.

DHWinterWolf:
Timing analysis is a bit of an interesting one. Some would consider the build done if a bitsream, PROM image or similar was produced - at that point you have a design that can be sent to the FPGA. They may consider timing being met to be another design issue — a timing issue might be treated similarly to any other bug. Alternatively, if trying to verify that a change has not introduced a huge combinatorial path, a failure of timing might be something you'd absolutely treat as an error.

And the whole things gets complicated if using something like DSE that produced multiple results. And goodness knows what happens if considering something like partial reconfiguration of a target device. So this is really a complex issue that there probably isn't a single right answer to. I'm thinking that failing due to this would be contentious. I think PAR not running far enough to be able to produce a bitstream or equivalent should probably be a failure, though, as at that point you couldn't really consider the system to have been built.

olofk:
Timing errors can't exit with a non-zero status. Too many people are just crap at writing timing constraints and the tools are sometimes way too moody, so that you have to go to extreme lengths to make them stop whining about completely unrelated issues. I was more thinking about things like running out of resources or DRC issues

PR #124, opened 2020-03-13
Correct vvp wave dump argument position and change from LXT2 to FST

GCHQDeveloper560:
When the -lxt2 argument is before the input file it is interpreted as
using the file "xt2" as a log file. The correct position is after the
input file. Also change the format from LXT2 to the newer FST.


olofk:
I have to make a confession. I have know that -lxt2 creates a log file called xt2 for about five years now, but no one ever complained so I never bothered fixing it :)

I guess that was just as well since lxt2 was deprecated in favor of fst. Thanks for the patch. Picked and pushed

PR #125, opened 2020-03-16
Fix string parameter passing for verilator

GregAC:
Strings parameters must be quoted in the .vc, otherwise you get errors
of the form:

'Illegal character in decimal constant:'

### DIFF
--- a/edalize/verilator.py
+++ b/edalize/verilator.py
@@ -124,7 +124,7 @@ class Verilator(Edatool):
             f.write('--exe\n')
             f.write('\n'.join(opt_c_files))
             f.write('\n')
-            f.write(''.join(['-G{}={}\n'.format(key, self._param_value_str(value)) for key, value in self.vlogparam.items()]))
+            f.write(''.join(['-G{}={}\n'.format(key, self._param_value_str(value, str_quote_style='\\"')) for key, value in self.vlogparam.items()]))
             f.write(''.join(['-D{}={}\n'.format(key, self._param_value_str(value)) for key, value in self.vlogdefine.items()]))
 
         with open(os.path.join(self.work_root, 'Makefile'), 'w') as makefile:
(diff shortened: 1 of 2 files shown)


GregAC:
This is specified in the verilator manual: https://www.veripool.org/wiki/verilator/Manual-verilator

under '-Gname=value':

```


    Overwrites the given parameter of the toplevel module. The value is limited to basic data literals:

    Verilog integer literals

        The standard verilog integer literals are supported, so values like 32'h8, 2'b00, 4 etc. are allowed. Care must be taken that the single quote (I') is properly escaped in an interactive shell, e.g., as -GWIDTH=8\'hx.
    C integer literals

        It is also possible to use C integer notation, including hexadecimal (0x..), octal (0..) or binary (0b..) notation.
    Double literals

        Double literals must contain a dot (.) and/or an exponent (e).
    Strings

        String must in double quotes (""). On the command line it is required to escape them properly, e.g. as -GSTR="\"My String\"" or -GSTR='"My String"'.
```

olofk:
Is this really correct? I know that it has been explicitly stated before that quotes and escapes are not allowed in the .vc files. Maybe this has changed? @wsnyder, can you clarify?

GregAC:
Certainly on verilator 4.028 this seems to be required, could of course be a verilator bug!

olofk:
I definitely think you're right that it needs to be quoted, but IIRC the problem was that no quotes or \ was allowed in the config file. I remember thinking that the only correct way was to pass string parameters as command-line args to verilator instead of writing them to the file. If this have changed I'm all happy, but would like to check with some older verilator versions (or hear from Wilson) to feel safe

olofk:
Found it! It was added in 4.022 according to the ChangeLog
Related issue https://github.com/verilator/verilator/issues/1535

After thinking some more I think this should be ok still. Those using older verilator versions will just have to stay away from quoted parameters

wsnyder:
@olofk found what I was going to say... if there's anything more let me know.

Issue #115, opened 2020-02-27
Support for Intel/Xilinx Power Analysis Tools


GCHQDeveloper560:
I've been experimenting with the Quartus (`quartus_pow`) and Vivado (`report_power`) power analysis tools and would like to add support to Edalize/FuseSoC

For the most accurate results these tools require post-place and route simulation to generate VCD (Quartus) or SAIF (Vivado) files with information on signal activity. Therefore, I need to add generation of a timing simulation netlist and SDF using `quartus_eda` or `write_verilog`/`write_sdf`. Is this low enough overhead that I should just add it to the existing default flows? If this is an uncommon enough use case that it shouldn't be a default, would adding a `netlist` option for `pnr`, a separate option like `timing_netlist`, or something else make the most sense? Is there a better integration strategy like a post-build hook?

It doesn't appear that Edalize (or IP-XACT) has file types for SDF, SAIF, or VCD, but I don't think Edalize does much with these and it's up to the tool to check `file_type`.

The next (and longest) step is running the post-place and route simulation. I'm not sure whether this will require any changes for ModelSim or other simulators. The main simulation challenge is the load of plusargs recommended, at least for the Quartus flow, and finding device libraries. This is more of a FuseSoC issue, but it would be nice to be able to chain together targets/tools to be able to do the build, netlisting, simulation, and power analysis with a single command.

Finally, the best way to integrate the power analysis step isn't clear. Would yet another `pnr` option (`power_analysis`?) make the most sense?

Thanks for any thoughts!

olofk:
Apologies for the long wait. As you say, netlist simulation is a pain with the current Edalize structure. I started looking into the next major version of Edalize and summarized some thoughts [here](https://github.com/olofk/edalize/wiki/Edalize-(Slight-return)). Netlist simulations is a covered use case there. At this point I'm not sure however when I would find time to look at this, so I hope you can (or at this point, preferably already have) come up with a solution that works for you in the meantime.

I think a power_analysis pnr option is fine with the current architecture

PR #127, opened 2020-04-06
Add Timing Simulation Support for ModelSim and XSim

GCHQDeveloper560:
As the first step in #115 these commits add an SDF file type, with information about the instance the file should be applied to and the corner the delays represent.

In addition to running the simulations with the SDF files, the commits add a `logged_hdl_objs` option that allows objects to be logged to VCD/SAIF files rather than solely using `$dumpvars`. An alternative would be to use a simpler run script and allow user Tcl files to be executed during the run phase for things like logging. However, the ModelSim backend currently uses Tcl files in the build phase and it's not clear how to support user Tcl in both situations.


mithro:
FYI - @kgugala

olofk:
I'm having a hard time judging the correctness of these patches since SDF and such really isn't my domain. But they are more invasive than what I would prefer, so I think we might need to discuss this a bit.

Unfortunately my time is limited right now and I have a ton of patches to go through, so realistically I don't expect to get to this very soon.

paller:
I don't know how much ModelSim is different but for xsim would it be possible to

- Reduce the base tcl file to something like
```tcl
if { [expr $argc > 0] } {
    foreach user_script $argv {
        source $user_script
    }
}

run -all
```
Then the user can modify the execution in any way they like by providing tcl files.

In that case I'm not sure how to handle
```tcl
flush_vcd
close_vcd
close_saif
```
- Change the run command
```make
run: xsim.dir/$(TARGET)/xsimk
	xsim --tclbatch edalize_run.tcl $(XSIM_OPTIONS) $(TARGET) $(LIST_OF_USER_SCRIPTS)
```

PR #129, opened 2020-04-06
veriblelint: pass rules as filetype

kgugala:
This PR refactores #128. The previous implementation was OK when used with edalize only, but it required passing the whole path to the rules file in fusesoc core file.

New implementation assumes that user simply provides `linterRules` file type and edalize will handle the rest. 

Here is the usage example: https://github.com/antmicro/ibex/commit/af4d23c558f298e65955564d63968c91ad01265b

@imphil @olofk what do you think?

### DIFF
--- a/edalize/veriblelint.py
+++ b/edalize/veriblelint.py
@@ -19,9 +19,6 @@ class Veriblelint(Edatool):
                          {'name': 'ruleset',
                           'type': 'String',
                           'desc': 'Ruleset: [default|all|none]'},
-                         {'name': 'rules_config_file',
-                          'type': 'String',
-                          'desc': 'Path to the rule config file'}
                     ],
                     'lists': [
                          {'name' : 'verible_lint_args',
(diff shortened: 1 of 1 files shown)


imphil:
Can there be multiple linter rule files? We also need to think a bit more about the file type; "linterRules" is a bit too generic, maybe 'veribleLinterRules'? And is it only linter rules, or config in general? Is the same config file also applicable to verible format?

tgorochowik:
@imphil for now verible only accepts one file with linter configuration. It uses exactly the same format as the `--rules` switch, so it does support configuration too (where applicable).
There is no equivalent for the formatter at the moment, please let us know if you think any of the features you mention could be useful for you (preferably by creating an issue, or at least confirming here).

imphil:
Thanks for the update @tgorochowik.

For this PR:
- Since the file is lint and other config, can name the file type something like `veribleConfig`?
- Since the file can be passed only once error out in the code if it appears more than once.

Changes beyond that are out of scope for this PR, but I can certainly imagine verible format to gain the same functionality in the future. I'm not sure about the ability to specify more than one config file, as merging will become a pain. That's something we support for Verilator, since it has waivers and other config in the same file -- so the next steps depend on how Verible will do waivers

olofk:
I agree with all that @imphil said. Also, make sure that the test suite still passes. Didn't see any updates to the reference files

olofk:
This one needs to be rebase. Also think that tests still needs to be updated

imphil:
Actually #139 did mostly the same thing which we already merged, so this one can be closed.

PR #131, opened 2020-04-10
Proposal for fixes in ghdl.py

ThomasHornschuh:
With reference to the discussion here:
https://github.com/olofk/fusesoc/issues/383

I propose the following changes to ghdl.py.

Set `std=93c` (ghdl default) when there is no vhdl version specific
file_type set.

Pass analyze options to the -r command to be compatible with ghdl mcode
backend.
According to ghdl doc the syntax of the -r command is:
`ghdl -r <[options...] primary_unit [secondary_unit] [simulation_options...]>`
The first set of options are the analyze/elaboration options. They are relevant with mcode because elaboration and simulation is run in one step.


olofk:
Thanks. I have lost count now of how many times the syntax for calling ghdl have been changed in this backend. Hopefully this is the last time we fix it :)

I would like you to move the `#added TH ` comment though, and make sure to update the test cases to that the test suite passes. The easiest way is probably to run `GOLDEN_RUN=1 pytest` and see which reference files that needs to be updated

ThomasHornschuh:
Well, I think the confusion comes from the different backends (mcode, gcc, llvm) ghdl supports and the resulting differences in workflow. In addition command line options have changed over time. In the past --elab-run was the "backend agnostic" option to compile & run a simulation. I think it is depreciated, not sure, only Tristan may know :-)

I think the current flow in this PR will work with all variants. But before continuing I would like to raise the more generic question if deducing the --std option from the Filetype without any hard override option. This looks fragile to me. A possible work around would be to check if the --std option was passed in ghdl analyze_options and in this case switch of the deduction. 

What do you think?

olofk:
I think a manual override is a good idea and checking if it's in `analyze_option` seems like a solid plan

GCHQDeveloper560:
I used GHDL for the first time in a while yesterday and ran across the crash when all of the files are `vhdlSource` rather than specifying a version. I believe this isn't caught by the test because it has a mix of VHDL versions in the test files. It might be nice to test with a set of `vhdlSource` files and a mix for better coverage.

For cases where all the files are `vhdlSource` rather than specifying a version perhaps it would be best to just leave off the `--std` argument and let GHDL use the default.

I also noticed that unless I'm doing something wrong GHDL doesn't allow the use of 1/0 for setting boolean generics. Perhaps this has been a change, or just hasn't worked before. We already cover this possibility for other tools, so changing
```
extra_options += ' -g{}={}'.format(k,self._param_value_str(v,'"'))
```
to
```
extra_options += ' -g{}={}'.format(k,self._param_value_str(v,'"',bool_is_str=True))
```
should fix it.

Are you able to check this and include the fix in this PR, or would you like me to do it separately later?

ThomasHornschuh:
@GCHQDeveloper560 
> For cases where all the files are vhdlSource rather than specifying a version perhaps it would be best to just leave off the --std argument and let GHDL use the default.

I just tried this, but it is not that simple. The "std" determination is not only used for setting the std option, it is also used for creating a dependency to the ghdl library database  (work<std>.cf) file in the generated Makefile:

```

$(TOPLEVEL): $(VHDL_SOURCES) work-obj93c.cf
	ghdl -m $(STD) $(ANALYZE_OPTIONS) $(TOPLEVEL)
```
In ghdl.py:
```
makefile.write(MAKEFILE_TEMPLATE.format(std=' '.join(stdarg),
                                                    toplevel=self.toplevel,
                                                    vhdl_sources=vhdl_sources,
                                                    >>>standard=stdarg[0].split("=")[1],<<<<
                                                    analyze_options=analyze_options,
                                                    run_options=' '.join(run_options),
                                                    make_libraries_directories=make_libraries_directories,
                                                    ghdlimport=ghdlimport))
```
the line marked with >>>><<<< is the one failing when stdarg is an empty list. A possibility is to leave that out, because the library database is not really a dependency, it is part of the output artifacts. 
Even if running the same makefile several times (e.g when using fusesoc with --no-export and --setup options during development) usually source files are updated and the rule is triggered. No idea about a use case where the work file is newer than the sources. 

The implementation is also incomplete, according to the documentation:
https://ghdl.readthedocs.io/en/latest/using/ImplementationOfVHDL.html?highlight=library%20database#library-database

the library database name will also change when using the --work option in ghdl. So we can either invest a lot of time (and test cases) to make this working robust in all cases our just leaf out the dependency in the makefile. I would vote for the later, because the value of it is unclear and the effort relatively high.

GCHQDeveloper560:
It definitely seems like it's more trouble than it's worth to not set a standard. Sticking with your `93c` default sounds fine to me.

ThomasHornschuh:
Hi, I finally managed to complete this PR and also update the tests.

The new solution has the following features:
* When no VHDL version is specified by the source file_type attributes the --std=93c option is used
* When --std=xx is specified in the  analyze_options for the ghdl tool settings, it is overriding any determination by file_type and a warning is emitted, like this:
````
WARNING: Analyze option --std=08 given, will override any vhdlSource-xxxx specification
````
Technically I move the --std option from the ANALYZE_OPTIONS variable to the STD variable in the Makefile,.
Also the work-obj<xx>.cf file name is changed. This is the part of the Makefile I did not like so much at all.
Nevertheless it works, https://github.com/ThomasHornschuh/fusesoc_ghdl_test 
contains a few variants of a project in different branches checking not only the override method, it also checks the handling of VHDL libraries and the --work / --workdir options. 

To cover the different variants in the test suite I have extended  test_ghdl.py with two no tests.
* test_ghdl_01 is the "old" test
* test_ghdl_02 covers the no version in file_type specified case
* test_ghdl_03 covers the override mechanism

I had do do some trickery (changing backend.files) to conduct the test, the standard test mockup always run with the same fileset and therefore cannot cover the special cases

Hope that the PR is now in a state that it can be integrated

ThomasHornschuh:
Updated this PR again, to adapt to  #174 which was merged in the meantime.
I really hope that this PR can be merged before it is outdated again.

olofk:
Hi @ThomasHornschuh. Sorry for the long wait. Your fix had already become outdated, so I untangled the squashed the commits. But I wasn't sure I got it right, so if you could please just look at [this branch](https://github.com/olofk/edalize/tree/ghdl_fix) to make sure I got it right

ThomasHornschuh:
Hi @olofk,
looks good. I have manually cross-checked the tests and have run "real" simulations on a bunch of my cores, also the ones which dumped with the old version.
Can be merged 👍 

Thank You !!!

olofk:
Perfect! Picked and pushed. Thank you for the contribution!

PR #134, opened 2020-04-13
Relax pinning of Jinja2 dependency

imphil:
Jinja 2.11.2 fixed https://github.com/pallets/jinja/issues/1138 which
prevented us from using the 2.11 line, we therefore can relax the version
constraint.


sharkcz:
LGTM

PR #135, opened 2020-04-13
Fix golden test references

imphil:
Two golden test references currently fail on master, fix them. (And yes, we really need CI for this repo.)

PR #136, opened 2020-04-13
Add GitHub Actions-based CI

imphil:
A simple CI setup which runs pytest in tox to also test the packaging of
edalize, catching common errors such as forgetting to add templates in
the manifest files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,30 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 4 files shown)


imphil:
Test results are available at https://github.com/olofk/edalize/runs/583962072?check_suite_focus=true as preview and will fail until #135 is fixed.

olofk:
Is this one good to go? I don't really feel I have the skills to judge so please commit when you're ok with it

imphil:
It will probably fail first, but that's a follow-up once we have the infrastructure in place.

PR #137, opened 2020-04-14
Add VHDL generic support for xsim

GCHQDeveloper560:
From olofk/fusesoc#383


olofk:
Thanks! Not much to comment here. All good

Issue #123, opened 2020-03-12
Cocotb support


gchqdev1729:
My team is starting to make use of FuseSoC. We have made extensive use of Cocotb for our testbenches, so we're looking for a good way to integrate the two. We have some workarounds that use the ModelSim backend, but they aren't ideal and don't really use the power of FuseSoC.

I've been chatting to my colleague @GCHQDeveloper992 and we think that one approach could be to write an Edalize backend for Cocotb. It would generate a Cocotb makefile and then call make and allow Cocotb to handle the simulator abstraction. This would be rather like the Quartus backend and it feels like the path of least resistance.

I've seen some comments on issues in FuseSoC and Cocotb projects that suggest others have had some thoughts about this too. (In particular @imphil.)

Here are a few of my thoughts to open a discussion. I'm keen to get something working for us fairly soon one way or another. We have time to do some development to put something together.

### Example config

The Cocotb backend could be configured by quite a simple a tools section, something like this:

```
tools:
    cocotb:
        module: test_my_top_level.py
        simulator: modelsim

```
As the top level entity and its language are already known by FuseSoC/Edalize I hope it shouldn't be necesary to set them here explicitly, but that's an option.

### Setting PYTHONPATH

Our testbenches use Python models of the components. They build on each other following the same hierarchy of dependencies as the HDL. That means PYTHONPATH needs to contain both the modules for the core at hand and the modules relating to dependencies. I wondered about doing this by adding Python files to the filesets, so they get copied in the usual way, and having the backend collect a list of the parent directories for addition to the path.

That works for our use case because our Python code isn't arranged in packages. I.e. adding the parent directory of each source file is exactly what we want to do. However, that isn't going to be universally correct. Maybe there's a better approach to this.

### Virtual Environments

I'm in two minds about Virtual Environments. Maybe it's something to think about later when we have something working.

The motivation is that we have several testbenches that depend on having particular versions of Python libraries installed, both open source and internal. Up to now, we've managed that by having our build system set up a virtual environment before running Cocotb.

It would be possible to provide an option to set up a virtual environment in the backend before running Cocotb. One disadvantage is that we have to pip install Cocotb again, with the associated recompile of the simulator libraries (library compilation on install in the latest Cocotb is great, thanks!). Anther is increased complexity. The advantage is that the core file can then manage all the dependencies and no wrappers or external config are necessary, which is attractive to us.

[...]

PR #138, opened 2020-04-17
Add basic cocotb

gchqdev1729:
Following up on my issue, #123, I'm putting this cocotb backend up for comment and review. It implements much of cocotb's capabilities, but not all. Also, I decided not to treat virtual environments in this implementation. Maybe that's something to be added later or not at all.

I've been using the latest cocotb master for testing, so this probably only supports version 1.4 onwards.

_I'm busy next week, so if there's a delay in replying to any comments, apologies._

## Limitations

I've added verilog include directory support in a slightly hacky way. If possible, it would be better if cocotb handled the simulator specific behaviour, but it works for now. I might discuss with cocotb whether I can make a patch for this.

For similar reasons, I have not implemented parameters, defines or generics, which is a shame. cocotb doesn't abstract the simulator specific syntax, so I've left these to be added into the compile_args, sim_args or whatever as appropriate. Another potential patch for cocotb.

If I've missed a trick here, let me know...

## Testing

I've created a set of test cores here: https://github.com/gchqdev1729/fusesoc-cores. 

Three of them run cocotb examples. I can't run the 'mean' example to the end because I don't currently have access to 64-bit ModelSim and it requires mixed HDL. The vsim line it generates looks good, but it'd be good to know it actually runs! The 'adder' example provides targets for VHDL and verilog.

The other three cores test the hierarchical Python model paradigm that was my main use case for developing this backend.


themperek:
>Limitations

Something to consider is running cocotb by just adding calls to VPI/VHPI/FLI in exist simulators integrations and settings few environmental variables rather than makefile generator. Some examples here: https://github.com/cocotb/cocotb/pull/1562. This way it would allow using all existing simulator infrastructure in edalize.

m-kru:
@gchqdev1729 Can you provide some [FuseSoc](https://github.com/olofk/fusesoc) `*.core` file example with target defined with cocotb backend? I would like to test this pull request.

gchqdev1729:
> Something to consider is running cocotb by just adding calls to VPI/VHPI/FLI in exist simulators integrations and set few environmental variables rather than makefile generator (this is easy now/v1.4). Some examples here: [cocotb/cocotb#1562](https://github.com/cocotb/cocotb/pull/1562). This way it would allow using all existing simulator infrastructure in edalize.

Thanks - that sounds good. You did mention that some helpful new things were on the way in 1.4, but I missed this aspect...

gchqdev1729:
> @gchqdev1729 Can you provide some [FuseSoc](https://github.com/olofk/fusesoc) `*.core` file example with target defined with cocotb backend? I would like to test this pull request.

Thanks for looking at this and apologies for the delay in replying. I linked some example cores in the associated issue (#123 ), but didn't put them in the PR. (New to Github!) Here they are: https://github.com/gchqdev1729/fusesoc-cores

themperek:
Some more information https://github.com/cocotb/cocotb/issues/1718

gchqdev1729:
@themperek 

I've been looking at the alternative method you suggest. It certainly gets around the limitations I listed. I like the option of bypassing make too.

There is a potential downside as well, I think. There's going to be a certain amount of simulator specific code to add to this edalize backend to add the correct VPI/VHPI/FLI module and configuration, plus some additional arguments. Comparing the ModelSim and RivieraPro implementations, for example, the Cocotb makefiles add VLOG_ARGS and ALOG_ARGS respectively. Maybe this isn't too much of a problem? From my point of view, it'll make testing more challenging because I can't test against the full set of simulators - the makefile approach has the advantage of abstracting that away for me.

I think this is probably worth pursuing and that the advantages win here. It always feels a bit nasty to duplicate code/knowledge in two code-bases, but I suppose we're already dealing with two tool abstractions here. Perhaps by implementing the VPI/VHPI/FLI abstraction here it might help pave the way to cocotb using edalize as a target itself, as I think you suggest here: https://github.com/cocotb/cocotb/pull/1562#issuecomment-615902345. Testing is perhaps my main concern.

Another point worth mentioning, though not a great problem, is that cocotb supports some simulators that edalize does not. Edalize doesn't support ActiveHDL, CVC, IUS or Xcelium. If people care about those simulators and want to use this backend, then adding support to edalize is probably the way forward for them.

One option would be to continue with this PR with its limitations to get some cocotb support in edalize. Then I can work on the alternative route as an improvement in a following change. Otherwise, I could hold off with this PR until I've worked on the alternative in the next few weeks.

Any thoughts?

themperek:
@imphil Any opinion?

m-kru:
> From my point of view, it'll make testing more challenging because I can't test against the full set of simulators - the makefile approach has the advantage of abstracting that away for me.

Can you elaborate on this? I think I do not understand what is the problem that disappears with makefile approach.

> Another point worth mentioning, though not a great problem, is that cocotb supports some simulators that edalize does not. Edalize doesn't support ActiveHDL, CVC, IUS or Xcelium. If people care about those simulators and want to use this backend, then adding support to edalize is probably the way forward for them.

In my opinion it is not a problem, at all. Edalize does not have to directly support simulator used by cocotb. Edalize calls cocotb, cocotb calls proper simulator.

> One option would be to continue with this PR with its limitations to get some cocotb support in edalize. Then I can work on the alternative route as an improvement in a following change. Otherwise, I could hold off with this PR until I've worked on the alternative in the next few weeks.

In my humble opinion, it is better to drop this PR as it is not merged yet, and work on solution that is already known to be better one. Continuing with this PR will be a waste of time.

gchqdev1729:
> > From my point of view, it'll make testing more challenging because I can't test against the full set of simulators - the makefile approach has the advantage of abstracting that away for me.
> 
> Can you elaborate on this? I think I do not understand what is the problem that disappears with makefile approach.

Sure. 

With the makefile approach, my testing is ensuring that the environment variables are set correctly and that I've included the cocotb makefiles. Adding the VPI/FLI options is handled by cocotb and I can assume that works because it's been tested in cocotb.

With the alternative approach, I'm going to be creating arguments in this backend that I pass to another edalize backend for inclusion on the simulator run command. For example for Modelsim I'm going to be setting `-pli` or `-foreign` with a library name and potentially extra options for mixed language tests (see https://github.com/cocotb/cocotb/blob/776a24a448d1c7162fedc7c2f736613efcdcd0f0/cocotb/share/makefiles/simulators/Makefile.questa#L70). These arguments are different for each simulator - and not exactly straightforward in some cases. It'd be easy for me to make a mistake in generating them and I'd have no way of testing them without access to the simulator in question.

Maybe I'm worrying unduly or perhaps there's another approach to testing?

> 
> > Another point worth mentioning, though not a great problem, is that cocotb supports some simulators that edalize does not. Edalize doesn't support ActiveHDL, CVC, IUS or Xcelium. If people care about those simulators and want to use this backend, then adding support to edalize is probably the way forward for them.
> 
> In my opinion it is not a problem, at all. Edalize does not have to directly support simulator used by cocotb. Edalize calls cocotb, cocotb calls proper simulator.
> 

That's true for my current approach because cocotb handles the simulator abstraction. However, my proposal for the alternative approach is that this new backend generates the appropriate args to add the VPI/FLI. Then it calls the edalize backend for the simulator, which handles the input files, verilog includes, parameters, generics and so on. I think that's what @themperek was suggesting and it seems a good approach because it makes more use of the power of edalize. With this approach, only simulators supported by both edalize and cocotb will work.

Apologies if I have not been clear in this. It probably would have been worth me outlining this for everyone's benefit in my earlier reply.

m-kru:
@gchqdev1729 ok, now I understand. Maintaining the same knowledge in more than one repository is never a good idea. The only acceptable way I can see for this alternative approach is `cocotb` API call allowing you to fetch all parameters needed for particular simulator. What is more, your cocotb backend would need to call another backend from edalize. I think right now there is no such scenario in edalize. Maybe it would be a good idea to also include @olofk to this discussion.

imphil:
Just a quick heads-up: I've seen this discussion and will get back to it this week hopefully.

gchqdev1729:
> Maintaining the same knowledge in more than one repository is never a good idea. The only acceptable way I can see for this alternative approach is `cocotb` API call allowing you to fetch all parameters needed for particular simulator. 

I agree. An API call to provide the abstraction in cocotb is an interesting option.

> What is more, your cocotb backend would need to call another backend from edalize. I think right now there is no such scenario in edalize. Maybe it would be a good idea to also include @olofk to this discussion.

I admit I hadn't actually tried to do this yet, it might not be possible... I'd hoped it might be possible for this backend to: add to the options passed to it in edam; instantiate the appropriate simulator backend; pass calls to configure*, build*, run* methods to the simulator object.

I'll wait to see what @olofk and @imphil have to say. Quite a few ways we could take this, I think.

ktbarrett:
From the thread in the cocotb repo, it seems that cocotb is interested in removing the makefiles entirely. All simulator-specific bits should end up in this repo (or any other supported regression framework). We don't want to provide additional abstraction as it's an unnecessary scope increase. Isn't that the duty of this project, to abstract away simulator-specific stuff? The actual requirements by cocotb are pretty low, so this shouldn't be too difficult. 

You probably don't have to worry about supporting Questa or Riviera with this PR because they are a bit different and will require more thought. I would focus on GHDL, Icarus, and Verilator, since that's what cocotb can test in CI right now. Once they are supported, we can asses this as a solution to our goal of replacing the makefiles and see if we want to press forward in this direction.

But these are my opinions. Would like to see what @eric-weiser has to say as well.

edalize/cocotb.py
```diff
@@ -0,0 +1,141 @@
...
+                         'type': 'String',
+                         'desc': 'Any arguments or flags to pass to the execution of the compiled simulation'},
+                        {'name': 'run_args',
+                         'type': 'String',
+                         'desc': 'Any argument to be passed to the “first” invocation of a simulator that runs via a TCL script'},
+                        {'name': 'extra_args',
+                         'type': 'String',
+                         'desc': 'Passed to both the compile and execute phases of simulators with'
+                         + ' two rules, or passed to the single compile and run command for simulators'
+                         + ' which don’t have a distinct compilation stage'},
```
eric-wieser, line 57: nit: faster to use parse-time concatenation:
```suggestion
                         'desc': 'Passed to both the compile and execute phases of simulators with '
                                 'two rules, or passed to the single compile and run command for simulators '
                                 'which don’t have a distinct compilation stage'},
```

tests/edalize_common.py
```diff
@@ -80,28 +80,30 @@ def setup_backend(paramtypes, name, tool, tool_options, use_vpi=False):
...
+    {'name' : 'vhdl2008_file'              , 'file_type' : 'vhdlSource-2008'},
+    {'name' : 'xci_file.xci'               , 'file_type' : 'xci'},
+    {'name' : 'xdc_file.xdc'               , 'file_type' : 'xdc'},
+    {'name' : 'bootrom.mem'                , 'file_type' : 'mem'},
+    {'name' : 'c_file.c'                   , 'file_type' : 'cSource'},
+    {'name' : 'cpp_file.cpp'               , 'file_type' : 'cppSource'},
+    {'name' : 'c_header.h'                 , 'file_type' : 'cSource', 'is_include_file' : True},
+    {'name' : 'c_header.h'                 , 'file_type' : 'cppSource', 'is_include_file': True},
+    {'name' : 'python_dir/python_file.py'  , 'file_type' : 'pythonSource'},
+    {'name' : 'python_dir2/python_file2.py', 'file_type' : 'pythonSource'},
```
eric-wieser, line 106: What's the effect of adding these here? Is this just for `PYTHON_PATH` population? Or does it also use it to track stale dependencies?

Not all python files should be added to the path, for instance `model/some_helper.py` might be intended to be imported as `import model.some_helper`.

gchqdev1729: Thanks for the comments.

The intention of having the python files in this module is to test that the PYTHONPATH is correctly populated. I'd considered the use case that you suggest and thought that it does not actually break that. (Maybe I'm wrong?)

I'd thought of a couple of ways of effecting the construction of a PYTHONPATH. The solution needs to work when FuseSoC cores provide a hierarchy of models as well as a hierarchy of RTL (example here https://github.com/gchqdev1729/fusesoc-cores/blob/master/gchqdev1729-add_and_invert.core). The two solutions I came up with were passing path components (i.e. directories) or passing the python files as here.

I think passing path components breaks down in a couple of ways. Firstly, FuseSoC explicitly doesn't allow directories in its file list just now. Secondly, it does not ensure that the files are copied from the provider into the FuseSoC workspace where Edalize is going to work on them, so it's unclear how they would be made available at runtime. 

In our own projects, we do have some external dependencies, both open source and proprietary and they would be imported exactly as you suggest (i.e by `import mypackage.mymodule`).

Hope that makes sense...

eric-wieser:
I think I agree with @ktbarrett here. Whether we drop the makefiles or not, their content is an attempt to abstract over simulators, which presumably is already the goal of edalize.

> Maintaining the same knowledge in more than one repository is never a good idea.

The amount of knowledge being maintained here is rather tiny though:
1. The names of the appropriate `lib` file
2. The fact that cocotb uses the `MODULE`, `TESTCASE`, `TOPLEVEL`, and `TOPLEVEL_LANG` variables

I suppose one way to resolve this would be to put the cocotb edalize support within cocotb itself. That would have the benefit of ensuring that if our interface changes, the eda hook updates with it.

tests/edalize_common.py
eric-wieser: In your example file, how am I suppose to import `add_and_invert/model/model_add_and_invert.py`?

1. `import add_and_invert.model.model_add_and_invert` &rarr; `.` should be in `sys.path`
2. `import model.model_add_and_invert` &rarr; `./add_and_invert` should be in `sys.path`
3. `import model_add_and_invert` &rarr; `./add_and_invert/model` should be in `sys.path`

Putting all three on the path is a very bad idea, as then you have multiple distinct copies of the same module depending on which path you import.

gchqdev1729:
> I think I agree with @ktbarrett here. Whether we drop the makefiles or not, their content is an attempt to abstract over simulators, which presumably is already the goal of edalize.
> 
> > Maintaining the same knowledge in more than one repository is never a good idea.
> 
> The amount of knowledge being maintained here is rather tiny though:
> 
> 1. The names of the appropriate `lib` file
> 2. The fact that cocotb uses the `MODULE`, `TESTCASE`, `TOPLEVEL`, and `TOPLEVEL_LANG` variables
> 
> I suppose one way to resolve this would be to put the cocotb edalize support within cocotb itself. That would have the benefit of ensuring that if our interface changes, the eda hook updates with it.

Thanks @eric-wieser and @ktbarrett for looking.

> I think I agree with @ktbarrett here. Whether we drop the makefiles or not, their content is an attempt to abstract over simulators, which presumably is already the goal of edalize.

Yes, I completely agree with this.

> 
> > Maintaining the same knowledge in more than one repository is never a good idea.
> 
> The amount of knowledge being maintained here is rather tiny though:
> 
> 1. The names of the appropriate `lib` file
> 2. The fact that cocotb uses the `MODULE`, `TESTCASE`, `TOPLEVEL`, and `TOPLEVEL_LANG` variables
> 
> I suppose one way to resolve this would be to put the cocotb edalize support within cocotb itself. That would have the benefit of ensuring that if our interface changes, the eda hook updates with it.

It is a small amount of knowledge, I agree. 

It seems like the right thing to do is to close this PR and try something along the lines of my comment (https://github.com/olofk/edalize/pull/138#issuecomment-621188800), as an experiment at least. That way I can discover how well that method works and what specific knowledge is required.

My concern about testing the command lines I generate remains though (https://github.com/olofk/edalize/pull/138#issuecomment-620731481). As a way to mitigate this, I liked @m-kru 's idea above whereby cocotb provides an API call (maybe part of cocotb-config) that provides the appropriate arguments to add to the simulator command. If cocotb used the same code to generate its arguments, then it would implicity be tested by cocotb. It also gets around problems when people have different versions of cocotb installed.

gchqdev1729:
As discussed, I'm going to close this PR and work on an alternative where Edalize does the simulator abstraction rather than delegating to cocotb. It looks like this will use more of the power of Edalize and might assist with the discussion on the cocotb project:  https://github.com/cocotb/cocotb/pull/1562#issue-398661092. 

Thanks to all who have commented.

m-kru:
@eric-wieser 

> I suppose one way to resolve this would be to put the cocotb edalize support within cocotb itself. That would have the benefit of ensuring that if our interface changes, the eda hook updates with it.

Can you elaborate on this? How would this work with [FuseSoc](https://github.com/olofk/fusesoc)? From my point of view, the work will be finished, when I will be able to define cocotb target in  FuseSoc `.core` and run this target automatically on the server as a regression test.

tests/edalize_common.py
gchqdev1729: OK, I understand. In our code, we have historically used option 3, so what I'm doing in my branch works, but I can see the problem.

I'll have a think to see if there is another solution. I think I need to add all the Python files to the FuseSoC core files to ensure that all the required files are copied to the workspace and made accessible in the file list passed to Edalize. Then I need to construct the PYTHONPATH in some way. I suppose there might be a way to use file attributes to add an ancestor directory other than the direct parent. Beyond that, unless I've missed something, I think it will require wider changes to FuseSoC.

eric-wieser: You're probably right that this is an issue with FuseSoC itself, and so probably isn't necessary to address immediately - although a comment drawing attention to the shortcoming would be nice.

What happens if you omit some files from FuseSoC? You might be able to list only a root `__init__.py` or `__init__.pyi`, and use that to construct the PYTHON_PATH.

gchqdev1729: Sadly, this wouldn't work as it is because __init__.py would get copied, but not the modules. FuseSoC passes the path of the copied files to Edalize, not where they were copied from, so the module files themselves wouldn't be in the path as constructed.

gchqdev1729: I will document this, as suggested though.

eric-wieser: Perhaps the python path should be set via some other means then, rather than inferring it from the filenames.

gchqdev1729: I'm pretty sure that the only thing a FuseSoC core inherits from its dependencies is the list of files. For my hierarchical model use case, I need the info to propagate up, so I think this is the only way without wider changes. (I'm a relative novice when it comes to FuseSoC though, so I might be proved wrong!)

eric-wieser:
I was hoping that `FuseSoc` had some sort of plugin mechanism, and the config itself could live in cocotb.

> From my point of view, the work will be finished, when I will be able to define cocotb target in FuseSoc .core and run this target automatically on the server as a regression test.

Indeed, what you'd do is run a regression test against the last release of cocotb. In turn, we'd run a regression test of the latest cocotb against the last release of FuseSoc.

m-kru:
FuseSoc uses edalize as backend for running targets. I can't see how it would work, when cocotb would support edalize instead of edalize supporting cocotb.

This is of course not so obvious, as cocotb has its own way for running EDA tools and edalize main goal is to provide abstraction for running different EDA tools. cocotb is a bit like simulator, that needs other simulators to run. This is very peculiar. Maybe cocotb should use edalize instead of Makefiles. Then calling chain for cocotb FuseSoc targets would look like:
```
FuseSoc -> edalize -> cocotb -> edalize
```

> Indeed, what you'd do is run a regression test against the last release of cocotb. In turn, we'd run a regression test of the latest cocotb against the last release of FuseSoc.

What I had in mind were regressions tests for my project/design, not for cocotb or FuseSoc. This is different story.

PR #139, opened 2020-04-17
[veriblelint] Add option to include config file via source files

msfschaffner:
This adds support for specifying the `veriblelint` configuration file as part of the design sources, similarly to the `ascentlint` target. This makes it possible to create a common core file that other core files can depend on. Otherwise, we would have to hardcode the path to the config file in each core file with a lint target, which is a bit brittle and cumbersome to maintain.

This implementation throws an error if more than one file of type `veribleLintConfig` is discovered in the sources. Further, if the `rules_config_file` option is specified in a core file, that option will override the config file discovered as part of the sources, and a WARNING will be printed that this has occurred.

Signed-off-by: Michael Schaffner <msf@google.com>

### DIFF
--- a/edalize/veriblelint.py
+++ b/edalize/veriblelint.py
@@ -19,9 +19,6 @@ class Veriblelint(Edatool):
                          {'name': 'ruleset',
                           'type': 'String',
                           'desc': 'Ruleset: [default|all|none]'},
-                         {'name': 'rules_config_file',
-                          'type': 'String',
-                          'desc': 'Path to the rule config file'}
                     ],
                     'lists': [
                          {'name' : 'verible_lint_args',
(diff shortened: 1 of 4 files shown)


msfschaffner:
@imphil, @olofk Let me know if you think we should handle this differently - this is just a suggestion. 
It would however be great if we had something like this in place.

edalize/veriblelint.py
```diff
@@ -54,16 +54,27 @@ def run_main(self):
...
+        config_files_filtered = []
         for src_file in src_files:
             ft = src_file.file_type
-            if not ft.startswith("verilogSource") and not ft.startswith("systemVerilogSource"):
-                continue
-            src_files_filtered.append(src_file.name)
+            if ft.startswith("verilogSource") or ft.startswith("systemVerilogSource"):
+              src_files_filtered.append(src_file.name)
+            elif ft.startswith("veribleLintConfig"):
+              config_files_filtered.append(src_file.name)
```
imphil, line 63: four source indent please.

imphil:
@kgugala FYI. Can you and @msfschaffner work together to get one of your PRs across the finish line?

edalize/veriblelint.py
msfschaffner: woops...

msfschaffner:
Thanks @imphil. I took the initial commit from @kgugala and added the multiple config file check. Also, I named the file type `veribleLintRules` since `linterRules` feels a bit too generic (there exist other linters for which we might add a rule file in the future).

imphil:
Looks good, thanks!

PR #141, opened 2020-04-23
Add section 'Testing edalize' to docs

cmarqu:
Add tests/README.rst which is pulled into documentation as well,
add docstrings in edalize_common.py.
Add some general comments in mock_commands/.


olofk:
Super! Many thanks for this. Picked and pushed

Issue #144, opened 2020-04-27
Go Python 3.5+-only


imphil:
Python 2.7 has reached EOL earlier this year, and fusesoc has removed support for it and is Python 3.5-only now. The relevant discussion is in https://github.com/olofk/fusesoc/issues/337, most arguments apply to edalize as well.

As noted before: nothing prevents users on Python 2.7 to continue using edalize 0.2. This change only affects people going forward, but does not break existing setups.

`pypistats python_minor edalize`

```
| category | percent | downloads |
|----------|--------:|----------:|
| 3.7      |  74.87% |    24,014 |
| 3.5      |  12.37% |     3,966 |
| 3.6      |   9.08% |     2,912 |
| 2.7      |   1.54% |       494 |
| null     |   1.14% |       367 |
| 3.8      |   0.98% |       314 |
| 3.4      |   0.02% |         7 |
| Total    |         |    32,074 |

Date range: 2019-10-29 - 2020-04-26
```

olofk:
Yes, let's drop it

GCHQDeveloper560:
I've added this to #143 and it now has the magical green check mark.

PR #143, opened 2020-04-27
Correct tests for Python 3.5

GCHQDeveloper560:
#132

PR #145, opened 2020-04-29
Add symbiflow support

acomodi:
This PR collects all the changes needed to run https://github.com/SymbiFlow/fpga-tool-perf.

Main changes are:

- addition of symbiflow tool that uses Yosys
- addition of nextpnr-xilinx tool
- changes on how the subprocesses are called during run_tool to have a silent build, as well as output the whole run log when an error occurs.
- some fixes to yosys subtool


acomodi:
@olofk FYI, this is the PR mentioned [here](https://github.com/SymbiFlow/edalize/pull/28#issuecomment-620871075)

edalize/edatool.py
```diff
@@ -104,6 +104,15 @@ def get_doc(cls, api_ver):
...
             logger.warning("Invalid API version '{}' for get_tool_options".format(api_ver))
 
+    def check_args(self, unknown):
+        # If a tool is using subtools some of the argument may be
+        # parsed by the subtool. This function is used to check if
+        # all the provided arguments were correct. A tool can override
+        # this function to provide custom args checking logic.
+
+        if unknown:
+            raise Exception(f'Unknown command line option {unknown[0]}')
```
GCHQDeveloper560, line 114: From #144 we plan to support Python 3.5, so I'm afraid you'll have to use `.format` instead of f-strings. Also, if this is an abstract method would just raising `NotImplementedError` if it's called be the most informative exception?

edalize/nextpnr.py
```diff
@@ -0,0 +1,126 @@
...
+        nextpnr_args = []
+        nextpnr_args.append(a['name'] for a in nextpnr_members)
+        nextpnr_args.append(a['name'] for a in nextpnr_lists)
+
+        for arg in args:
+            if not arg.startswith('-'):
+                continue
+            argname = arg.strip('-')
+            if argname not in nextpnr_args:
+                raise Exception(f'Unknown command line option {arg}')
```
GCHQDeveloper560, line 60: Another f-string slipped in

edalize/nextpnr.py
```diff
@@ -0,0 +1,126 @@
...
+                'name'          : self.name,
+                'toplevel'      : self.toplevel,
+                'parameters'    : self.parameters,
+                'tool_options'  : {'yosys' : {
+                                        'arch' : 'xilinx',
+                                        'yosys_synth_options' : yosys_synth_options,
+                                        'yosys_as_subtool' : True,
+                                        }
+                                }
+                }
```
GCHQDeveloper560, line 86: Did the indentation go wonky here? I believe FuseSoC is using black, so it may be worth considering using it for new files like this.

edalize/nextpnr.py
```diff
@@ -0,0 +1,126 @@
...
+
+        for f in src_files:
+            if f.file_type in ['synthJson']:
+                synth_design = f.name
+            elif f.file_type in ['bba']:
+                chipdb = f.name
+            elif f.file_type in ['xdc']:
+                xdc = f.name
+            else:
+                continue
```
GCHQDeveloper560, line 110: The documentation in `doc/edam/api.rst` has attemped to list file types supported by Edalize that aren't part of the IP XACT standard. I believe XDC is already included, but you may want to consider including the others.

edalize/symbiflow.py
```diff
@@ -0,0 +1,176 @@
...
+import re
+import subprocess
+
+from edalize.edatool import Edatool
+from edalize.yosys import Yosys
+from importlib import import_module
+
+logger = logging.getLogger(__name__)
+
+""" Symbiflow backtend
```
GCHQDeveloper560, line 13: ~~backtend~~ backend 😄

edalize/symbiflow.py
```diff
@@ -0,0 +1,176 @@
...
+                             'Makefile',
+                             makefile_params)
+
+    def configure_vpr(self):
+        (src_files, incdirs) = self._get_fileset_files(force_slash=True)
+
+        has_vhdl     = 'vhdlSource'      in [x.file_type for x in src_files]
+        has_vhdl2008 = 'vhdlSource-2008' in [x.file_type for x in src_files]
+
+        assert (not has_vhdl and not has_vhdl2008), 'VHDL files are not supported in Yosys'
```
GCHQDeveloper560, line 109: I don't think you need to check the files twice, and an assertion is probably not correct for incorrect user input. Perhaps the following?

```suggestion
        has_vhdl = [x.name for x in src_files if x.file_type.startsWith("vhdlSource")

        if has_vhdl:
            logger.error("VHDL files are not supported in Yosys: {}".format(", ".join(has_vhdl)))
```

edalize/symbiflow.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # copy user files to builddir
+        for f in user_files:
+            shutil.copy(f, self.work_root + "/" + builddir)
+
+        part = self.tool_options.get('part', None)
+        package = self.tool_options.get('package', None)
+
+        assert part is not None, 'Missing required "part" parameter'
+        assert package is not None, 'Missing required "package" parameter'
```
GCHQDeveloper560, line 138: I don't think assertions are the right approach for checking user input. I would use `logger` instead.

edalize/symbiflow.py
```diff
@@ -0,0 +1,176 @@
...
+        nextpnr = getattr(import_module("edalize.nextpnr"), 'Nextpnr')(nextpnr_edam, self.work_root)
+        nextpnr.configure()
+
+        builddir = self.tool_options.get('builddir', 'build')
+
+        part = self.tool_options.get('part', None)
+        package = self.tool_options.get('package', None)
+
+        assert part is not None, 'Missing required "part" parameter'
+        assert package is not None, 'Missing required "package" parameter'
```
GCHQDeveloper560, line 81: I would use `logger` instead of assertions for checking user input.

edalize/vivado.py
```diff
@@ -72,9 +80,33 @@ def configure_main(self):
 
         self.jinja_env.filters['src_file_filter'] = self.src_file_filter
 
+        has_vhdl     = 'vhdlSource'      in [x.file_type for x in src_files]
         has_vhdl2008 = 'vhdlSource-2008' in [x.file_type for x in src_files]
         has_xci      = 'xci'             in [x.file_type for x in src_files]
 
+        self.synth_tool = self.tool_options.get('synth', 'vivado')
+        if self.synth_tool == 'yosys':
+            assert (not has_vhdl and not has_vhdl2008), 'VHDL files are not supported in Yosys'
```
GCHQDeveloper560, line 89: Another place to use `logging` instead of an assertion.

edalize/symbiflow.py
```diff
@@ -0,0 +1,176 @@
...
+                         'type' : 'String',
+                         'desc' : 'Target architecture. Currently only "xilinx" is supported'},
+                        {'name' : 'pnr',
+                         'type' : 'String',
+                         'desc' : 'Place and Route tool. Currently only "vpr" and "nextpnr" are supported'},
+                   ]}
+
+            symbiflow_members = symbiflow_help['members']
+
+            return {'description' : "The Symbilow backend executes Yosys sythesis tool and VPR place and route. It can target multiple different FPGA vendors",
```
GCHQDeveloper560, line 51: ~~Symbilow~~ Symbiflow

edalize/vivado.py
```diff
@@ -88,21 +120,43 @@ def configure_main(self):
...
+                                 template_vars)
+
+            self.render_template('vivado-run.tcl.j2',
+                                 self.name+"_run.tcl")
+
+            self.render_template('vivado-synth.tcl.j2',
+                                 self.name+"_synth.tcl")
+
+        vivado_settings = self.tool_options.get('vivado-settings', None)
+        vivado_command = "source {} && vivado".format(vivado_settings) if vivado_settings else "vivado"
```
GCHQDeveloper560, line 150: I've always assumed any required environment setup would be done outside of the tool. What's the use case for doing this in edalize?

setup.py
```diff
@@ -29,6 +29,9 @@ def read(fname):
         'templates/trellis/trellis-makefile.j2',
         'templates/ascentlint/Makefile.j2',
         'templates/ascentlint/run-ascentlint.tcl.j2',
+        'templates/symbiflow/symbiflow-makefile.j2',
+        'templates/yosys/yosys-makefile.j2',
+        'templates/yosys/yosys-script-tcl.j2',
```
GCHQDeveloper560, line 34: Did you catch all of the new templates? Weren't there additional Yosys templates for Vivado?

acomodi:
@GCHQDeveloper560 Thanks for the review, I'll go through solving the comments by today

edalize/vivado.py
acomodi: I think that sourcing, at least, the Vivado settings prior to the command invocation is better as it may happen that the environment created by Vivado alters some of the core libraries such as libc.

By doing like this, we can be sure that the environment of the caller is preserved and it is altered only in the time frame in which the command is executed.

edalize/edatool.py
acomodi: Done

edalize/nextpnr.py
acomodi: Done

edalize/nextpnr.py
acomodi: Done, I have used `black` to format vivado.py, symbiflow.py and nextpnr.py

edalize/nextpnr.py
acomodi: Added bba and synthJson to the docs

edalize/symbiflow.py
acomodi: Done

edalize/symbiflow.py
acomodi: Done

edalize/symbiflow.py
acomodi: Done

edalize/symbiflow.py
acomodi: Done

setup.py
acomodi: Templates added

acomodi:
@GCHQDeveloper560 I have addressed the comments, apart from the Vivado source one, which would require some further discussion. If it is an issue to have the source settings in there, I could probably remove it.

I have added also a test that checks vivado_yosys. I believe I should do the same for symbiflow and nextpnr correct?

tests/test_vivado.py
```diff
@@ -64,3 +64,37 @@ def test_vivado_minimal():
     compare_files(ref_dir, work_root, [
         'vivado.cmd',
     ])
+
+def test_vivado_yosys():
```
GCHQDeveloper560, line 68: Are you able to base this off `test_vivado` rather than the minimal test? Not only would it be a more extensive test but the code should be shorter too. In fact I wonder if you could use the [pytest parameterisation support](https://docs.pytest.org/en/latest/parametrize.html) to run `test_vivado` with the two different synth settings?

GCHQDeveloper560:
My main concern was ensuring we don't break existing tools, so the vivado test is welcome. I would definitely add tests for the new tools as well.

tests/test_vivado.py
acomodi: Sure, done.

acomodi:
@GCHQDeveloper560 Sure, I have fixed the vivado tests, as well as icestorm ones which were failing locally (for icestorm the golden results needed to be updated).

I have also added symbiflow tests, which test both VtR and Nextpnr

acomodi:
I do not know why, but CI is failing with the following:

- py 3.6, 3.7 and 3.8:

```
        except subprocess.CalledProcessError as e:
            _s = "'{}' exited with an error code\n{}"
>           raise RuntimeError(_s.format(cmd, e.output.decode()))
E           RuntimeError: '/home/runner/work/edalize/edalize/.tox/py/bin/python' exited with an error code
E           Traceback (most recent call last):
E             File "/tmp/vunit_5_gu75bn/run.py", line 16, in <module>
E               runner = load_runner_hooks().VUnitRunner()
E             File "/tmp/vunit_5_gu75bn/run.py", line 14, in load_runner_hooks
E               return __import__('edalize.vunit_hooks', fromlist=['vunit_hooks'])
E             File "/home/runner/work/edalize/edalize/.tox/py/lib/python3.6/site-packages/edalize/vunit_hooks.py", line 3, in <module>
E               from vunit.ui import Library
E           ModuleNotFoundError: No module named 'vunit'

.tox/py/lib/python3.6/site-packages/edalize/edatool.py:334: RuntimeError
=========================== short test summary info ============================
FAILED tests/test_vunit.py::test_vunit_hooks - RuntimeError: '/home/runner/wo...
========================= 1 failed, 39 passed in 4.36s =========================
ERROR: InvocationError for command /home/runner/work/edalize/edalize/.tox/py/bin/pytest (exited with code 1)
```

- py 3.5: almost all tests fail

```
tests/test_ascentlint.py F                                               [  2%]
tests/test_diamond.py F.                                                 [  7%]
tests/test_edam.py ....                                                  [ 17%]
tests/test_ghdl.py F                                                     [ 20%]
tests/test_icarus.py F.                                                  [ 25%]
tests/test_icestorm.py FFFFFF                                            [ 40%]
tests/test_ise.py F.                                                     [ 45%]
tests/test_isim.py F                                                     [ 47%]
tests/test_modelsim.py F                                                 [ 50%]
tests/test_quartus.py F                                                  [ 52%]
tests/test_rivierapro.py F                                               [ 55%]
tests/test_spyglass.py FF                                                [ 60%]
tests/test_symbiflow.py FF                                               [ 65%]
tests/test_vcs.py FF.                                                    [ 72%]
tests/test_veribleformat.py .                                            [ 75%]
tests/test_veriblelint.py .                                              [ 77%]
tests/test_verilator.py F..                                              [ 85%]
tests/test_vivado.py FFF                                                 [ 92%]
tests/test_vunit.py .F                                                   [ 97%]
tests/test_xsim.py F                                                     [100%]
```

GCHQDeveloper560:
See #143 for Python 3.5. It looks like it was just merged.

Python 3.6 is okay on master, so I'm not sure what might have gone wrong there.

edalize/edatool.py
```diff
@@ -310,15 +321,17 @@ def _run_tool(self, cmd, args=[]):
         logger.debug("args  : " + ' '.join(args))
 
         try:
-            subprocess.check_call([cmd] + args,
+            sp_output = subprocess.check_output([cmd] + args,
```
acomodi, line 324: @GCHQDeveloper560 I discovered that the following error is given by this:

```
>           raise RuntimeError(_s.format(cmd, e.output.decode()))
E           RuntimeError: '/home/runner/work/edalize/edalize/.tox/py/bin/python' exited with an error code
E           Traceback (most recent call last):
E             File "/tmp/vunit_lesqgqxj/run.py", line 16, in <module>
E               runner = load_runner_hooks().VUnitRunner()
E             File "/tmp/vunit_lesqgqxj/run.py", line 14, in load_runner_hooks
E               return __import__('edalize.vunit_hooks', fromlist=['vunit_hooks'])
E             File "/home/runner/work/edalize/edalize/.tox/py/lib/python3.6/site-packages/edalize/vunit_hooks.py", line 3, in <module>
E               from vunit.ui import Library
E           ModuleNotFoundError: No module named 'vunit'

.tox/py/lib/python3.6/site-packages/edalize/edatool.py:334: RuntimeError
=========================== short test summary info ============================
FAILED tests/test_vunit.py::test_vunit_hooks - RuntimeError: '/home/runner/wo...
========================= 1 failed, 39 passed in 3.96s =========================
ERROR: InvocationError for command /home/runner/work/edalize/edalize/.tox/py/bin/pytest (exited with code 1)
```

IMO this should be fixed as probably check_call may not identify the issue.
This change was needed to suppress the output of `run_tool` and redirect it to log files.

edalize/vivado.py
olofk: I like this approach. As @GCHQDeveloper560 says, this has traditionally been done outside of fusesoc/edalize. The main reason for that was to leave full control to the user and avoid any attempts at autodetecting these kind of things. But there are definitely cases where sourcing the scripts file would be better and I haven't thought of doing it with tool_options. This is pretty nice since we can do it on the command-line.

olofk:
Many thanks to @acomodi and @GCHQDeveloper560 for digging into this. I haven't had time yet to look at the patches. I have however started looking into the symbiflow side and I think it would be better if Edalize calls yosys directly instead of going through the synth wrapper. There are several reasons for this.
1. The synth wrapper only accepts verilog and xdc files, so we can't add tcl, vhdl, systemverilog etc.
2. By calling yosys directly we avoid a dependency on bash, which can be beneficial on non-linux systems (e.g. windows, mac and yosys compiled to webassembly. The last one will likely still have problems calling python for splitting the ports, but still)
3. We can hopefully reuse more code this way and just load an extra tcl file and set some env vars

I haven't looked at the later parts of the toolchain in detail, but I notice that several of the other wrappers call vpr with different options hardcoded. Generally I prefer to avoid hard coded options if it makes sense for some users to change them. Also in this case, I wonder if edalize should call vpr directly. Unfortunately I'm not skilled enough with vpr to make a good judgement

acomodi:
@olofk Thanks for the comments.

Now I see the reason behind https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1461 and it actually makes sense. Having only one command to call yosys, we can use it as a subtool in Edalize. Being that connected to this PR I could take over it and make it green so we can proceed in integrating symbiflow within Edalize.
Correct me if I am wrong, but I think this will require some changes also in the yosys edatool, in particular, the possibility of inserting exports of env variables in the makefile template, so that we can pass the various options in the synth script. 

Regarding the VPR options yes, right now they are hardcoded as they are needed to have a successful build. One thing left to do though (but this is on the symbiflow side), is to have a dynamic override of the hardcoded options, so that the user can arbitrarily choose his own options.

edalize/edatool.py
GCHQDeveloper560: It looks like the `vunit_hdl` package isn't getting installed, though I'm not sure why your change is uncovering the error. From tox-dev/tox#13 it appears that tox doesn't pay any attention to the `test_requires` line in `setup.py`. I'm not sure what the correct solution is, but perhaps adding `vunit_hdl` to the `deps` line in `tox.ini` would be the quickest solution.

acomodi: Thanks that worked, but now I see further error:
```
>           raise RuntimeError(_s.format(cmd, e.output.decode()))
E           RuntimeError: '/home/runner/work/edalize/edalize/.tox/py/bin/python' exited with an error code
E             ERROR - Could not detect known LLVM backend by parsing 'ghdl --version'
E           Traceback (most recent call last):
E             File "/home/runner/work/edalize/edalize/.tox/py/lib/python3.5/site-packages/vunit/ui/__init__.py", line 679, in main
E               all_ok = self._main(post_run)
E             File "/home/runner/work/edalize/edalize/.tox/py/lib/python3.5/site-packages/vunit/ui/__init__.py", line 723, in _main
E               return self._main_compile_only()
E             File "/home/runner/work/edalize/edalize/.tox/py/lib/python3.5/site-packages/vunit/ui/__init__.py", line 861, in _main_compile_only
E               simulator_if = self._create_simulator_if()
E             File "/home/runner/work/edalize/edalize/.tox/py/lib/python3.5/site-packages/vunit/ui/__init__.py", line 745, in _create_simulator_if
E               args=self._args, output_path=self._simulator_output_path
E             File "/home/runner/work/edalize/edalize/.tox/py/lib/python3.5/site-packages/vunit/sim_if/ghdl.py", line 74, in from_args
E               backend=cls.determine_backend(prefix),
E             File "/home/runner/work/edalize/edalize/.tox/py/lib/python3.5/site-packages/vunit/sim_if/ghdl.py", line 138, in determine_backend
E               "No known GHDL back-end could be detected from running 'ghdl --version'"
E           AssertionError: No known GHDL back-end could be detected from running 'ghdl --version'
E           Expected to find one of dict_keys(['GCC back-end code generator', 'mcode code generator', 'llvm code generator'])
E           == Output of 'ghdl --version'============================================================
E           
E           =========================================================================================
```

It appears that calling the ghdl tool causes troubles as ghdl is not installed in the environment, hence the assertion.

GCHQDeveloper560: Even with `vunit_hdl` installed `test_vunit` fails on your branch. The first problem seems to be that the GHDL mock command needs to support a `--version` argument. I added one with the following:
```
--- a/tests/mock_commands/ghdl
+++ b/tests/mock_commands/ghdl
@@ -2,6 +2,21 @@
 import os
 import sys

+version = """
+GHDL 0.36-dev (Ubuntu 0.35+git20181129+dfsg-4ubuntu1) [Dunoon edition]
+ Compiled with GNAT Version: 8.3.0
+  GCC back-end code generator
+  Written by Tristan Gingold.
+
+  Copyright (C) 2003 - 2015 Tristan Gingold.
+  GHDL is free software, covered by the GNU General Public License.  There is NO
+  warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+"""
+
+if sys.argv[1] == '--version':
+  print(version)
+  exit(0)
+
 cmd_file = 'analyze.cmd' if sys.argv[1] == '-i' else 'elab-run.cmd'

 with open(cmd_file, 'a') as f:
```

There's still an assertion failing with this change, but I've not looked any deeper.

acomodi: The assertion I am seeing is the following:

```
    def assert_called_once_with(_mock_self, *args, **kwargs):
        """assert that the mock was called exactly once and that that call was
        with the specified arguments."""
        self = _mock_self
        if not self.call_count == 1:
            msg = ("Expected '%s' to be called once. Called %s times." %
                   (self._mock_name or 'mock', self.call_count))
>           raise AssertionError(msg)
E           AssertionError: Expected 'VUnitRunner' to be called once. Called 0 times.

```

I see there are two mock patches called, once with `backend.build()` and the second with `backend.run()`.

Apparently, the backend.build() does not call the VUnitRunner, hence the assertions.

By removing the first mock (with the `backend.build()`, I get an additional error, which is a failure in the vunit tests:

```
E           Compiling into vunit_lib: ../../home/alessandro/.local/lib/python3.6/site-packages/vunit/vhdl/check/src/check.vhd                                  passed
E           Compiling into libx:      ../../data/edalize/tests/test_vunit/minimal/tb_minimal.vhd                                                               passed
E           Compile passed
E           
E           Starting libx.tb_minimal.testcase_1
E           Output file: vunit_out/test_output/libx.tb_minimal.testcase_1_2b9ae7a3e5456ab25a0a139dc55517d9c5e472b6/output.txt
E           fail (P=0 S=0 F=1 T=2) libx.tb_minimal.testcase_1 (0.0 seconds)
E           
E           Starting libx.tb_minimal.testcase_2
E           Output file: vunit_out/test_output/libx.tb_minimal.testcase_2_58710eec74ad7967b1fef6c6efaa30a99be64718/output.txt
E           fail (P=0 S=0 F=2 T=2) libx.tb_minimal.testcase_2 (0.0 seconds)
E           
E           ==== Summary ======================================
E           fail libx.tb_minimal.testcase_1 (0.0 seconds)
E           fail libx.tb_minimal.testcase_2 (0.0 seconds)
E           ===================================================
E           pass 0 of 2
E           fail 2 of 2
E           ===================================================
E           Total time was 0.1 seconds
E           Elapsed time was 0.1 seconds
E           ===================================================
E           Some failed!
```

olofk:
After looking at this some more, I realize there are too many changes for me to keep track of. So I would need this to be split up into several PR to be able to review. Sorry for the inconvenience. I have identified the following changes, which I would like as separate PRs

1. Add yosys support to the vivado backend
2. Reformatting with black (this was just done for the vivado backend. Perhaps we should do it for all backends at the same time. Preferably at a time when there aren't too many outstanding PRs)
3. Use nextpnr standalone. Some of the changes bring out nextpnr as a separate backend. Is this something you need to run standalone, or would it be better to just implement as an ordinary class that can be instantiated in the backends that needs it. (Somewhat related to this, I have recently started describing some ideas for the next generation of edalize https://github.com/olofk/edalize/wiki/Edalize-(Slight-return) . In that document, pretty much all tools will be run standalone, so we might be going in this direction anyway)
4. Getting output from the EDA tools. This is already being discussed [here](https://github.com/olofk/edalize/issues/21) Now that we have decided to drop python2 I think it should be pretty straightforward to implement
5. Add symbiflow support

In addition to that I think some changes could be dropped, like the check_args stuff and also avoid setting xilinx as the default arch in some places.

imphil:
> Reformatting with black (this was just done for the vivado backend. Perhaps we should do it for all backends at the same time. Preferably at a time when there aren't too many outstanding PRs)

This needs totally be a separate PR and done together with CI changes enforcing that, as we've done for fusesoc (ideally we copy the config over from there). I can do that.

acomodi:
@olofk Sure, I'll start splitting this PR as suggested, with the same order, as for instance `Add symbiflow support` depends on the previous items.

Just a couple of questions:

> Use nextpnr standalone.

I have currently set nextpnr for symbiflow as a subtool which, eventually can be run as a standalone tool I believe. One thing I don't understand is whether this patch needs some more work, and, in case the answer is yes, what would be the changes required?

> Getting output from the EDA tools.
I believe that the `check_output` solution could be the right path here. The stderr could be redirected to some temp file, or just being output during an exception (if any), while the output get captured in the return value of check_output as well.

olofk:
Just to make sure, is this superseded by #156?

acomodi:
@olofk Yes. I have further modified the Symbiflow/edalize fork and I would need to update https://github.com/olofk/edalize/pull/156 once again, but I believe this PR can be closed.

olofk:
I lost track a bit. Is there anything in this PR now that isn't covered by the other ones?

acomodi:
This was superseded by https://github.com/olofk/edalize/pull/191. Closing.

PR #147, opened 2020-04-30
Add preliminary Xcelium support

tudortimi:
This is basically a copy/paste of the ModelSim backend. It doesn't support VPI
modules yet.


tudortimi:
`test_xcelium` runs fine on its own, but fails when run together with all other tests.

I printed `PATH` and it contains the following:

```
/home/tudor/projects/edalize/tests/mock_commands/xcelium:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/projects/edalize/tests/mock_commands:/home/tudor/virtualenvs/fusesoc/bin:/opt/python/Python-3.5.9/bin:...
```

Each test adds an entry to `PATH` for `mock_commands`. It seems that the longer path gets truncated or something, because it can't find the mock `xmroot` executable.

tudortimi:
It's not that `PATH` is too long. I tried removing duplicates, to get the same `PATH` as when running the test by itself, but it still fails when running all of them.

cmarqu:
Is #130 solving your problem?

tudortimi:
I think it's because only writes to elements of `os.environ` trigger an `os.putenv(...)` in the background. I can imagine `os.environ` was never meant to be overwritten.

tudortimi:
I think the idiomatic way would be to use the `monkeypatch` fixture in all the tests.

olofk:
Recently, pytest is starting to make me cry almost as often as jinja2. I'm not the least surprised to hear that the code I have written around mocking and environment variables is scetchy. But if I understand correctly, you found a workaround for the failing test case. Perhaps we just shove the problem further under the carpet, but I think it's good enough for now until we find a unified way to deal with env vars that the EDA tools need to have. I have been thinking that perhaps that required env vars should be a property of the tool classes so that each backend doesn't have to deal with these things.

Anyway, please confirm if you think it's ready to go in and I'll push it

tudortimi:
The workaround is the extra `putenv` call, yes. Sweeping it under the rug sounds fine for now.

olofk:
Alright then. Xcelium, here we come! Thanks for your contribution. Picked and pushed

tudortimi:
Seems like something got busted on master, even though the checks ran on the branch.

PR #149, opened 2020-05-04
Fix xsim test failure for Python 3.5

GCHQDeveloper560:
Support for VHDL generics added in #137 introduced a non-ordered
dictionary. This changed the order of arguments, making the test
fail for Python 3.5


olofk:
Ah, right. So that's why it failed. Thanks. Picked and pushed

PR #151, opened 2020-05-05
Yosys support for vivado

acomodi:
This is the first PR split from [the symbiflow PR](https://github.com/olofk/edalize/pull/145) as suggested in [this comment](https://github.com/olofk/edalize/pull/145#issuecomment-623441246).

It adds support for yosys within Vivado. In additions there are also some changes to the vivado templates as well.


acomodi:
@olofk FYI

edalize/yosys.py
```diff
@@ -19,12 +19,37 @@ def get_doc(cls, api_ver):
...
                     'lists' : [
                         {'name' : 'yosys_synth_options',
                          'type' : 'String',
-                         'desc' : 'Additional options for the synth command'}
-                        , ]}
+                         'desc' : 'Additional options for the synth command'},
+                        ]}
+
+    @classmethod
+    def validate_args(cls, args):
```
olofk, line 38: Do we need this function? I suspect it was added to handle argument parsing, but that should not be done by Edalize anymore

edalize/yosys.py
```diff
@@ -51,27 +76,36 @@ def configure_main(self):
         for key, value in self.vlogparam.items():
             if type(value) is str:
                 value = "{\"" + value + "\"}"
-            _s = r"chparam -set {} {} {}"
+            _s = r"chparam -set {} {} \$abstract\{}"
```
olofk, line 79: Rebase error. Edalize was recently updated to drop the old chparam syntax that used $abstract

tests/test_icestorm/nextpnr/test_icestorm_0.tcl
```diff
@@ -28,11 +28,12 @@ foreach f ${file_table} {
   }
 }
 
-chparam -set vlogparam_bool 1 top_module
-chparam -set vlogparam_int 42 top_module
-chparam -set vlogparam_str {"hello"} top_module
+chparam -set vlogparam_bool 1 \$abstract\top_module
```
olofk, line 31: Rebase error

edalize/yosys.py
acomodi: Actually no, removed.

edalize/yosys.py
acomodi: Fixed

tests/test_icestorm/nextpnr/test_icestorm_0.tcl
acomodi: Fixed

olofk:
I think Vivado is run twice when yosys is used for synthesis. It seems also like it is trying to run it twice when vivado is used for synthesis but fails because `synth_1 needs to be reset before relaunching`. Haven't analyzed the makefile rules properly yet, but I guess the issue is likely there

One thing that confuses me a bit, is `edalize/templates/icestorm/icestorm-yosys-script.j2` used anywhere? I couldn't find a reference to it anywhere in the code. Not necessarily relevant to this PR but I got curious

olofk:
I got it working with this vivado-makefile.j2. Tested with and without `--synth=yosys` and with and without `--pnr=none`
 Can probably drop the phony changes. I don't know. Makefiles are confusing

```
SHELL=/bin/bash
NAME := {{ name }}
BITSTREAM := {{ bitstream }}
PART := {{ part }}

.PHONY : all synth pgm
all: $(BITSTREAM)
{% if yosys -%}
$(NAME).edif: $(NAME).mk
	$(MAKE) -f $<
$(BITSTREAM): $(NAME)_run.tcl $(NAME).xpr $(NAME).edif
	{{ vivado_command }} -notrace -mode batch -source $< $(NAME).xpr
{% else %}
$(BITSTREAM): $(NAME)_run.tcl $(NAME).xpr
	{{ vivado_command }} -notrace -mode batch -source $< $(NAME).xpr

{%- endif %}

$(NAME).xpr: $(NAME).tcl
	{{ vivado_command }} -notrace -mode batch -source $<

build-gui: $(NAME).xpr
	{{ vivado_command }} $<

$(NAME).runs/synth_1/__synthesis_is_complete__: $(NAME)_synth.tcl $(NAME).xpr
	{{ vivado_command }} -notrace -mode batch -source $^

{% if yosys -%}
synth: $(NAME).edif
{% else %}
synth: $(NAME).runs/synth_1/__synthesis_is_complete__
{%- endif %}


pgm: $(NAME)_pgm.tcl $(BITSTREAM)
	export HW_TARGET=$(HW_TARGET); \
	export JTAG_FREQ=$(JTAG_FREQ); \
	{{ vivado_command }} -quiet -nolog -notrace -mode batch -source $< -tclargs $(PART) $(BITSTREAM)

```

acomodi:
@olofk Ok, I have added some other changes to the Makefile. 

I think the double build issues where due to the `synth` target being called two times.

I am confused about the `__synthesis_is_complete__`. How does that work?

olofk:
@acomodi I'm not sure. Just found a file called `__synthesis_is_complete__` and thought that sounded like just the thing we needed :)

It didn't work for me however. Had to add the following change
```
diff --git a/edalize/templates/vivado/vivado-makefile.j2 b/edalize/templates/vivado/vivado-makefile.j2
index 9334aaf..06a37e8 100644
--- a/edalize/templates/vivado/vivado-makefile.j2
+++ b/edalize/templates/vivado/vivado-makefile.j2
@@ -20,7 +20,7 @@ synth: $(NAME).runs/synth_1/__synthesis_is_complete__
 {% if yosys %}
 $(BITSTREAM): $(NAME)_run.tcl $(NAME).xpr synth
 {% else %}
-$(BITSTREAM): $(NAME)_run.tcl synth
+$(BITSTREAM): $(NAME)_run.tcl $(NAME).xpr
 {% endif %}
        {{ vivado_command }} -notrace -mode batch -source $< $(NAME).xpr
``` 
I think the confusion comes from that the bitstream doesn't depend on the synthesis. The `$name_run` script does both synthesis and P&R while the `$name_synth` script only does synthesis. They can't be combined

olofk:
@acomodi I can just amend it locally and pull it in, if you like

acomodi:
@olofk Sure, done ;)

olofk:
Fuck yeah! Picked and pushed. Many thanks

PR #152, opened 2020-05-05
Add nextpnr xilinx support

acomodi:
This PR builds on top of https://github.com/olofk/edalize/pull/151.

It adds the nextpnr tool.


acomodi:
@olofk I have rebased this on top of master after the yosys to vivado support and fixed the tests

doc/edam/api.rst
```diff
@@ -31,6 +31,8 @@ Different EDA tools handle different subsets of files and are expected to ignore
 - vhdlSource-2008 : VHDL 2008 source
 - xci : Xilinx Vivado IP file
 - xdc : Xilinx Vivado constraint file
+- bba : Nextpnr-Xilinx chip data file
+- synthJson : Yosys-generated synthesys output file
```
olofk, line 35: Typo s/synthesys/synthesis
Is bba specific to Nextpnr-Xilinx, or are they used for other chip types supported by Nextpnr?

edalize/nextpnr.py
```diff
@@ -0,0 +1,140 @@
...
+                if l["name"] not in [i["name"] for i in combined_lists]
+            )
+
+            return {
+                "description": "Open source Place and Route tool targeting many different FPGAs",
+                "members": combined_members,
+                "lists": combined_lists,
+            }
+
+    @classmethod
```
olofk, line 67: I believe we can drop this function

mithro:
> I have however one big issue with this. Calling the backend nextpnr isn't very good since we already have the icestorm and trellis backends using nextpnr. I'm not sure what the best name would be, but traditionally, the backends have been named after the tools that generate the bitstream (e.g. trellis and icestorm). Looking at it from that perspective, this backend would be called fasm, but that's not very good either. I don't have a good answer right now for what to call it. Is this considered part of symbiflow? Maybe we should call the backend symbiflow and then just start adding more options as we get more symbiflow flows using e.g. vpr instead of nextpnr.

It's actually reasonable complicated, see https://docs.google.com/document/d/1Q8vaqD67aqPO7JbSDZtVvbg4BD7XBVuA_JJTwcYHprk/edit

Can we move to a flow named something like 'syn tool + pnr tool + bitstream tool' ?

Then we could have - `yosys+nextpnr+trellis` and  `yosys+nextpnr+icestorm` and  `yosys+ise+ise` and.....

acomodi:
As @mithro pointed out, the issue is that, to reach the bitstream generation step there can be multiple paths that can be taken, and, to adopt the suggested flow (`synth_tool + pnr_tool + bitstream_tool`) would require IMO a restructure of the current implementation of edalize.

In such a case, there could be different kinds of backends, namely synth, pnr and bitstream, each one providing the tool configuration that the user wants for a particular step of the flow.

On the other hand, it could probably be less of an effort to incorporate nextpnr-xilinx into a symbiflow backend, that, combined with https://github.com/olofk/edalize/pull/156 would likely solve the case.

olofk:
Several backends already support multiple synth and/or p&r tools but the assumption has been that there's a single bitstream generation tool. This is starting to fall apart a bit. Long-term I have started mapping out a new architecture for Edalize [here](https://github.com/olofk/edalize/wiki/Edalize-(Slight-return)) which I think is somewhat close to what @acomodi proposes.

But let's look at how to solve this short-term. I also think that a symbiflow backend could be the way to go. To begin with it would use yosys and have an option to choose between vpr+fasm or nextpnr.

Could we start with pulling out nextpnr to a separate file (for easier reuse) and then create a symbiflow backend that pulls in yosys+nextpnr+fasm?

acomodi:
Closing, superseded by https://github.com/olofk/edalize/pull/193

Issue #55, opened 2019-10-10
Support resource information reporting from various backends


mithro:
Here is a simple example for Vivado.

https://github.com/SymbiFlow/fpga-tool-perf/blob/f8438d0447ae808fd1434073b037fa30b274c423/fpgaperf.py#L444-L506
```python
    def vivado_resources(self):
        report_path = self.out_dir + "/" + self.project_name + ".runs/impl_1/top_utilization_placed.rpt"
        with open(report_path, 'r') as fp:
            report_data = fp.read()
            report_data = report_data.split('\n\n')
            report = dict()
            section = None
            for d in report_data:
                match = re.search(r'\n-+$', d)
                if match is not None:
                    match = re.search(r'\n?[0-9\.]+ (.*)', d)
                    if match is not None:
                        section = match.groups()[0]
                if d.startswith('+--'):
                    if section is not None:
                        # cleanup the table
                        d = re.sub(r'\+-.*-\+\n','', d)
                        d = re.sub(r'\+-.*-\+$','', d)
                        d = re.sub(r'^\|\s+','', d, flags=re.M)
                        d = re.sub(r'\s\|\n','\n', d)


                        report[section.lower()] = asciitable.read(d, delimiter='|', guess=False, comment=r'(\+.*)|(\*.*)', numpy=False)


        return report


    def resources(self):
        lut = 0
        dff = 0
        carry = 0
        iob = 0
        pll = 0
        bram = 0


        report = self.vivado_resources()


        for prim in report['primitives']:
            if prim[2] == 'Flop & Latch':
                dff += int(prim[1])
            if prim[2] == 'CarryLogic':
                carry += int(prim[1])
            if prim[2] == 'IO':
                iob += int(prim[1])
            if prim[2] == 'LUT':
                lut += int(prim[1])


        for prim in report['clocking']:
            if prim[0] == 'MMCME2_ADV' or prim[0] == 'PLLE2_ADV':
                pll += prim[1]


        for prim in report['memory']:
            if prim[0] == 'Block RAM Tile':
                bram += prim[1]


        ret = {
            "LUT" : str(lut),
            "DFF" : str(dff),
            "BRAM" : str(bram),
            "CARRY" : str(carry),
            "GLB" : "unsupported",
            "PLL" : str(pll),
            "IOB" : str(iob),
            }
        return ret
```

GCHQDeveloper560:
Would you expect each backend class to have reporting methods, or would they be colocated in an `edalize.reporting` module?

[...]

PR #154, opened 2020-05-07
Add resource and timing reporting for several backends

GCHQDeveloper560:
#55


GCHQDeveloper560:
I just tried running FuseSoC for the first time in a venv with this installed, and it looks like the naming convention I used broke Edalize's `get_edatool` in `__init__.py`. It's expecting classes to match `<module name>.capitalize()`. I used module names with underscores (`vivado_reporting.py`) and CapWords class names (`VivadoReporting`).

I'm not sure of the correct approach. These are likely `NON_TOOL_PACKAGES` so I could add them to that list. I could also change the class names to the expected format (e.g. `Vivado_reporting`) or change `get_edatool` to something more complex like
```python
def get_edatool(name):
    return getattr(
        import_module("{}.{}".format(__name__, name)),
        "".join(map(str.capitalize, name.split("_"))),
    )
```

Thoughts?

GCHQDeveloper560:
I've gone ahead and added these as `NON_TOOL_PACKAGES` but can take a different approach if desired.

olofk:
Thanks for this. It looks great. I hope to find some time soon to test it on a few of my own designs to see how it works. My main issue right now is that it adds quite a lot of new hard dependencies on other python packages. Could we instead just fail gracefully if e.g. the pandas packages aren't installed. I envision some corporate settings where it could be of importance to keep the number of edalize dependencies down.

GCHQDeveloper560:
The quickest solution might be to add reporting as an "extra" feature through [`extras_require`](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies)? I don't have any experience with it, but it sounds like it would allow applications not using reporting to avoid the dependencies.

If that doesn't sounds appropriate I suspect PyParsing (already used in FuseSoC) might be required, but I could look into whether something could be done without NumPy/Pandas.

I'm certainly interested in feedback on how it works with your designs. I would be shocked if testing with more devices and tool versions didn't turn up bugs.

olofk:
If you have identified numpy/pandas as a good solution, I definitely think we should stay with that. There is no written down policy for dependencies but I'm generally more fine with adding dependencies to FuseSoC than to Edalize, since Edalize is used within other projects to a higher degree. Ideally I would like to get rid of jinja2 as well, since that has caused about as much issues as it has saved time.

If extras_require is the way to go, please put it there (I have given up trying to understand the python packaging intricacies at this point and trust that other do the right thing). But I also think we could just do a try/except around the module imports and (optionally?) print out a note of warning if it couldn't be loaded. Then add some documentation and possibly some CLI command to let users know about the reporting infrastructure and how to install it if missing.

Come to think of it, I think the long-term solution would be something like an edalize namespace so that different backends can be developed more stand-alone and users can pick and choose the ones they want. But let's focus on getting this stuff in first. If you figure out how to make it optional, I will try to find time to run it on a couple of my designs to see how it holds up

GCHQDeveloper560:
I won't claim to have any expertise in Python packaging either, but the latest commit should avoid installing the reporting dependencies unless Edalize is installed with the reporting option (`edalize[reporting]`). The reporting modules are still installed. There is currently no fallback functionality, so the modules check for their dependencies, log a (hopefully) useful error, and throw the import exception. It might make more sense to just not install the modules, but it's not clear to me how to do that.

olofk:
I _finally_ got around to test it and it's crazy good :) Picked and pushed. Many thanks for taking the time to implement this

PR #155, opened 2020-05-09
templates: yosys-makefile.j2: add missing edif file in clean rule

trabucayre:
Since 2fce13111de2e7d071d2ddd08598fae5d4fe6a68#diff-3906f3cf8d9b1bb7335f0c02908fd07a a new target is added to generate .edif file. But the clean rule has not beeing updated accordinly.

### DIFF
--- a/edalize/templates/yosys/yosys-makefile.j2
+++ b/edalize/templates/yosys/yosys-makefile.j2
@@ -13,4 +13,4 @@ all: $(TARGET).{{ default_target }}
 	yosys -l yosys.log -p "tcl $?"
 
 clean:
-	rm -f $(TARGET).blif $(TARGET).json
+	rm -f $(TARGET).blif $(TARGET).json $(TARGET).edif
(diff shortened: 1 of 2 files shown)


olofk:
This looks correct but test suite is failing because the reference files have note been updated accordingly. You can run `GOLDEN_RUN=1 pytest` from the edalize root. This will run the test suite and (in most cases)  automatically update the reference files. Add these changes and I'm happy to approve

trabucayre:
I do that. 
But in a first time in need to understand why tests fails with vivado and xcelium. And why some tests files are revert to state before last commit...
True with my, and your, repository (up to date).

trabucayre:
Okay. I'm stupid... Wrong computer...
But fails again for missing space between params for spyglass (at least). By adding some spaces in templates it's work quite better. For xcelium it's fails but I haven't this tool.
I suppose I need to create a new PR to fix missing space?

olofk:
Hmmm...regarding the spaces, I wonder if it's that awful jinja templating thing acting up again. I got some indication that different versions have slightly different behavior. @imphil, any clues?

In the meantime, could you just add the updated reference files that are relevant for your change. If it passes the CI check I think we're good and we can figure out if there is something in particular about your environment that makes it fail

trabucayre:
Done.
Finally, after configuring my computer to use python3 by default + cleanup in ~/.local to drop jinja everything work fine in all my computers.

olofk:
Great! Thanks

Issue #158, opened 2020-05-16
Add `include_path` property to include files


tudortimi:
Needed to support `include` directives that contain paths, not just simple file names (as discussed at https://github.com/olofk/fusesoc/issues/397):

```
`include "some_dir/some_file.svh"
```

`some_file.svh` would have the properties `is_include_file` and `include_path: some_dir`.

`include_path` has to be a relative path starting from the core root (where the `.core` file is located), which points to a directory. We could also check that that directory is the parent or a grandparent directory of the file on which the property is attached.

olofk:
Agree. Specifying the include path as relative to core_root seems to make modt sense.

I think we can skip checking if it's a parent dir. That way we can probably also support things like `include ../dir/file.vh` in case someone does that.

Question: do we allow setting include_path without `is_include_file : true`. I think we can allow it but ignore include_path in that case, and let the frontend (e.g. FuseSoC) emit a warning

tudortimi:
Why not an error if `include_path` is set on a file that is not marked with `is_include_file: true`?

olofk:
I don't know really. Can probably come up with some convoluted example but don't have a real use case. Just think we don't have to make it illegal in case we realize later on that a valid use case exists

tudortimi:
Isn't it easier (in terms of keeping backward compatibility) to relax existing restrictions than it is to add new ones? Plus: https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

olofk:
I believe we can close this now

PR #159, opened 2020-05-18
Add 'include_path' property for include files

tudortimi:


tudortimi:
This is currently a dummy PR to show my intention of working on #158

tudortimi:
I'm thinking it would be better to throw the exception for illegal include path in the `Edatool` constructor.

tudortimi:
Ideally, I think it would be better to use a `File` class instead of a dictionary, which can do the required validations in its constructor. This would be a major API change, though. We could introduce it internally in the `Edatool` constructor:

```
class Edatool(object):
    def __init__(self, edam=None, work_root=None, eda_api=None):
        # ...
        for file in files:
            self.files.append(File(edam.get('files', []))
```

tudortimi:
If we merge #160, the validation of `include_path` can be moved there.

edalize/edatool.py
```diff
@@ -268,6 +271,10 @@ def __init__(self, name, file_type, logical_name):
         for f in self.files:
             if 'is_include_file' in f and f['is_include_file']:
                 _incdir = os.path.dirname(f['name']) or '.'
+                if 'include_path' in f and f['include_path']:
+                    if os.path.isabs(f['include_path']):
+                        raise EdamException("'include_path' of file {} cannot be absolute".format(f))
```
olofk, line 276: I think we should allow absolute paths for now. There might be cases where users need to include some system library. It's not portable, but at least we wouldn't prevent the use from doing it. Maybe we could print a warning instead

olofk:
Thanks. Looks great, but I would prefer to drop the check for absolute paths

tudortimi:
The check exists mostly to show that I only thought about what to do with relative paths. What should be the behavior for absolute paths? The same `os.path.relpath` call?

tudortimi:
I removed the check for absolute paths.

olofk:
Thanks! I think this should do it

olofk:
@tudortimi FYI, I realized now that there was some misunderstanding, so I need to change this a bit. We can't use relpath, because that doesn't work with paths that start with `..`, which is the case for all paths generated by FuseSoC (e.g. `../src/core_name/subdir/file.v`). The following spec will be implemented

In the EDAM file, `include_path` will be a path relative to the working directory (this is consistent with how file paths are currently specified)

In the FuseSoC CAPI2 core files, `include_path` will be a path relative to the core root.

Issue #118, opened 2020-03-04
Tidy up DSE/Quartus interface


a-gibson:
Following on from this [PR comment](https://github.com/olofk/edalize/pull/113#issuecomment-594532058) I'm starting this discussion to find the best way forward:

Desire:
Remove all DSE specific options and replace with one core file parameter such as `dse_options`.

Options:
The data provided by `dse_options` is written directly to a DSE file which is generated on-the-fly.
E.g.
```
targets:
  build:
    default_tool: quartus
    filesets: [stuff]
    tools:
      quartus:
        device: abc
        family: xyz
        dse_options:
          - 'seeds="1,2,3"'
          - 'num_concurrent=4'
          - 'num_parallel_processors=0'
```

Would produce a `.dse` file with the contents:
```
# Auto generated by Edalize
seeds="1,2,3"
num_concurrent=4
num_parallel_processors=0
```

Another option might be native support a `dse` file type.

E.g.
```
filesets:
  synth:
    files:
      - /path/to/dse/file {file_type: dse}
```

The file would be supplied as-required by DSE. This option will probably require some resolution logic to determine what to do if zero or more than one `.dse` files are provided.

There are probably other options. Would be grateful for thoughts and opinions.

GCHQDeveloper560:
Would it make more sense for `dse_options` to be a dict rather than a list of strings? At the FuseSoC YAML level:
```
dse_options:
    - seeds: [1, 2, 3]
    - num_concurrent: 4
    - num_parallel_processors: 4
```

I don't think any Edalize backends use the `dict` argument type (just `member` and `list`) but it does exist.

I don't have an opinion on whether a DSE file type would be a cleaner solution. I would think not since as you mention other than one file doesn't make sense. It also doesn't apply for to other tools, though we do have tool-specific things like QIP, XCF, etc.

On a related note, Xilinx and Vivado now have a `pnr` option that's effectively a faint reflection of the vendor flow steps. DSE has overloaded this, and my first thought for #115 was to add `netlist` and `power` values. Of course, Quartus can run simulators and Vivado has a built in simulator, so if we wanted to support them the obvious next step would be to add a `pnr` value of `simulation` which seems a little off.

Is there a better way we should be describing how far to run the tool flow, hopefully in a tool-independent manner? I wonder if this might also help with something cleaner than top-level options for very specific tool flows like `dse_options`. Perhaps
```
flow: dse
    options:
        - seeds: [1, 2, 3]
        - num_concurrent: 4
[...]

PR #164, opened 2020-06-02
118 tidy dse interface

a-gibson:
This PR should address #118 .

Sorry it took so long, lockdown kinda got in the way of life! :-) Hope everyone is safe and well!


olofk:
Thanks for the patches and for the concern. This looks fine to me. Not a DSE user myself but trust you're doing the right thing here

PR #165, opened 2020-06-03
Vivado: fix Yosys options passing

kgugala:
This PR fixes a bug where synth options were not passed to Yosys in Yosys -> Vivado flow.


olofk:
All good. Thanks

PR #169, opened 2020-06-23
Allow verbose output from nextpnr

### LINKED ISSUE #168: Provide verbose output option
When trying to debug a nextpnr issue with fusesoc and edalize, I was unable to remove the -q from the nextpnr-ice40 invocation with --verbose on the fusesoc side. 

It would be nice if edalize could be passed a verbose option, and that apply to the backends where applicable. My work around for now is to just run the command manually. 

Seems that the -q option is forced here: https://github.com/olofk/edalize/blob/79050a966bcb377f4a33a699f072bc38f0d39922/edalize/templates/icestorm/icestorm-makefile.j2#L23

nevercast:
Allow verbose output from edalize for nextpnr-ice40 and nextpnr-ecp5 by removing the quiet flag (-q)

Users may add the flag upstream via NEXTPNR_OPTIONS
For example, this can be done in the Fusesoc frontend by using nextpnr_options on the respective toolchain:

https://fusesoc.readthedocs.io/en/master/ref/capi2.html#icestorm
https://fusesoc.readthedocs.io/en/master/ref/capi2.html#trellis

Closes #168

### DIFF
--- a/edalize/templates/icestorm/icestorm-makefile.j2
+++ b/edalize/templates/icestorm/icestorm-makefile.j2
@@ -20,7 +20,7 @@ stats: $(TARGET).stat
 %.json: %.mk
 	$(MAKE) -f $<
 %_next.asc: %.json
-	nextpnr-ice40 -l next.log -q $(NEXTPNR_OPTIONS) --pcf $(PCF_FILE) --json $? --asc $@
+	nextpnr-ice40 -l next.log $(NEXTPNR_OPTIONS) --pcf $(PCF_FILE) --json $? --asc $@
 %.bin: %_$(PNR).asc
 	icepack $< $@
 %.tim: %_$(PNR).asc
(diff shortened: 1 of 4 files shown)


nevercast:
I see that my changes to the test files have caused a "known good truth" comparison to fail. I see two solutions:

1. Update the known good reference files
2. Remove my change to the tests, keeping the -q option there 

Would prefer a maintainer advise on the course of action here.

imphil:
You can update the reference files with `GOLDEN_RUN=1 pytest`. See https://edalize.readthedocs.io/en/latest/source/tests.html for a bit of documentation on that.

olofk:
Sorry, this took quite a bit longer than intended, but I pulled in these fixes together with updates to the reference files. Thanks for your contribution

nevercast:
Thanks for finishing up this PR for me. I really appreciate it.

PR #170, opened 2020-06-29
Add support for C++ in ModelSim VPI modules

GCHQDeveloper560:
Use built-in rules for object file creation rather than a C-specific
pattern rule and set CXXFLAGS where needed

### DIFF
--- a/edalize/modelsim.py
+++ b/edalize/modelsim.py
@@ -11,7 +11,8 @@ $(error Environment variable MODEL_TECH was not found. It should be set to <mode
 endif
 
 CC ?= gcc
-CFLAGS := -c -std=c99 -fPIC -fno-stack-protector -g
+CFLAGS   := -fPIC -fno-stack-protector -g -std=c99
+CXXFLAGS := -fPIC -fno-stack-protector -g
 
 LD ?= ld
 LDFLAGS := -shared -E
@@ -19,8 +20,9 @@ LDFLAGS := -shared -E
 #Try to determine if ModelSim is 32- or 64-bit.
 #To manually override, set the environment MTI_VCO_MODE to 32 or 64
 ifeq ($(findstring 64, $(shell $(MODEL_TECH)/../vco)),)
-CFLAGS  += -m32
-LDFLAGS += -melf_i386
+CFLAGS   += -m32
+CXXFLAGS += -m32
+LDFLAGS  += -melf_i386
 endif
 
 RM ?= rm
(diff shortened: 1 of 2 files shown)


olofk:
Thanks. I have no idea how this enables c++, but I get a feeling your Makefile skills are far better than so I'm happy to trust you on this

PeterCDMcLean:
I tried out commit 15e2501 in my application.
It worked, but there is another problem. using ld as a linker does not automatically link in certain standard C++ libraries. For example, I was getting issues with undefined __dso_handle which is in crtendS.o 
Now, I did try using g++ as the linker front-end to avoid these problem, but parameter passing to the linker is impossible from .core file. Using g++ as linker front-end you have to prefix many options with -Wl,--linker-only-option. This presents a problem to the core file as follow:

This is my hack to pass linker options via the libs field:

libs: [c --whole-archive ../../../cmake_build/libdpi_axi.a --no-whole-archive ...]

However, if g++ was linker front-end I need to do this:
libs: [c -Wl,--whole-archive ../../../cmake_build/libdpi_axi.a -Wl,--no-whole-archive ...]

But the commas screw up the parameter passing.

GCHQDeveloper560:
I would have hoped including `stdc++` in libs would have covered the C++ case, but I'm not surprised linking C++ using ld directly is challenging.

I don't have access to a ModelSim manual at the moment, but how do they recommend compiling VPI modules? Do they address C++?

There's a little discussion of additional parameter passing in olofk/fusesoc#403 so perhaps something similar could be used to allow setting `LDFLAGS` for `-Wl` options.

Your example seems more complex than a basic linker option though. I'm surprised the hacks work since I thought Edalize prefixed all of the libraries with `-l` so I would have thought you'd have gotten `-l--whole-archive -l../../../cmake_build/libdpi_axi.a -l--no-whole-archive` in the Makefile.

Would being able to set LDFLAGS work for your case? I'd guess the best approach is figuring out a Makefile that works for you, and then we could figure out the best way to create that from FuseSoC/Edalize.

olofk:
Ok, so reading this I see there are still things needing to be fixed. VPI is a bit tricky since it adds all the intricate problems of C/C++ compiling which is somewhat still an unresolved problem despite 30 years of different build systems. From the FuseSoC/Edalize perspective I think it would be handy to support VPI building natively as long as it's a matter of specifying a bunch of source files, perhaps some libs and what not. But at some point it's better to just give up, let something external handle building the actual library and then just have FuseSoC pass the compiled lib to the simulator. Perhaps the most flexible way would be to add a vpiLibrary file type and then a generator can do whatever they want to actually build the vpi lib and just add the compiled file to the EDAM

PR #171, opened 2020-06-29
Remove dead check in jinja_filter_param_value_str

rswarbrick:
If value is a bool, the previous branch of the 'if' will
return (regardless of whether bool_is_str is true or not).


imphil:
Thanks!

PR #173, opened 2020-06-29
SymbiYosys support

rswarbrick:
**EDIT**: After rewriting the tool to use a Jinja template, I rewrote the commit message. This PR text is now the new commit message.

The SymbiYosys tool is a little tricky to support because the .sby
file that it uses for configuration contains a list of files, as well
as a list of commands. The user is going to want to edit the latter,
but the former needs to come from fusesoc or similar.

This patch assumes that the .sby file will be a Jinja2 template.
There's detailed documentation about the format in the Symbiyosys
class's _description (which should appear with --help).

A minimal example would be something like this:

    [options]
    mode bmc
    depth 3

    [script]
    {{"-sv"|gen_reads}}
    prep -top my_top_module

    [files]
    {{files}}

When run, the 'configure' step interpolates the .sby configuration
script, writing to <work_root>/test.sby. It also dumps a list of RTL
file paths (to <work_root>/files.txt) and include directories (to
<work_root>/incdirs.txt). This is handy if you need to run some sort
of preprocessing step (sv2v?) on your input files to massage them into
a form that Yosys will support.

The 'build' step does nothing.

The 'run' step runs 'sby'. The backend has a single parameter,
"tasknames". This defaults to empty, in which case, the command is
something like

    sby -d build test.sby

However, if you want multiple tasks in your config file, you'll need
to specify which ones to run. "tasknames" let's you pass them, ending
up with a command like

    sby -d build test.sby task1 task2

Note: This adds a file to the mock file list in
edalize_common.py (because the symbiyosys backend needs a .sby file).
The Vivado test needed regenerating with GOLDEN_RUN=1; the others
seemingly didn't need any changes.


rswarbrick:
It looks like this fails CI on macos-latest. Unfortunately, I don't seem to be able to see the logs. Could someone with more permissions take a look? I don't think I know how to debug it otherwise :-/

rswarbrick:
@imphil: Thanks for helping me debug this!

For anyone else, the problem was that I'd added an entry to FILES in `tests/edalize_common.py` and needed to regenerate the golden values in the Vivado test.

rswarbrick:
For an example of how this can be used with FuseSoc (and sv2v) for a full flow, see the 3rd patch of https://github.com/lowRISC/ibex/pull/994.

rswarbrick:
Force-push rebases onto recent test infrastructure changes. No other change.

rswarbrick:
(Close/Open dance because the CI seems to have picked up an old version of the patch for testing)

imphil:
Thanks for this patch @rswarbrick. I have been thinking about ways to make this backend more aligned with the other backends we have in edalize. Currently, the edalize backends generally have two properties:

- They provide a default "out of the box" configuration. This configuration can be modified by either tool options (in core files), or by including other config files in tool-specific ways.

- Templating is done with Jinja2, if so.

Applying that to the symbiyosys backend, can we:

- Provide a default .sby file which is fully generated by edalize from a Jinja2 template. With a task name given, plus the default file list and the toplevel from the core file, we should have all necessary information available for a minimal, but functional sby file.

- (Optional) Think about other configuration which might be sufficiently common to provide a tool option in edalize for.

- As extension vector, give the user the option to override the Jinja2 template used to generate the sby file with their own one. The user can do that by copying the template used by edalize somewhere, and adding a file with a (say) sbyConfigTemplate file type to the file list. I've been thinking about doing that for other backends for quite a while, and this one seems to be a good motivation for trying it out.

Two more things:

- According to the docs at https://symbiyosys.readthedocs.io/en/latest/reference.html#files-section, files with absolute paths are not copied by symbiyosys. Since fusesoc already (optionally) copies files, using absolute file paths should avoid an additional copy, and also make two files with the same file name possible. The "build" directory isn't considered portable between computers anyway, so having an absolute path in there doesn't hurt.

- Once we've settled on a design, we need at least a bit of documentation on how this sby file mechanism works.

rswarbrick:
Hmm. I thought quite hard about the .sby file before I came up with this design. The problem is that there isn't really a "standard mode". For a simple example, see https://github.com/lowRISC/ibex/blob/master/formal/icache/run.sby. This defines two different modes (prove and cover) and different options for the two modes.

Obviously, we *could* reflect all of these things in the Edalize input; in practice, this would be in some fusesoc core file. But what would be the point? Now we'd have a separate format that we'd have to document, and the user would have to read two documents, rather than just one (the sby tool's documentation).

I can see a strong argument for doing this sort of thing when you have multiple simulators, for example, and you want to be able to say "simulate this!" without worrying about internals. But it makes much less sense when there's just one tool you support.

Long term, we'll still need something like what's in this PR in the moment, as an "expert mode". This is what you call an extension vector in your comment. Personally, I'd prefer to merge just an expert mode for now, then try to figure out a less-expert-mode that could be shared between different formal tools later. In particular, I think it's wiser not to do that until we support at least one other formal tool in Edalize. Otherwise, how do we know we've made the right generalisations? To keep backwards compatibility, we'll have to support any decisions forever, so I'd rather not make them until we have to! :-)

Does this seem reasonable?

Good point about the absolute paths: I'll push up a new version later today that uses them.

olofk:
Thanks for doing this. I have been wishing for symbiyosys support for a long time! Running riscv-formal is my main use case to begin with as well, so happy to see you got that covered. And just like you, I identified the mix between automatic and manual entries in the sby file as the biggest issue. Can't believe I'm saying this, but a tcl API would have helped here :) I just noticed there is support for inlining python code. Is that something we could use here or do we just make things unnecessarily complicated? I also wonder if we could convince the SymbiYosys devs to support including files. That way we could make it easier to combine snippets.

But it might well be that the proposed solution is a good middle ground already. In that case we should just make sure it's easy to switch to another mode if we need more or less control in the future.

And about absolute paths, the build dirs are actually supposed to be portable to the largest extent possible. I routinely use this myself to archive projects or deliver them to customers without them having to install fusesoc

imphil:
Sorry for the late reply, I forgot about this one, but today I'm in review frenzy :-)

I thought when reading the SymbiYosys docs that we could come up with a minimal standard configuration, but I'm fine not tackling this now. So let's go for "expert mode only", as you wrote @rswarbrick.

That leaves us with one request from my side: make the sby file a Jinja2 template; I know it's overkill here, but I'd really like to stay with one templating approach within edalize.

edalize/symbiyosys.py
```diff
@@ -0,0 +1,297 @@
...
+If you need more control over the read commands at the start of the script
+section, there are some variables that can help you.
+
+  - ``{{chparam}}`` expands to the chparam command at the end of gen_reads as
+    shown above.
+
+  - ``{{flags}}`` expands to the flags (defines and includes) used for each
+    read line.
+
+  - ``{{src_files}}`` expands to a list with the basename of each source file.
```
imphil, line 84: Can you also make the name of the toplevel available?

edalize/symbiyosys.py
```diff
@@ -0,0 +1,297 @@
...
+        self.rtl_paths = []
+        bn_to_path = {}
+        sby_names = []
+
+        # RTL files have types verilogSource or systemVerilogSource*. We
+        # presumably want some of them. The .sby file has type sbyConfig: we
+        # want exactly one of them.
+        for file_obj in src_files:
+            ft = file_obj.file_type
+            if ft.startswith('verilogSource') or ft == 'systemVerilogSource':
```
imphil, line 154:
```suggestion
            if ft.startswith('verilogSource') or ft.startsWith('systemVerilogSource'):
```

(These file types are a superset of the IP-XACT file types. IP-XACT specifies systemVerilogSource, systemVerilogSource-3.0, systemVerilogSource-3.1, and systemVerilogSource-3.1a. [Yes, super old.])

tests/test_symbiyosys/config.sby.j2
```diff
@@ -0,0 +1,21 @@
...
+cv: mode cover
+cv: depth 32
+
+[engines]
+smtbmc boolector
+
+[script]
+{{"-sv"|gen_reads}}
+
+prep -top my_top_module
```
imphil, line 18: Can you use a toplevel template variable here?

imphil (approved):
Nice one, thanks! Two small comments, and then we're ready to go.

edalize/symbiyosys.py
rswarbrick: Done! It's probably not quite as useful as the other bits: you generally know the top-level module of your design but (with fusesoc core files) you don't necessarily know the list of files that goes into it.

edalize/symbiyosys.py
rswarbrick: Done! (With a regex, which seemed a bit neater than a "double startswith".)

tests/test_symbiyosys/config.sby.j2
rswarbrick: Done! And, amusingly, I'd got the name wrong - it should have just been `top_module`.

imphil: Yeah, I saw that :-)

imphil:
Thanks Rupert! I quite like how this backend turned out in the end, and I'm sure we'll use it as example for similar backends in the future.

PR #174, opened 2020-06-30
Use pytest fixtures to avoid boilerplate in tests

rswarbrick:
This patch was actually inspired by trying to stop the tests leaving
temporary directories in /tmp. But it turns out there was a lot of
repeated boilerplate in tests, where they call setup_backend with
pretty standard arguments.

There's now a TestFixture class defined in edalize_common.py. Objects
of type TestFixture are called "tf". We create these TestFixture
objects with a "make_edalize_test" pytest fixture. This is a factory
because it lets the individual tests customize things like
param_types, ref_dir etc.

The changes to tests/test_*.py are pretty mechanical; the interesting
code changes are probably in tests/edalize_common.py.

### DIFF
--- a/tests/edalize_common.py
+++ b/tests/edalize_common.py
@@ -1,21 +1,117 @@
-import os.path
-import tempfile
 from collections import OrderedDict
+import os.path
+import shutil
+
+import pytest
 
 from edalize import get_edatool
 
+
 tests_dir = os.path.dirname(__file__)
+
+
+class TestFixture:
+    '''A fixture that makes an edalize backend with work_root directory
+
+    Create this object using the make_edalize_test factory fixture. This passes
+    through its `tool_name` and sets up a temporary directory for `work_root`,
+    then passes its keyword arguments through to the TestFixture initializer.
+
+    Args:
+
+        tool_name: The name of the tool
+
+        work_root: The directory to treat as a work root
+
+        test_name: The name to call the backend. Defaults to
+                   `'test_<tool_name>_0'`
+
+        param_types: A list of parameter types. Defaults to `['plusarg',
+                    'vlogdefine', 'vlogparam']` (the parameter types supported
+                    by most simulators).
+
+        files: A list of files to use. Defaults to `None`, which means to use
+               :py:data:`FILES`.
+
+        tool_options: Dictionary passed to _setup_backend. Defaults to `{}`.
+
+        ref_dir: A reference directory relative to `test_<tool_name>`. Defaults
+                 to `'.'`
+
+        use_vpi: If true, set up backend with definitions from :attr:`VPI`.
+                 Defaults to `False`.
+
+    '''
+    def __init__(self,
+                 tool_name,
+                 work_root,
+                 test_name=None,
+                 param_types=['plusarg', 'vlogdefine', 'vlogparam'],
+                 files=None,
+                 tool_options={},
+                 ref_dir='.',
+                 use_vpi=False):
+
+        raw_ref_dir = os.path.join(tests_dir, 'test_' + tool_name, ref_dir)
+
+        self.test_name = ('test_{}_0'.format(tool_name)
+                          if test_name is None else test_name)
+        self.ref_dir = os.path.normpath(raw_ref_dir)
+        self.work_root = work_root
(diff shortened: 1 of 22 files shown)


rswarbrick:
Note: This conflicts with PR #173. Once either is merged, I'll update the other to match.

tests/edalize_common.py
```diff
@@ -1,21 +1,90 @@
...
+        work_root.mkdir()
+        ref_dir = os.path.normpath(os.path.join(tests_dir,
+                                                'test_' + tool_name,
+                                                ref_dir))
+        fixture = TestFixture(test_name, tool_name, param_types, files,
+                              tool_options, ref_dir, str(work_root), use_vpi)
+        created.append(fixture)
+        return fixture
+
+    yield _fun
```
imphil, line 78: What's the benefit of using a `yield` fixture here? There's no tear-down code after the yield, so you can make this a simple `return`.

tests/edalize_common.py
```diff
@@ -1,21 +1,90 @@
...
+class TestFixture:
+    def __init__(self, test_name, tool_name, param_types, files,
+                 tool_options, ref_dir, work_root, use_vpi):
+        self.test_name = test_name
+        self.ref_dir = ref_dir
+        self.work_root = work_root
+        self.backend = _setup_backend(test_name, tool_name, param_types, files,
+                                      tool_options, work_root, use_vpi)
+
+    def compare_files(self, files, ref_subdir='.'):
```
imphil, line 23: Please document that this function "forwards" to edalize_common.compare_files(); the docstring in there with the GOLDEN_RUN environment variable is rather useful information.

tests/edalize_common.py
```diff
@@ -1,21 +1,108 @@
...
 
+
+class TestFixture:
+    '''A fixture that makes an edalize backend with work_root directory
+
+    Create this object using the make_edalize_test factory fixture. This passes
+    its keyword arguments through to the TestFixture initializer, which
+    supports the following:
+
+    test_name: The name to call the backend. Default: test_<tool_name>_0
```
imphil, line 21: Please add an

```
Args:
  test_name: ...
```

to render the arguments in the docs. You can try it out by switching over to the `doc` directory, and typing `make html`.

tests/edalize_common.py
```diff
@@ -25,8 +112,8 @@ def compare_files(ref_dir, work_root, files):
         if 'GOLDEN_RUN' in os.environ:
             shutil.copy(generated_file, reference_file)
 
-        with open(reference_file) as fref, open(generated_file) as fgen:
-            assert fref.read() == fgen.read(), f
+        assert filecmp.cmp(reference_file, generated_file, shallow=False), \
+            "Mismatch between {} and {}".format(reference_file, generated_file)
```
imphil, line 116: Sorry, I didn't see that one before: The benefit of the `assert` is that pytest gives a diff if there's a mismatch. It would be great if we can keep that property.

tests/edalize_common.py
rswarbrick: Turns out that's not enough to get pretty parameter lists. I've updated the patch with the [correct Sphinx syntax](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html), which now looks pretty horrible in the source code. Is that better or worse?

tests/edalize_common.py
rswarbrick: I'm not sure that's true: I think the ", f" tells pytest to just print the file name. But I'll revert the hunk: it doesn't really make much difference.

rswarbrick:
Latest force-push removes the unused `filecmp` import.

tests/edalize_common.py
imphil: It should be enough (and I tested the diff I've posted initially), since edalize uses the napoleon sphinx extension (https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html). Maybe it needs slightly more indentation?

rswarbrick: Ahah! Yes, it looks like it was just an indentation problem. Much better :-)

imphil:
Nice one, thanks!

ThomasHornschuh:
Seems to conflict with my PR #131.  Have to find time to rewrite my test cases in this PR...

PR #179, opened 2020-07-08
Pick '.' for qsys "srcdir" when the file is in '.'

christiaanb:
Otherwise the generated Makefile will look for qsys file in `/`, because "srcdir" was empty.


olofk:
Thanks for this. The tests were failing so I took the liberty to amend the patch with the required changes to the reference files. Make sure to run `pytest` before committing. You can also run `GOLDEN_RUN=1 pytest` to have the reference files automatically updated. Anyway, picked and pushed

christiaanb:
Ah yes.. apologies for that. I meant to get back to actually fixing this after figuring out how to develop python packages, but it slipped my mind. Thanks for that tip on running `pytest`, it would've taken me some time to figure that out. Perhaps a `Hacking.md` in the root of the repo with these very simple instructions would be quite beneficial for the uninitiated such as myself.

olofk:
That's a good idea. It's hidden deep inside the docs so I have to mention it in about half the PRs I get. Better to give contributors an easier way to figure out what should be done

PR #188, opened 2020-08-30
Add target for morty (source file pickler)

zarubaf:
Morty is an SV source file pickler (https://github.com/zarubaf/morty) helping with some of the deficiencies of SV (like no namespacing). 
This commit adds support for calling `morty` from tools using `edalize`.

### DIFF
--- a/edalize/morty.py
+++ b/edalize/morty.py
@@ -1,0 +1,68 @@
+import os
+import logging
+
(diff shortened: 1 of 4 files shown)


olofk:
Fantastico! Thanks for this. Pickled and pushed!

Issue #189, opened 2020-09-08
logger not imported in yosys.py


DaveBerkeley:
$ ./build.py
Traceback (most recent call last):
  File "./build.py", line 33, in <module>
    backend.configure()
  File "/tmp/edalize/edalize/edatool.py", line 112, in configure
    self.configure_main()
  File "/tmp/edalize/edalize/yosys.py", line 71, in configure_main
    logger.error("ERROR: arch is not defined.")
NameError: name 'logger' is not defined

Needs  :

import logging
logger = logging.getLogger(__name__)

Error occurs if arch is not defined. It could throw an Exception instead as the error is fatal?

imphil:
Thanks for your bug report. Can you create a pull request to get your proposed change integrated? (The fix looks right to me.)

PR #190, opened 2020-09-08
added missing 'import logging'

DaveBerkeley:
Fix for logger not imported in yosys.py issue #189

### DIFF
--- a/edalize/yosys.py
+++ b/edalize/yosys.py
@@ -6,7 +6,6 @@ from edalize.edatool import Edatool
 
 logger = logging.getLogger(__name__)
 
-
 class Yosys(Edatool):
 
     argtypes = ['vlogdefine', 'vlogparam']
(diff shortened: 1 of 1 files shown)


edalize/yosys.py
```diff
@@ -1,7 +1,12 @@
+
```
imphil, line 1: Can you remove this empty line?

imphil (commented):
Thanks for this patch! One small change and we're ready to get this merged. Please rebase the existing commit and force-push, if possible.

edalize/yosys.py
DaveBerkeley: why is this necessary?

imphil: Thank you for your update to this pull request. You removed an empty line further down in the file, which is fine. However, the first line in this file is still an empty line, please remove it as well (it's unnecessary).

DaveBerkeley: what is the problem with it?

olofk:
Thanks for the patch. Squashed and merged!

PR #156, opened 2020-05-12
Add new SymbiFlow edatool

acomodi:
This PR adds support for SymbiFlow. 

Currently, this patch is part of the SymbiFlow/edalize fork and it is used in [FPGA tool perf](https://github.com/SymbiFlow/fpga-tool-perf)

PR #191, opened 2020-09-24
Add new SymbiFlow edatool

kamilrakoczy:
Supersede of #156

This PR adds support for SymbiFlow.

Currently, this patch is part of the SymbiFlow/edalize fork and it is used in FPGA tool perf


edalize/symbiflow.py
```diff
@@ -0,0 +1,160 @@
...
+                        "type": "String",
+                        "desc": "FPGA chip package (e.g. clg400-1)",
+                    },
+                    {
+                        "name": "part",
+                        "type": "String",
+                        "desc": "FPGA part type (e.g. xc7a50t)",
+                    },
+                    {
+                        "name": "builddir",
```
olofk, line 44: I would prefer to just use `.` if that's an option. All backends are assumed to be run from a clean build dir. If you need a subdir for some reason, just hardcode the name instead of exposing to the user

edalize/symbiflow.py
```diff
@@ -0,0 +1,160 @@
...
+                        "type": "String",
+                        "desc": 'Place and Route tool. Currently only "vpr" is supported',
+                    },
+                    {
+                        "name": "options",
+                        "type": "String",
+                        "desc": "Tool options. If not used, default options for the tool will be used",
+                    },
+                    {
+                        "name" : "seed",
```
olofk, line 64: Do we really need to expose seed as a separate option. Looking at the Makefile template below I see that all other options are sent to symbiflow_pack and symbiflow_place, symbiflow_route and symbiflow_write_fasm. Is it ok to pass --seed to all stages as well, or is that one special?

edalize/symbiflow.py
```diff
@@ -0,0 +1,160 @@
...
+                        "type": "String",
+                        "desc": 'Target architecture. Currently only "xilinx" is supported',
+                    },
+                    {
+                        "name": "pnr",
+                        "type": "String",
+                        "desc": 'Place and Route tool. Currently only "vpr" is supported',
+                    },
+                    {
+                        "name": "options",
```
olofk, line 59: Can we name this one vpr_options instead to be consistent with other backends?

edalize/symbiflow.py
```diff
@@ -0,0 +1,160 @@
...
+        builddir = self.tool_options.get("builddir", "build")
+
+        # copy user files to builddir
+        for f in user_files:
+            shutil.copy(f, self.work_root + "/" + builddir)
+
+        part = self.tool_options.get("part", None)
+        package = self.tool_options.get("package", None)
+
+        assert part is not None, 'Missing required "part" parameter'
```
olofk, line 117: Use logger.error instead of assert

edalize/templates/symbiflow/symbiflow-vpr-makefile.j2
```diff
@@ -0,0 +1,58 @@
...
+SDC_OPTS = -s ${SDC}
+{%- endif %}
+
+all: ${TOP}.bit
+
+${BUILDDIR}:
+	mkdir ${BUILDDIR}
+
+${TOP}.eblif: | ${BUILDDIR}
+	cd ${BUILDDIR} && symbiflow_synth -t ${TOP} -v ${VERILOG} -d ${BITSTREAM_DEVICE} -p ${PARTNAME} ${XDC_OPTS}
```
olofk, line 39: I'm not super excited about calling the wrapper script instead of calling yosys directly. The wrapper only accepts verilog and xdc files, so we can't add tcl, vhdl, systemverilog etc and we can't pass yosys_synth_options. I'm ok with using the wrapper to begin with if we intend to fix it later on.

edalize/yosys.py
```diff
@@ -32,6 +32,9 @@ def get_doc(cls, api_ver):
                         {'name' : 'yosys_synth_options',
                          'type' : 'String',
                          'desc' : 'Additional options for the synth command'},
+                        {'name' : 'yosys_additional_commands',
```
olofk, line 35: I prefer additional commands to come in as tcl files instead as is done in the other tcl-supporting backends

olofk (changes requested):
Thanks for working on this. We're getting closer for each time. There are still some changes to be made but I think we can start with a basic version and improve over time so that we can at least get the first one merged soon

edalize/symbiflow.py
kamilrakoczy: Fixed. I removed this option completely, as we used always ``.`` anyways.

edalize/symbiflow.py
kamilrakoczy: Done.

edalize/yosys.py
kamilrakoczy: Done.

edalize/symbiflow.py
kamilrakoczy: No, we do not need to expose ``seed`` as a separate option. I updated toolchain wrappers to allow passing additional vpr options (that are added to the default one from the wrapper). (https://github.com/SymbiFlow/symbiflow-arch-defs/pull/1682)

Yes, we can set ``seed`` option in all vpr stages, but ``seed`` is only used in ``place``. I changed seed to be part of ``additional_vpr_options`` and added this to all stages of vpr.

edalize/symbiflow.py
kamilrakoczy: Changed name to ``additional_vpr_options``.

edalize/templates/symbiflow/symbiflow-vpr-makefile.j2
kamilrakoczy: Yes, we intend to fix it later on.

edalize/symbiflow.py
kamilrakoczy: I see this is only unresolved comment.

I named this option as ``additional_vpr_options`` to emphasize that, this options, are added to the default one defined in toolchain_wrapper. 
Currently there is no option to overwrite ``vpr_options`` from toolchain_wrapper command-line argument, as for now, there is no real usage for it, but in the future we are planning to add this option and update this backend with ``vpr_options`` that will overwrite default options.

@olofk What do you think about it? Is missing ``vpr_options`` is a blocker for this PR and I should add this now?

olofk: Sorry. The real issue was that I couldn't find time to finish the review. But yes, I would prefer to just call it vpr_options. Ideally, the wrapper should only set values that it absolutely has to set and leave the rest up to the user

edalize/symbiflow.py
```diff
@@ -0,0 +1,145 @@
...
+            if f.file_type in ["SDC"]:
+                timing_constraints.append(f.name)
+            if f.file_type in ["PCF"]:
+                pins_constraints.append(f.name)
+            if f.file_type in ["xdc"]:
+                placement_constraints.append(f.name)
+            if f.file_type in ["user"]:
+                user_files.append(f.name)
+
+        # copy user files to builddir
```
olofk, line 97: Can we remove this as well?

olofk:
Ok, so we're getting really close now. Renaming additional_vpr_options to vpr_options and removing the shutil.copy and I'm happy to merge

edalize/symbiflow.py
kamilrakoczy: Ok, done.

edalize/symbiflow.py
kamilrakoczy: done

olofk: Awesome! Thanks

olofk:
And we're done! Thanks everyone who helped out. It took a while, but I'm really excited about having a SymbiFlow backend now!

PR #192, opened 2020-09-27
add lattice radiant support

trabucayre:
Lattice radiant is required for nexus devices (crosslink and certus) and ice40ultraPlus.


olofk:
Fantastic! Can't find a single thing to complain about :) Many thanks. Picked and pushed!

PR #194, opened 2020-10-08
symbiflow: Add quicklogic support

kowalewskijan:
This PR adds support for `quicklogic` targets to `symbiflow` toolchain.


kowalewskijan:
@olofk I can see that the CI failed in `test_modelsim.py` and it didn't directly fail because of changes in this PR, hence we could not confirm if this PR is green.

olofk:
Yes. The CI should be fixed now.

Thanks for this PR. I will add some comments but not finish the review today, FYI

edalize/templates/symbiflow/symbiflow-vpr-makefile.j2
```diff
@@ -28,22 +39,22 @@ SDC_OPTS = -s ${SDC}
 all: ${TOP}.bit
 
 ${TOP}.eblif:
-	symbiflow_synth -t ${TOP} -v ${VERILOG} -d ${BITSTREAM_DEVICE} -p ${PARTNAME} ${XDC_OPTS}
+	${SOURCE_OPTS} && ${TOOLCHAIN_PREFIX}synth -t ${TOP} -v ${VERILOG} -d ${BITSTREAM_DEVICE} ${PARTNAME_OPT} ${XDC_OPTS}
```
olofk, line 42: Drop the `&&` here (and below), otherwise this fails if environment_script is empty because the lines below will start with `&&` which causes syntax errors

kowalewskijan: Fixed

kowalewskijan:
@olofk could you finish the review?

olofk:
Sorry. I was hoping to do a test run on real hardware before finishing the review, but that will still be a couple of days away, and building the FPGA image works fine with the backend now, so all good

Many thanks for the contribution. Picked and pushed!

PR #196, opened 2020-10-23
Add environment script for icestorm

wtatarski:
This PR enables environment script option for icestrom toolchain.

Signed-off-by: Wojciech Tatarski <wtatarski@antmicro.com>


olofk:
Sorry for not getting to this earlier. Looking at it now I will decline it for two reasons. The first one is that bash scripts are a linux-only thing so this will not work on e.g. windows. In some cases I would think it's ok anyway. But I also added a patch last week that lets you define a custom launcher to run the EDA tools. This is a more generalized solution than sourcing a script.

This is how you can use this to source a file called `sourceme` before running yosys, nextpnr and icepack

1. Save the following code to `source_launcher.py` and make executable

```python
#!/usr/bin/python
import subprocess
import sys

if sys.argv[1] in ['yosys', 'nextpnr-ice40', 'icepack']:
    
    prefix="source /path/to/sourceme && "
    subprocess.call(prefix+' '.join(sys.argv[1:]), shell=True)
else:
    subprocess.call(sys.argv[1:])
```

2. Run with environment variable `EDALIZE_LAUNCHER` set to the path of the source_launcher.py script

I will close this issue now but please let me know if you need any help with the EDALIZE_LAUNCHER method

PR #198, opened 2020-11-02
Capture tool output

### LINKED ISSUE #21: Getting stdout from tool

robtaylor:
Support for capturing tool & script stdout and stderr. Closes #21

### DIFF
--- a/edalize/edatool.py
+++ b/edalize/edatool.py
@@ -10,6 +10,59 @@ logger = logging.getLogger(__name__)
 
 if sys.version[0] == '2':
     FileNotFoundError = OSError
+try:
+    import msvcrt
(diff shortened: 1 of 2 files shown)


olofk:
Fantastico! Thanks for fixing this. Picked and pushed!

Issue #203, opened 2020-12-14
Run scripts only executing first script (pre_build, etc..)


PeterCDMcLean:
I noticed that pre_build script list is only executing the first script in the list. I looks like the 'return' in the following snippet of edatool.py of version 0.2.3 is probably the issue:

```
       def _run_scripts(self, scripts, hook_name):
        for script in scripts:
            _env = self.env.copy()
            if 'env' in script:
                _env.update(script['env'])
            logger.info("Running {} script {}".format(hook_name, script['name']))
            logger.debug("Environment: " + str(_env))
            logger.debug("Working directory: " + self.work_root)
            try:
                cp = run(script['cmd'],
                                    cwd = self.work_root,
                                    env = _env,
				    capture_output=not self.verbose,
                                    check = True)
            except FileNotFoundError as e:
                msg = "Unable to run {} script '{}': {}"
                raise RuntimeError(msg.format(hook_name, script['name'], str(e)))
            except subprocess.CalledProcessError as e:
                msg = "{} script '{}': {} exited with error code {}".format(hook_name, script['name'], e.cmd, e.returncode)
                logger.debug(msg)
                if e.stdout:
                    logger.info(e.stdout.decode())
                if e.stderr:
                    logger.error(e.stderr.decode())
                    logger.debug("=== STDERR ===")
                    logger.debug(e.stderr)
                raise RuntimeError(msg)

            return (cp.returncode, cp.stdout, cp.stderr) #<--- probably this line. Looks to be returning after first script
```

olofk:
I believe we can close this now wrt #204 . Please reopen otherwise

PR #204, opened 2020-12-14
Remove early return from run_scripts

PeterCDMcLean:
Fix for Issue #203


olofk:
Many thanks. I also had another user reporting a bug around this who was helped by your patch.

PR #206, opened 2021-01-09
Add license headers to files

imphil:
Add a license header to all files, similar to what we did in FuseSoC.


olofk:
This one was easy! Thanks

PR #208, opened 2021-01-12
Suppress echo shell command from GHDL Makefile.

m-kru:
Without supression it prints the same message twice.


olofk:
Super. Thanks for cleaning this up. Picked and pushed

PR #211, opened 2021-01-22
Adds 'source_mgmt_mode' tool option to Vivado

Bucknalla:
Adds ability to pass source_mgmt_mode (None | DisplayOnly | All) to Vivado as a tool_option to allow for better control when building block designs from tcl scripts.


olofk:
I prefer to keep the number of options exposed to an absolute minimum. Is it possible to set this with a tcl file instead?

Bucknalla:
@olofk the problem is that in the standard project template `source_mgnt_mode` is set right before launching a run so even if `source_mgnt_mode` set in a tcl source file, it is overwritten before the run begins by this line https://github.com/olofk/edalize/blob/6ebf41f422c076d13cd98c37b678cd9f2a2580f6/edalize/templates/vivado/vivado-project.tcl.j2#L42

olofk:
Ah, right, I see. IIRC that line was added because Vivado apparently uses a different parser when figuring out the file hierarchy and that one caused Vivado to crash on someone's code. Think it was @imphil or @wallento who had that problem.

So in regard to that, I think your solution is the right one as it leaves the default to the current value but allows users to override.

I'm a bit surprised to see that the tests still pass with this patch. Was expecting some reference had to be updated as well. Anyway, picked and pushed. Thanks for you contributions

imphil:
https://github.com/olofk/fusesoc/commit/86c5808dd66f998cbd1f04c59d332ade90d456e7 explains the motivation to disable the source management mode. It's likely that Vivado fixed things in the meantime. I wonder: what motivates you to enable the source management? Does it help you in GUI mode, or are there other benefits to it?

imphil:
> I'm a bit surprised to see that the tests still pass with this patch.

The default behavior doesn't change, and hence existing tests won't break. At the same time, it would be nice to add tests for this new feature.

Bucknalla:
@imphil I have some block designs (built from a tcl script) that incorporate standalone rtl. This ends up with an unusual hierarchy and compile order so it's simply easier to allow Vivado to manage the compile order rather than to fight against it.

PR #212, opened 2021-01-26
Add support for Microsemi Libero backend

carlosedp:
This PR adds support for Microsemi Libero backend.

Still a Work in progress for review and tips.

To Do items:

- [ ] ~Libero requires different `hex` formatted files. Apparently Intel HEX is supported.~ Added workdir to Synth tool so .hex files get loaded correctly.
- [ ] Tox tests passes on Linux but fails on Mac. There is a file on Icarus project (vvp.cmd) that gets wrong parameters, my guess is the make command generates things differently.


olofk:
Hi @carlosedp Looks like you're on track. If possible, it would be great to separate the project creation and build into different tcl scripts so that users that prefer the GUI have an option to get a project file from edalize that they can open themselves. I haven't used Libero in almost ten years, so I'm not sure how these things work anymore.

I would also like to see if we can minimize/avoid hardcoded options like `-project_description {} -block_mode 0 -standalone_peripheral_initialization 0 -instantiate_in_smartdesign 1 -ondemand_build_dh 1 -linked_files_root_dir_env {} -hdl {VERILOG}`

Other than that it looks just fine after a quick review. Good job! Looks like I need to install Libero now so I can try out my MicroSemi board that has been collecting dust since 2016 :)

carlosedp:
Awesome, gonna take a look at it tomorrow and clean it up as you suggested.

I still haven't loaded the script into Libero to test it out... can I call one script from another? Have to investigate that.

carlosedp:
@olofk Separate project generation/file import and project run (synth, pnr, etc) on separate scripts.

Some of those parameters are default and also hardcoded like on Litex (https://github.com/enjoy-digital/litex/blob/42025dcbfa71d257cc678ddeb9a6407405d356ae/litex/build/microsemi/libero_soc.py#L70). I don't think it might be customized requiring variables.

carlosedp:
Due to some errors after adding a PDC file, I've fixed the tests for the other tools.
Could you take a look? I also rebased on master.

carlosedp:
Weirdly enough, it passes here:

```
❯ python3 -m pytest
==================================================================== test session starts =====================================================================
platform darwin -- Python 3.9.1, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /Users/cdepaula/repos/edalize
collected 81 items

tests/test_ascentlint.py .                                                                                                                             [  1%]
tests/test_diamond.py ..                                                                                                                               [  3%]
tests/test_edam.py ......                                                                                                                              [ 11%]
tests/test_ghdl.py ...                                                                                                                                 [ 14%]
tests/test_icarus.py ..                                                                                                                                [ 17%]
tests/test_icestorm.py ......                                                                                                                          [ 24%]
tests/test_ise.py ..                                                                                                                                   [ 27%]
tests/test_isim.py .                                                                                                                                   [ 28%]
tests/test_libero.py .                                                                                                                                 [ 29%]
tests/test_modelsim.py .                                                                                                                               [ 30%]
tests/test_morty.py .                                                                                                                                  [ 32%]
tests/test_quartus.py .                                                                                                                                [ 33%]
tests/test_radiant.py ..                                                                                                                               [ 35%]
tests/test_reporting.py ...............................                                                                                                [ 74%]
tests/test_rivierapro.py .                                                                                                                             [ 75%]
tests/test_spyglass.py ..                                                                                                                              [ 77%]
tests/test_symbiflow.py .                                                                                                                              [ 79%]
tests/test_symbiyosys.py .                                                                                                                             [ 80%]
tests/test_vcs.py ...                                                                                                                                  [ 83%]
tests/test_veribleformat.py .                                                                                                                          [ 85%]
tests/test_veriblelint.py .                                                                                                                            [ 86%]
tests/test_verilator.py ...                                                                                                                            [ 90%]
tests/test_vivado.py ...                                                                                                                               [ 93%]
tests/test_vunit.py ..                                                                                                                                 [ 96%]
tests/test_xcelium.py .                                                                                                                                [ 97%]
tests/test_xsim.py ..                                                                                                                                  [100%]

==================================================================== 81 passed in 29.42s =====================================================================
```

imphil:
You need to add the newly added files to https://github.com/olofk/edalize/blob/master/setup.py#L11.

If you run the test through tox you'll see the same errors as in CI. Running only pytest skips the packaging step, hiding problems like the one you're seeing.

carlosedp:
It passes too... look:

```
❯ tox
GLOB sdist-make: /Users/cdepaula/repos/edalize/setup.py
py3 create: /Users/cdepaula/repos/edalize/.tox/py3
py3 installdeps: pytest
py3 inst: /Users/cdepaula/repos/edalize/.tox/.tmp/package/1/edalize-0.2.3.zip
py3 installed: attrs==20.3.0,edalize @ file:///Users/cdepaula/repos/edalize/.tox/.tmp/package/1/edalize-0.2.3.zip,iniconfig==1.1.1,Jinja2==2.11.2,MarkupSafe==1.1.1,numpy==1.19.5,packaging==20.8,pandas==1.2.1,pluggy==0.13.1,py==1.10.0,pyparsing==2.4.7,pytest==6.2.2,python-dateutil==2.8.1,pytz==2020.5,six==1.15.0,toml==0.10.2
py3 run-test-pre: PYTHONHASHSEED='3763333780'
py3 run-test: commands[0] | pytest
==================================================================== test session starts =====================================================================
platform darwin -- Python 3.9.1, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
cachedir: .tox/py3/.pytest_cache
rootdir: /Users/cdepaula/repos/edalize
collected 81 items

tests/test_ascentlint.py .                                                                                                                             [  1%]
tests/test_diamond.py ..                                                                                                                               [  3%]
tests/test_edam.py ......                                                                                                                              [ 11%]
tests/test_ghdl.py ...                                                                                                                                 [ 14%]
tests/test_icarus.py ..                                                                                                                                [ 17%]
tests/test_icestorm.py ......                                                                                                                          [ 24%]
tests/test_ise.py ..                                                                                                                                   [ 27%]
tests/test_isim.py .                                                                                                                                   [ 28%]
tests/test_libero.py .                                                                                                                                 [ 29%]
tests/test_modelsim.py .                                                                                                                               [ 30%]
tests/test_morty.py .                                                                                                                                  [ 32%]
tests/test_quartus.py .                                                                                                                                [ 33%]
tests/test_radiant.py ..                                                                                                                               [ 35%]
tests/test_reporting.py ...............................                                                                                                [ 74%]
tests/test_rivierapro.py .                                                                                                                             [ 75%]
tests/test_spyglass.py ..                                                                                                                              [ 77%]
tests/test_symbiflow.py .                                                                                                                              [ 79%]
tests/test_symbiyosys.py .                                                                                                                             [ 80%]
tests/test_vcs.py ...                                                                                                                                  [ 83%]
tests/test_veribleformat.py .                                                                                                                          [ 85%]
tests/test_veriblelint.py .                                                                                                                            [ 86%]
tests/test_verilator.py ...                                                                                                                            [ 90%]
tests/test_vivado.py ...                                                                                                                               [ 93%]
tests/test_vunit.py ..                                                                                                                                 [ 96%]
tests/test_xcelium.py .                                                                                                                                [ 97%]
tests/test_xsim.py ..                                                                                                                                  [100%]

==================================================================== 81 passed in 31.48s =====================================================================
__________________________________________________________________________ summary ___________________________________________________________________________
  py3: commands succeeded
  congratulations :)
```

The two template files were already on setup.py. Any ideas?

carlosedp:
Squashed some commits and updated script generation to just link to the original files instead of importing them all.
Only constraint is imported to be loaded into the PnR stage.
Don't know why it's still failing, locally all tests pass on both pytest and tox.

carlosedp:
No idea why but tests pass on Linux (using a Docker container) but fails if run natively on MacOS.

If using `docker run -it --rm -v $PWD:/host --rm python:3 sh -c "cd /host; pip3 install tox; tox;"`, it passes. If I run `tox` natively on Mac, Icarus tests fails.

Same versions:

```
platform darwin -- Python 3.9.1, pytest-6.2.2, py-1.10.0, pluggy-0.13.1

platform linux -- Python 3.9.1, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
```

edalize/libero.py
```diff
@@ -0,0 +1,136 @@
+import logging
+import os.path
+import os
+import platform
+import subprocess
+import re
+import xml.etree.ElementTree as ET
+from functools import partial
+from edalize.edatool import Edatool
```
olofk, line 9: logging and Edatool are the only imports that are used in the code. Remove the rest

edalize/libero.py
```diff
@@ -0,0 +1,136 @@
...
+
+    argtypes = ['vlogdefine', 'vlogparam']
+
+    tool_options_defaults = {
+        'speed': '-1',
+        'dievoltage': '1.0',
+        'range': 'IND',
+        'defiostd': 'LVCMOS 1.8V',
+        'hdl': 'VERILOG',
+    }
```
olofk, line 55: I'm always a bit sceptical when it comes to setting default values. Are these the most common values? Do all devices support dievoltage 1.0 etc. Unless we can confirm these are reasonable defaults I think it's better to leave them completely to the user and not write them to the template unless they have been set by the user

edalize/libero.py
```diff
@@ -0,0 +1,136 @@
...
+                            % (key, str(default_value)))
+                self.tool_options[key] = default_value
+
+    """ Initial setup of the class
+
+    This calls the parent constructor, but also identifies whether
+    the current system is using a Standard or Pro edition of Quartus.
+    """
+
+    def __init__(self, edam=None, work_root=None, eda_api=None):
```
olofk, line 70: We can remove the constructor here. The super class constructor will be called automatically for us

edalize/libero.py
```diff
@@ -0,0 +1,136 @@
...
+        escaped_name = self.name.replace(".", "_")
+
+        template_vars = {
+            'name': escaped_name,
+            'src_files': src_files,
+            'incdirs': incdirs,
+            'tool_options': self.tool_options,
+            'toplevel': self.toplevel,
+            'generic': self.generic,
+            'prj_root': "../prj",
```
olofk, line 98: All files needs to reside with in the work directory. If I get this correctly, files would end up in a sibling directory to the work root.

edalize/templates/libero/libero-project.tcl.j2
```diff
@@ -0,0 +1,29 @@
...
+
+set_device -family {{op}}{{tool_options.family}}{{cl}} -die {{op}}{{tool_options.die}}{{cl}} -package {{op}}{{tool_options.package}}{{cl}} -speed {{op}}{{tool_options.speed}}{{cl}} -die_voltage {{op}}{{tool_options.dievoltage}}{{cl}} -part_range {{op}}{{tool_options.range}}{{cl}} -adv_options {IO_DEFT_STD:{{tool_options.defiostd}}{{cl}} -adv_options {RESTRICTPROBEPINS:1} -adv_options {RESTRICTSPIPINS:0} -adv_options {SYSTEM_CONTROLLER_SUSPEND_MODE:0} -adv_options {TEMPR:{{tool_options.range}}{{cl}} -adv_options {VCCI_1.2_VOLTR:EXT} -adv_options {VCCI_1.5_VOLTR:EXT} -adv_options {VCCI_1.8_VOLTR:EXT} -adv_options {VCCI_2.5_VOLTR:EXT} -adv_options {VCCI_3.3_VOLTR:EXT} -adv_options {VOLTR:{{tool_options.range}}{{cl}}
```
olofk, line 7: Remove this. Since we need to supply most of these parameters to new_project, I don't think we need a separate call to set_device

edalize/templates/libero/libero-project.tcl.j2
```diff
@@ -0,0 +1,29 @@
...
+
+# Link directly to source files
+create_links \
+         -convert_EDN_to_HDL 0 \
```
olofk, line 11: I haven't worked with Libero in ten years and I find the documentation a bit confusing. The hdl_source_folder is not documented and I can't really find what the difference is between import_files and create_links

But with that said we can't import a directory like that. In many cases it's totally fine to have an undefined order between files, but FuseSoC and Edalize keeps guarantees on the file order because there are many cases where that matters. So we need to list all files. Question is if create_links or import_files should be used.

We can drop the `convert_EDN_to_HDL` option as that is only valid for importing EDIF files as I understand it

We also need to set the correct library (at least for VHDL files) instead of hardcoding work. The library name can be found in the `logical_name` property of files

edalize/libero.py
```diff
@@ -0,0 +1,136 @@
...
+        file_types = {
+            'PDC': 'constraint/io/',
+        }
+        _file_type = f.file_type.split('-')[0]
+        if _file_type in file_types:
+            filename = f.name.split("/")[-1]
+            return file_types[_file_type] + filename
+        return ''
+
+    def build_main(self):
```
olofk, line 132: Here you want to call libero twice, either directly or create a makefile in the configure step that does it.
First you call `libero SCRIPT:<name of project script>` and then `libero SCRIPT:<name of run script>` If you put these commands as the default target in a makefile, you can drop the build_main function, as the superclass will call make by default

edalize/templates/libero/libero-project.tcl.j2
```diff
@@ -0,0 +1,29 @@
...
+        -convert_EDN_to_HDL 0 \
+{% for src_file in src_files if src_file|src_file_filter%}
+        {{src_file|src_file_filter}}{{cl}} \
+{% endfor %}
+
+build_design_hierarchy
+set_root -module {{op}}{{toplevel}}::work{{cl}}
+
+{% for src_file in src_files if src_file|pdc_file_filter%}
+organize_tool_files -tool {PLACEROUTE} -file {{op}}{{prj_root}}/{{src_file|pdc_file_filter}}{{cl}} -module {{op}}{{toplevel}}::work{{cl}} -input_type {constraint}
```
olofk, line 26: I haven't really figured out myself how this fits together so I have no idea if it's write or wrong. Just looking for an explanation :)

edalize/templates/libero/libero-run.tcl.j2
```diff
@@ -0,0 +1,14 @@
+# This file is generated by Edalize.
+# Microsemi Tcl Script
+# Libero
+
+run_tool -name {SYNTHESIZE}
```
olofk, line 5: This needs to start with `open_project {path/to/project/generated/by/first/tcl/file}`

edalize/libero.py
carlosedp: Great, I'm too used to Go where imports not used generate compile errors or VSCode plugin manages import automatically. :)
Cleaned.

edalize/libero.py
carlosedp: Since the tcl file will be placed in generated dir (ex. `fusesoc_utils_blinky_1.0/polarfireeval_es-libero`, I've set the script to create a `prj` dir to save project in the `fusesoc_utils_blinky_1.0` dir (for example).
Is it ok?

edalize/templates/libero/libero-project.tcl.j2
carlosedp: Most scripts I saw create the project and also set the device. I'll test without the set_device command to see it it inherits from new_project.

edalize/libero.py
carlosedp: Shouldn't it be in `run_main` instead of `build_main`?
I ask because I usually build it all in one machine (a Mac for example) and then move files to another machine (Windows) to use on Libero.

edalize/templates/libero/libero-project.tcl.j2
carlosedp: This adds the constraint file to the PnR phase thru the script and sets the toplevel module

edalize/templates/libero/libero-run.tcl.j2
carlosedp: Done.

edalize/templates/libero/libero-project.tcl.j2
carlosedp: I was using import_files before but some files didn't get imported like a `.vh` file from corescore.. will investigate this further.
Also I understand that Libero puts all user files under `work` library. Should we use from logical_name?

edalize/libero.py
carlosedp: Also in my case, I'd use "fusesoc build" to only build the files but I get a deprecation message telling to use "run".
If I use "run", "libero" doesn't exists in this machine.

edalize/templates/libero/libero-project.tcl.j2
carlosedp: Also when changing the logic to import the files, I don't see files marked as "is_include_file" like "rtl/serv_params.vh" from serv being listed on src_files.

How this should be handled?

edalize/libero.py
olofk: Yes, let's remove all of them. It's better that more experience Libero users tell us later which defaults that make sense

edalize/libero.py
olofk: When FuseSoC sets up the directories, it will look like you describe. The problem is for users who use Edalize standalone since the prj directory will be outside of their workspace.

The best way would had been to generate the project files in `.`, like pretty much all the other EDA tools do but I noticed that Libero doesn't support that, so we have to put the `prj` directory below the work root, like

```
.
├── polarfireeval_es-libero
│   ├── prj                 <---- project dir will be created here      
│   ├── fusesoc_utils_blinky_1.0.eda.yml
│   ├── fusesoc_utils_blinky_1_0-project.tcl
│   └── fusesoc_utils_blinky_1_0-run.tcl
└── src
    └── fusesoc_utils_blinky_1.0
        ├── blinky.v
        └── polarfireeval
            ├── blinky_polarfireeval.v
            └── polarfireeval.pdc
```

edalize/templates/libero/libero-project.tcl.j2
olofk: Give it a try. Maybe it is needed, but my experience tells me that in many cases these kind of scripts have just been copy-pasted around without any attempts at cleanup

edalize/templates/libero/libero-project.tcl.j2
olofk: Verilog include files work pretty similar to C include files so instead of putting them in the list of files to compile,  we just make sure that the directories of these files can be found. The `get_fileset_files` helper returns a list of files that are not include files and a list of directories where include files are found. The problem here is that I can't find a way in Libero to set include directories so I'm not really sure what to do yet. I need to investigate this further

And regarding library/logical_name, `work` is the default name, so we can use that unless the user has explicitly set a logical_name

edalize/libero.py
olofk: The three stages are defined like this
1. configure: The EDAM file is converted to tool-specific project files
2. build: A binary is being built. For an FPGA flow this is the FPGA image. For a simulation flow this will be the compiled simulation model
3. run: Not all flows define the run phase. A simulator will run the simulation here. Some FPGA flows will program a board at this stage, but most just leave it empty. You can check out my [award-winning Edalize video](https://www.youtube.com/watch?v=HuRtkpZqB34) for more details :)
So in your case you would only do the configure step on the first computer. You can do that by running `fusesoc run --target=yourtarget --setup your_core`. Then you can copy the build directory to the computer where you have libero

edalize/templates/libero/libero-project.tcl.j2
olofk: I see the documentation now. I wonder when we will set tool to something other than PLACEROUTE. I need to dive deeper into Libero constraint files. What happens if we leave this out? Will the PnR not pick it up? Could we set it to all tools and have the tools just ignore constraint files that it doesn't recognize?

edalize/templates/libero/libero-project.tcl.j2
olofk: I realized a similar problem is how to set top-level parameters and defines which are things we need to do as well. I'm starting to suspect we need to look in to the `configure_tool` command and send all these things directly to synplify_pro, which is the tool that handles the actual synthesis inside of Libero. I am not sure yet though

carlosedp:
I've pushed a couple changes adding suggestions.
On import, I see an error "Error: the module priority_encoder is instantiating itself.".
Another thing is that files marked with `is_include_file` are not listed and not added to the tcl script. Also I don't know how to add import dirs in Libero.

edalize/templates/libero/libero-project.tcl.j2
carlosedp: Just tested without the set_device and it loads fine. Already on last push.

edalize/libero.py
```diff
@@ -0,0 +1,136 @@
...
+                        {'name': 'range',
+                         'type': 'String',
+                         'desc': 'FPGA temperature range (e.g. IND)'},
+                        {'name': 'defiostd',
+                         'type': 'String',
+                         'desc': 'FPGA default IO std (e.g. "LVCMOS 1.8V"'},
+                        {'name': 'hdl',
+                         'type': 'String',
+                         'desc': 'Default HDL (e.g. "VERILOG"'}
+                    ]
```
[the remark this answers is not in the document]
carlosedp: Removed them as default parameters and added some validation for mandatory fields to print error in case they are not defined.

edalize/libero.py
carlosedp: See next comment.

olofk:
Thanks @carlosedp. It keeps looking better and better. Some of the remaining problems I'm really not sure how to solve yet. Out of all the 26 backends this has been the most confusing one so far. But I'll keep digging.

olofk:
Alright! I think I got it figured out now. We need to pass include directories, defines and verilog parameters directly to synplify via the configure_tool commands. So for example, adding the following line to the project creation tcl file would set the `clk_freq_hz` verilog parameter
`configure_tool -name {SYNTHESIZE} -params {SYNPLIFY_OPTIONS:set_option -hdl_param -set "clk_freq_hz=10000"}`

olofk:
Ok, so we need to write something like this in the project tcl file to handle top-level parameters, defines and include dirs

`configure_tool -name {SYNTHESIZE} -params {SYNPLIFY_OPTIONS:{params} {defines} {incdirs}}`
where params is something like
`'\n'.join(['set_option -hdl_param -set " '+k + '='+v+'"' for k,v in self.vlogparams.items()])`
and defines are set similarly by
`'\n'.join(['set_option -hdl_define -set " '+k + '='+v+'"' for k,v in self.vlogdefines.items()])`

Include directories is a bit trickier since they are not relative to the work root. synplify is executed from the created_project_dir/synthesis, so we must do something like

`'\n'.join(['set_option -include_path ../../' + d for d in incdirs])`

Unless Libero uses backslash as path separator on windows and we need to replace ../../ with a proper os.path.join('..','..',d)

Probably still need some experimenting to get things right. Haven't tested everything yet so I don't know e.g. how strings need to be quoted for parameters and defines

olofk:
Another annoying thing I discovered is that it seems like libero has a separate HDL parser outside of the one in synplify pro that is used to calculate the design hierarchy. If that fails, then we get no project file even if synplfiy can handle the files just fine. Would be fantastic if we could disable that but it doesn't seem like it. Vivado has a TCL option for that but haven't seen anything in Libero so far

carlosedp:
I've been battling with this for a while and while it's possible to pass the include path via options to Synth, having relative paths is a total pain. Since the include files are not imported into the project, it's required to reference them in the original `src` path and I can't figure out how to do this.
 
When I tried just linking them to the `src` dir, it synthesized correctly with no errors related to importing an include.

Should I revert back to linking it?

tmeissner:
I didn't check the whole PR, and I don't have any experience with edalize and all this Python stuff.

We set options for Synplify in two ways: With the `configure_tool` command and by including a separate tcl script which contains Synplify options which Libero doesn't know or directly supports by it's tcl interpreter.

It looks similar to that:

```tcl
configure_tool   \
  -name {SYNTHESIZE} \
  -{RETIMING:true} \
  -SYNPLIFY_TCL_FILE:../../config/synplify.tcl}  \
} 
```

The references synplify.tcl file looks similar to this
 
```tcl
set_option -synbolic_fsm_compiler 1;
set_option -resource_sharing 1;
set_option -hdl_param -set DEBUG false;
```

edalize/libero.py
carlosedp: Done on last push.

edalize/libero.py
carlosedp: Great! Fixed on last push.

edalize/templates/libero/libero-project.tcl.j2
carlosedp: Yep, PnR would not use the constraint file if not set here... would require user to configure it thru "Manage Constraints" tool.

carlosedp:
Yay, thanks @tmeissner! It worked.. I created the Synplify TCL script on template generation and configured Libero to use it passing it to Symplify.

@olofk, I've fixed some parts related to your comments but we have a couple more problems to tackle:

* On project import, I see an error saying that `Error: the module priority_encoder is instantiating itself.` on Corescore. In the bottom is the full log. Maybe we discuss this on Corescore PR (although it might require changes to Edalize)
* Blinky works perfecly but Corescore doesn't. One issue is that Libero requires the `.hex` files to be in Intel format. I'll update Corescore PR with details.
* Tox tests passes on Linux but fails on Mac. There is a file on Icarus project (`vvp.cmd`) that gets wrong parameters, my guess is the `make` command generates things differently.

edalize/templates/libero/libero-project.tcl.j2
carlosedp: Fixed the import paths passing it thru a TCL script to Symplify by using a Libero config.
Thanks for you pointing that out.

carlosedp:
@olofk currently I'm generating all `vlogparam`s and `vlogdefine`s in the Synthesize TCL script but I wonder if some params should be configured on PnR or other phases... ?
Any idea @tmeissner?

tmeissner:
I'm not the Verilog guy, but if these `vlogparam`s and `vlogdefine`s are like generics in VHDL, then Synplify have to know them.

carlosedp:
Tests seem flaky as everything passes here, on both MacOS and Linux. Might be something about Python 3.5?

olofk:
@tmeissner Thanks for the feedback. It seems we got include paths, vlog parameters (and likely vhdl generics with that) and defines right. One thing I've been wondering about is constraints. The docs is a bit cryptic around that. We got pdc constraints for io and sdc constraints for timing (and other things like settings for bitstream generation?). Are we required to specify for each constraint file when they should be used (e.g. SYNTHESIZE, PLACEROUTE, VERIFYTIMING) or can we just pass all files to all tools and they will figure out themselves which are approriate?

@carlosedp Regarding the hex files, I think I just realized what is going on. CoreScore copies the hex files to work_root. All the other EDA tools are using this directory as its working directory which means that $readmemh("file.hex",..) works fine since the file will be in the current working directory. However! Libero will launch synplify from <work_root>/prj/synthesis which means it will not find the files. I think we unconditionally need to add ../../ as an include directory to synplify to get around this.

Will add some more comments inline in the code

edalize/libero.py
```diff
@@ -0,0 +1,181 @@
...
+                        {'name': 'defiostd',
+                         'type': 'String',
+                         'desc': 'FPGA default IO std (e.g. "LVCMOS 1.8V"'},
+                        {'name': 'hdl',
+                         'type': 'String',
+                         'desc': 'Default HDL (e.g. "VERILOG"'},
+                        {'name': 'synth_options',
+                         'type': 'String',
+                         'desc': 'Additional Synthesize tool options separated by commas (e.g. "RETIMING:true,CLOCK_GLOBAL:2")'},
+                        {'name': 'pnr_options',
```
olofk, line 41: While this helps for setting pnr options, there are just too many things that different users want to change and we can't expose all these things. What we do instead is to make it possible for users to add any extra options they want to set as tcl files that we source from the project creation script. We do this by looking for files with `file_type : tclSource` and source them. You can compare with e.g. the quartus, ise or vivado backends which offer the same possibilities.

The situation is a bit more complicated with the synthesis options as we need to mix some hard-coded commands together with user-supplied ones. I think I have an idea but it's not super pretty

edalize/libero.py
```diff
@@ -0,0 +1,181 @@
...
+                            % (key, str(default_value)))
+                self.tool_options[key] = default_value
+
+    def _check_mandatory_options(self):
+        shouldExit = 0
+        for key in self.mandatory_options:
+            if not key in self.tool_options:
+                logger.error("Libero option \"%s\" must be defined", key)
+                shouldExit = 1
+        if shouldExit:
```
olofk, line 73: We can't use exit in the backends. Raise RuntimeError instead

edalize/templates/libero/libero-project.tcl.j2
```diff
@@ -0,0 +1,98 @@
...
+# Libero
+
+# Create a new project with device parameters
+new_project -location {{op}}{{prj_root}}{{cl}} -name {{name}} -project_description {} -block_mode 0 -standalone_peripheral_initialization 0 -instantiate_in_smartdesign 1 -ondemand_build_dh 1 -hdl {{op}}{{tool_options.hdl}}{{cl}} -family {{op}}{{tool_options.family}}{{cl}} -die {{op}}{{tool_options.die}}{{cl}} -package {{op}}{{tool_options.package}}{{cl}} {% if tool_options.speed -%} -speed {{op}}{{tool_options.speed}}{{cl}}{{sp}}{%- endif %}{% if tool_options.dievoltage -%} -die_voltage {{op}}{{tool_options.dievoltage}}{{cl}}{{sp}}{%- endif %}{% if tool_options.range -%} -part_range {{op}}{{tool_options.range}}{{cl}}{{sp}}{%- endif %}{% if tool_options.defiostd -%} -adv_options {IO_DEFT_STD:{{tool_options.defiostd}}{{cl}}{{sp}}{%- endif %}-adv_options {RESTRICTPROBEPINS:1} -adv_options {RESTRICTSPIPINS:0} -adv_options {SYSTEM_CONTROLLER_SUSPEND_MODE:0} -adv_options {TEMPR:{{tool_options.range}}{{cl}} -adv_options {VCCI_1.2_VOLTR:EXT} -adv_options {VCCI_1.5_VOLTR:EXT} -adv_options {VCCI_1.8_VOLTR:EXT} -adv_options {VCCI_2.5_VOLTR:EXT} -adv_options {VCCI_3.3_VOLTR:EXT} -adv_options {VOLTR:{{tool_options.range}}{{cl}}
```
olofk, line 6: Drop all the `adv_options`. I believe most of those can be set in otherwise if users need them. Also drop
-block_mode 
-standalone_peripheral_initialization
-instantiate_in_smartdesign
-ondemand_build_dh

In fact, I diffed to project files and most of these options doesn't seem to do anything at all

edalize/templates/libero/libero-project.tcl.j2
```diff
@@ -0,0 +1,98 @@
...
+
+# Import HDL sources and constraints
+import_files \
+        -convert_EDN_to_HDL 0 \
```
olofk, line 10: Drop -convert_EDN_to_HDL 0. It's not needed

edalize/libero.py
```diff
@@ -0,0 +1,181 @@
...
+                             template_vars)
+
+        # Render the Synthesize TCL file
+        self.render_template('libero-syn-user.tcl.j2',
+                             escaped_name + '-syn-user.tcl',
+                             template_vars)
+
+        logger.info("Cores and Libero TCL Scripts generated.")
+
+    def src_file_filter(self, f):
```
olofk, line 147: In this function we need to check vhdl (and verilog?) files if they have `logical_name`set to anything. For those files we need to add `-library {<logical_name>}` to the generated tcl

edalize/templates/libero/libero-project.tcl.j2
```diff
@@ -0,0 +1,98 @@
...
+        {% endfor -%}{%- endif %}
+        -params {SYNPLIFY_TCL_FILE:../../{{name}}-syn-user.tcl}
+
+puts "Configured Synthesize tool to use script: {{name}}-syn-user.tcl"
+puts "Configured Synthesize tool to include dirs:"
+{% if incdirs -%}{%- for dir in incdirs %}
+puts "- ../../{{dir}}"
+{% endfor -%}{% endif %}
+
+{%+ if pnr_options -%}
```
olofk, line 39: Drop this when we add support for tclSource

tmeissner:
> @tmeissner Thanks for the feedback. It seems we got include paths, vlog parameters (and likely vhdl generics with that) and defines right. One thing I've been wondering about is constraints. The docs is a bit cryptic around that. We got pdc constraints for io and sdc constraints for timing (and other things like settings for bitstream generation?). Are we required to specify for each constraint file when they should be used (e.g. SYNTHESIZE, PLACEROUTE, VERIFYTIMING) or can we just pass all files to all tools and they will figure out themselves which are approriate?

@olofk 
It's a long time ago that I've created that script, but this is what we do:

We first import all relevant constraint files.

```tcl
import_files -sdc    ../constraints/libero.sdc
import_files -io_pdc ../constraints/libero_io.pdc
import_files -fp_pdc ../constraints/libero_fp.pdc
import_files -sdc    ../constraints/synplify.sdc
```
And then we define variables for easier use. We use the paths after import, I don't know, if there is another, more elegant way. Too long ago :laughing: 

```tcl
set sdcl    ../work/libero/constraint/libero.sdc
set io_pdc  ../work/libero/constraint/libero_io.pdc
set fp_pdc  ../work/libero/constraint/libero_fp.pdc
set sdcs    ../work/libero/constraint/synplify.sdc
```
Finally, we use this variables when "organizing" the tool files before running them:

```tcl
organize_tool_files \
  -tool {SYNTHESIZE} \
  -file ${sdcs} \
  -module ${topmodule} \
  -input_type {constraint}

organize_tool_files \
  -tool {COMPILE} \
  -file ${sdcl} \
  -file ${io_pdc} \
  -file ${fp_pdc} \
  -module ${topmodule} \
  -input_type {constraint}

run_tool -name {SYNTHESIZE}

run_tool -name {COMPILE}
```

olofk:
@tmeissner Thanks. This is very useful info. So for the constraints we basically have two categories
1. Synthesis (sdc) constraints passed to synplify
2. Other constraints (I/O, timing, other stuff) passed to the COMPILE phase. 

I think the initial approach will be to actually pass all the files to the COMPILE phase and see if we can perhaps set synthesis constraint files through the TCL api (like include files, hdl_param, etc). I think that is a reasonable way to do it even if it doesn't cover all cases. Now I would like who came up with the idea to have a separate io_pdc and fp_pdc category where the latter is only valid for PolarFire devices. I'm not sure what to do here. Should we just check if the family is set to PolarFire and use import_files -fp_pdc and import_files -io_pdc otherwise?

edalize/libero.py
carlosedp: Done!

edalize/templates/libero/libero-project.tcl.j2
carlosedp: Done. I just kept default IO std since it's already a user parameter.

edalize/templates/libero/libero-project.tcl.j2
carlosedp: Done!

olofk:
@carlosedp I just noticed a good thing about configure_tool that I think will solve things. If you run the command several times, the commands are added. This means we can have a default `configure_tool -name {SYNTHESIZE}` that loads the tcl file with our include files, hdl_params and hdl_defines.

If a user wants to add other parameters like CLOCK_ASYNC, the can just add a tcl file that runs configure_tool again with their extra options, like `configure_tool -name {SYNTHESIZE} -params {CLOCK_ASYNC:10}`

I think the problem comes if they want to add another tcl file with SYNPLIFY_TCL_FILE since that will overwrite the one that Edalize has created. But in that case, the user could potentially start their tcl script by manually loading our generated tcl file. A bit awkward but might be the best way for now

edalize/libero.py
carlosedp: I saw how it's done on Vivado but on Libero docs there is no "-library" setting to import files.

Here's the import_files doc:

image

tmeissner:
> Now I would like who came up with the idea to have a separate io_pdc and fp_pdc category where the latter is only valid for PolarFire devices. I'm not sure what to do here. Should we just check if the family is set to PolarFire and use import_files -fp_pdc and import_files -io_pdc otherwise?

@olofk
It's better not to think too much about the Libero build flow. Trust me, it's better for your mental health :smile: 

We use SmartFusion2 devices and also have such a floor planner constraint file. But it's all commented out, as we don't need the area / placement constraints anymore. It was important only during early stages of our development. So, I assume one can omit this file.

Hint: We developed these scripts years ago for version 11.5 or 11.6. But AFAIK we didn't changed that much all these years since then.

edalize/libero.py
carlosedp: Done, added sourcing user TCL scripts. Gonna remove the synth and pnr opts.

edalize/templates/libero/libero-project.tcl.j2
carlosedp: Done.

edalize/libero.py
carlosedp: For this, I'll have to iterate thru all "logical_name"s and import each since the expected script is:

```
import_files \
         -library {work2} \
         -hdl_source {file.vhdl} 
```

carlosedp: Did it, it would become something like:

```
...
# Import HDL sources and constraints
import_files \
        -sdc {sdc_file} \
        -hdl_source {sv_file.sv} \
        -hdl_source {vlog_file.v} \
        -hdl_source {vlog05_file.v} \
        -hdl_source {vhdl_file.vhd} \
        -hdl_source {vhdl2008_file} \
        -hdl_source {another_sv_file.sv} \
        -io_pdc {pdc_constraint_file.pdc} \

# Import HDL sources on specifit libraries (logical_names)
import_files \
        -library {libx} \
        -hdl_source {vhdl_lfile} \
        -hdl_source {vhdl_lfile2} \

import_files \
        -library {libxy} \
        -hdl_source {vhdl_lfile3} \
...
```

carlosedp:
What about this:
image

Corescore running on Polarfire with 5 SERV cores! :)
Adding the root work dir to Synplify solved the HEX loading :D

edalize/libero.py
carlosedp: Done! TCL scripts are sourced from main script.

olofk:
Whew! Finally found time for a last review. Thanks for all the work. Squashed and merged!

carlosedp:
Yay!

PR #213, opened 2021-01-26
experimental support for gowin backend

infphyny:
Experimental support for gowin backend


infphyny:
I will make a pull request to support tang nano board at fusesoc/blinky if the experimental support for gowin backend is accepted.

edalize/apicula.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @classmethod
+    def get_doc(cls, api_ver):
+        if api_ver == 0:
+            yosys_help = Yosys.get_doc(api_ver)
+            apicula_help = {
+                    'members' : [
+                        {'name' : 'pnr',
+                         'type' : 'String',
+                         'desc' : 'Select Place & Route tool. Legal values are *arachne* for Arachne-PNR or *next* for nextpnr. Default is arachne'}],
```
olofk, line 19: drop the pnr option. I don't think arachne-pnr works for gowin devices so nextpnr is our only option here

edalize/templates/apicula/apicula-makefile.j2
```diff
@@ -0,0 +1,21 @@
...
+NEXTPNR_OPTIONS     := {{ nextpnr_options|join(' ') }}
+
+all: $(TARGET).fs
+
+%.json: %.mk
+	$(MAKE) -f $<
+%.pack: %.json
+	nextpnr-gowin -l next.log --device $(DEVICE)  $(NEXTPNR_OPTIONS) --cst $(CST_FILE) --json $? --write $@
+%.fs: %.pack
+	gowin_pack -d GW1N-1 -o  $@ $^
```
infphyny, line 14: One problem I see now is that the model GW1N-1 is hardcoded in the makefile for packing.  Maybe it will be necessary to add an option in the .core file for packing like nextpnr_options : [GW1N-LV1QN48C6/I5] for example.
I will wait for your feedback before I start to work on this.

olofk: Good catch. I didn't see that in my first review. Yes, we need to remove this hardcoded value, but I'm a bit confused about how to do this. Most of the FPGA toolchain backends only require a single `part` option to decide which device to target. The yosys+nextpnr variants however are a bit different. Because yosys doesn't seem to care at all (and arachne-pnr/nextpnr used slighlty different names for parts) we have previously just defined the part in nextpnr_options. I was looking at how the other pack commands worked and it seems that gowin_pack is the only one that explicitly needs a part name here. So in that case we need to inform both nextpnr and gowin_pack about the part name, and I'm not sure yet if the are using the same naming schemes. I really want to avoid having some lookup table in the backend. Perhaps @pepijndevos could shed some light on this.

pepijndevos: Yea so Yosys doesn't care at all. Nextpnr takes the whole part number because it cares about timing and such. The packer only needs the family because they have different padding values. So the part number is `GW1N{variant}-(LV|UV){luts}{package}{speed}` And the packer only needs the `GW1N-{luts}` part. (eventually GW2A maybe). If this is problematic I'm open to a PR making the packer take the whole part number, or maybe even encode the part number in the nextpnr output somehow.

https://github.com/YosysHQ/nextpnr/blob/master/gowin/main.cc#L57

infphyny: I will continue to work on the gowin backend when there will be a solution for setting the part number for gowin_pack. I'm afraid that I have not enough skills right now to make a PR for the packer problem.

olofk: Yes, I think that's the best thing to do right now. It's a small thing but better to get right from the start.
@daveshah1, @clairexen how does ecppack and icepack handle this?

pepijndevos: It's actually a pretty trivial fix. Might do it this weekend. Basically just copy-paste the linked regex and add a new `--part` option that uses the regex to set the device to `GW1N-{luts}`. https://github.com/YosysHQ/apicula/blob/master/apycula/gowin_pack.py#L126

olofk: Ok, so if I understand correctly, with such a fix applied, we could pass the same part number to nextpnr and gowin_pack? If so, that would be great.

For reference, I just looked at icepack ecppack and they both seem to look at the `.device` line in the input file. Could that be an option for apicula too?

edalize/templates/apicula/apicula-makefile.j2
```diff
@@ -0,0 +1,21 @@
...
+TARGET   := {{ name }}
+CST_FILE := {{ cst_file }}
+NEXTPNR_OPTIONS     := {{ nextpnr_options|join(' ') }}
+
+all: $(TARGET).fs
+
+%.json: %.mk
+	$(MAKE) -f $<
+%.pack: %.json
+	nextpnr-gowin -l next.log --device $(DEVICE)  $(NEXTPNR_OPTIONS) --cst $(CST_FILE) --json $? --write $@
```
infphyny, line 12: For consistency with trellis back end,  in the line 12, --device should be defined in NEXTPNR_OPTIONS : [--device,GW1N-LV1QN48C6/I5]. Otherwise in the .core file the first parameter must be the device part number or nextpnr-gowin will stop with error. Furthermore, $(DEVICE) is an empty string.

edalize/templates/apicula/apicula-makefile.j2
pepijndevos: Done https://github.com/YosysHQ/apicula/commit/7624fdcd4fbebe880296aba4d54818e243ddb0a4

icepack and ecppack both use a custom text format for writing nextpnr output, while Apicula uses the built-in json output. This does not contain the part number. Writing a custom fasm exporter is not high on my priority list...

olofk: Awesome! That's all we need to proceed.

For my understanding, is the built-in json format fasm, or are those two different formats? And would the long-term solution be to add a device field tl this format?

edalize/templates/apicula/apicula-makefile.j2
olofk: Now that we can pass the same device identifier to both nextpnr and gowin_pack it's better to have an explicit device (or maybe part) option. We should also error out if the user didn't set this option

edalize/apicula.py
```diff
@@ -0,0 +1,81 @@
...
+                }
+
+        yosys = getattr(import_module("edalize.yosys"), 'Yosys')(yosys_edam, self.work_root)
+        yosys.configure()
+
+        cst_files = []
+        for f in src_files:
+            if f.file_type == 'CST':
+                cst_files.append(f.name)
+            elif f.file_type == 'user':
```
olofk, line 62: We can drop this elif

olofk:
I think I found the last minor issue. Other than that it looks great

edalize/templates/apicula/apicula-makefile.j2
pepijndevos: No fasm is a completely different thing: https://github.com/SymbiFlow/fasm
So moving from JSON to FASM might be a veeeeery long term plan, but it only really makes sense if there are multiple competing PnR implementations. JSON is very Nextpnr-specific. VTR uses fams, but I don't have any plans to add a VTR backend. So JSON is fine with me.

olofk: Got it. Thanks

edalize/templates/apicula/apicula-makefile.j2
```diff
@@ -0,0 +1,23 @@
+#Auto generated by Edalize
+
+TARGET   := {{ name }}
+CST_FILE := {{ cst_file }}
+NEXTPNR_OPTIONS     := {{ nextpnr_options|join(' ') }}
+DEVICE := {{ device|join('') }}
```
infphyny, line 6: In the .core I forgot to put device option between square brackets so maybe no need to add |join(''). I will wait your review to modify this and other things if needed.

edalize/apicula.py
```diff
@@ -0,0 +1,89 @@
...
+
+    argtypes = ['vlogdefine', 'vlogparam']
+
+    @classmethod
+    def get_doc(cls, api_ver):
+        if api_ver == 0:
+            yosys_help = Yosys.get_doc(api_ver)
+            apicula_help = {
+                    'lists' : [
+                    	{'name' : 'device',
```
olofk, line 17: device should be in `members` instead of `lists`. This way it will be a single value rather than a list of values. Compare e.g. with [the Quartus backend](https://github.com/olofk/edalize/blob/master/edalize/quartus.py#L30)

edalize/templates/apicula/apicula-makefile.j2
olofk: If the device option is moved to `members` as noted above, then I believe we can also drop this join

infphyny:
The only hurdle to use the gowin backend right now is that building the latest project apicula library need a specific version of proprietary gowin toolchain to produce the chip database for nextpnr-gowin. Version 1.9.1 is recommended and I have version 1.9.7 installed on my computer that fail to build database.  I found easier to install the current stable version of [project apicula](https://pypi.org/project/Apycula/)  then overwrite the gowin_pack.py file with the latest version from github repository.

pepijndevos:
You could use the Dockerfile to generate the latest database. It the image is also available at https://hub.docker.com/r/umarcor/xyce
Even easier if you're not planning to contribute to the project is to download the python package including the database from the latest github action https://github.com/YosysHQ/apicula/actions
But I'll just make a new release once this gets merged so that people can just pip install a working version.

infphyny:
> You could use the Dockerfile to generate the latest database. It the image is also available at https://hub.docker.com/r/pepijndevos/apicula
> Even easier if you're not planning to contribute to the project is to download the python package including the database from the latest github action https://github.com/YosysHQ/apicula/actions
> But I'll just make a new release once this gets merged so that people can just pip install a working version.

Thank you pepijndevos for the information.

olofk:
Cool. From the edalize perspective this looks all fine now. Looking forward to see it in action. It would be great if we could have support for some Gowin board add to [Project LED to Belive](https://github.com/fusesoc/blinky) that uses this backend

Would also appreciate some added tests for this backend, but I'm ok with adding it anyway so that people can try it out sooner. Thanks for your contribution!

pepijndevos:
Cool. I pushed a release with the part number support https://pypi.org/project/Apycula/

infphyny:
I will try to add test for gowin backend. I did not see test for trellis 
backend.  Maybe I can add test for trellis backend too based on 
test_icestorm.

PR #217, opened 2021-02-07
Run pnmainc instead of diamondc if using Windows.

cr1901:
`diamondc` does not exist on Windows. Contrary to what Lattice [says](http://www.latticesemi.com/support/answerdatabase/5/5/2/5522) on the subject, `pmainc` is properly set up on Windows to "just work". `diamondc` should still work fine on Linux.

### DIFF
--- a/edalize/diamond.py
+++ b/edalize/diamond.py
@@ -1,5 +1,6 @@
 import logging
 import os.path
+import sys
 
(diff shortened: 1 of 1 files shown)


olofk:
All good! Your domain expertise on Windows compatibility is most welcome. Picked and pushed

PR #222, opened 2021-03-01
Add support for Vivado jobs.

m-kru:


olofk:
Thanks. This overall idea is fine to me, but I would like to see two changes:
-  Use `Integer` instead of `String` for the argument type
- Only pass `-jobs` to `launch_runs` if the user has set a value. I couldn't find any info on what value Vivado uses by default, but I think we shouldn't force it to 1 unless the user has said so

m-kru:
> Only pass -jobs to launch_runs if the user has set a value. I couldn't find any info on what value Vivado uses by default, but I think we shouldn't force it to 1 unless the user has said so

The default value is 1, but I understand your point. In this way we are insensitive to the potential Vivado changes for defaults.

I wonder how this should be implemented. Should I pass an empty string `""` to the jinja template when the `jobs` is not set, or should I not pass the `jobs` argument at all in such case? Is there any similar example within the edalize?

olofk:
I was thinking not passing it at all. I guess this would be a a similar example where we only pass `-pvector bra` if arch is set to xilinx

imphil:
Can you add documentation when setting the jobs parameter is helpful with the setup we're having? Even with a single "job", Vivado uses multi-threading for all steps. I assume out of context syntheses are the primary case where going beyond one job is helpful?

m-kru:
@olofk take a look, I have addressed your comments.

olofk:
Perfect. Thanks for this!

PR #223, opened 2021-03-03
Fix block diagram generation in Vivado projects

morequri:
Vivado projects with block diagrams cannot be correctly generated by fusesoc with Vivado producing
`ERROR: [Vivado 12-3563] The Nested sub-design '...xci' can only be generated by its parent sub-design.`

The same problem has been also  mentioned on [Gitter](https://gitter.im/librecores/fusesoc?at=5f7c8c2b3d3d0d749a324999)

### DIFF
--- a/edalize/templates/vivado/vivado-project.tcl.j2
+++ b/edalize/templates/vivado/vivado-project.tcl.j2
@@ -44,6 +44,15 @@ set_property source_mgmt_mode {{ tool_options.source_mgmt_mode if tool_options.s
 {%- endif %}
 
 {% if has_xci -%}
-upgrade_ip [get_ips]
-generate_target all [get_ips]
+# Vivado treats IP integrator entities as nested sub-designs and prevents core
+# generation from the base project in non-GUI flow raising
+# ERROR: [Vivado 12-3563] The Nested sub-design '...xci' can only be generated
+#   by its parent sub-design.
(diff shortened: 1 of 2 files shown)


imphil:
Thank you for your contribution.

* Can you add a comment in the template as to why the this flag is used? These things add up and I'm sure in a couple months nobody remembers why we had this flag. Is there maybe even something in the Vivado user guide that documents this?
* Can you combine all commits into a single one using `git rebase -i`?

morequri:
> Can you add a comment in the template as to why the this flag is used?
> Can you combine all commits into a single one using git rebase -i?

Done

> Is there maybe even something in the Vivado user guide that documents this?

I wish it was, but IP core scoping seems to be too subtle a detail to be covered there.

I have also replaced `get_ips -exclude_bd_ips` with more generic `get_ips -filter {SCOPE !~ "*.bd"}`, since the -exclude_bd_ips flag has been introduced in Vivado 2019.1 and is not recognized by earlier versions. The new command runs correctly in earlier versions as well and does effectively the same thing.

imphil (approved):
Great, thanks! LGTM, @olofk can you have a quick look as well?

olofk:
Thanks! I trust your judgement on this. Just one of those obscure details that I would never have noticed myself due to a different workflow :)

PR #225, opened 2021-03-05
Diamond backend fixes

ThomasHornschuh:
Hi,
as I notice that there is a lot of merging going on to edalize/fusesoc I think it is a good idea also to add these changes to the diamond backend. I have some ideas to improve the backend further, but did not had time yet. So I decided to make a PR of my current minimal fixes:
When there are vhdlSource-2008 files in the project the 
````prj_strgy set_value -strategy Strategy1 syn_vhdl2008=True````
option is set, otherwise diamond will not compile vhdl-2008 files. 
VHDL Generics of type strings are put in quotes e.g.:
```prj_impl option HDL_PARAM {generic_bool=1;generic_int=42;generic_str="hello"}```
(I just noticed that generic_bool changed to 1 with the latest edalize master instead of True. No idea what is the correct value.  But in general generic passing to diamond seems to be broken, if in Diamond or edalize I can't say yet, so it may be something to fix with a future version. At least passing strings work with my fix)

I added a file of type "LPF" to edalize_common to cover this file type in the tests. As a (most likely unwanted) side effect the LPF file is also added to the tcl scripts for Vivado.  But the Vivado backend collects all sorts of files, so I think this is a problem of the Vivado backend itself. But it is a bit beyond the scope of this PR to fix this.


olofk:
Thanks @ThomasHornschuh . Looks good to me. We should probably investigate the generic passing but as you say, let's do that later. I actually tried to test this right now but it turns out my license expired 6 months ago and I have zero interest in dealing with that right now (I'm so f*in tired of FPGA vendors and their stupid licenses)

PR #226, opened 2021-03-08
Modelsim: Handle cSource/cppSource and header files with modelsim

tarikgraba:
Source files can be compiled by passing them to `vlog`.
Header files search path is also passed using the `-I` option.

TODO: All incdirs (RTL and DPI) are passed to `vlog` which could be too
much. This is done because DPI support in Fusesoc is not yet finalized.
See https://github.com/olofk/fusesoc/issues/311

### DIFF
--- a/edalize/modelsim.py
+++ b/edalize/modelsim.py
@@ -111,6 +111,17 @@ class Modelsim(Edatool):
                 if f.file_type.startswith("systemVerilogSource"):
                     args += ['-sv']
                 args += vlog_include_dirs
+            # XXX: C/CPP DPI sources can also be passed to vlog
+            # workaround until we have proper DPI support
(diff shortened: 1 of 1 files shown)


GCHQDeveloper560:
I recently experimented with DPI in ModelSim, so this would be nice to have! Thanks!

The build logs have expired to see exactly why the tests checks failed, but I suspect the ModelSim test is failing because you didn't update `tests/test_modelsim/edalize_build_rtl.tcl` to match the modified `vlog` command.

You can run the tests locally from your edalize directory with pytest or tox using `PYTHONPATH=. pytest` or just `tox` and if they pass there you should be alright on GitHub unless there are Python version or platform issues.

olofk:
I had no idea you could do this. Does this generally work good enough? Would be great to have some second opinion

PR #231, opened 2021-04-08
Support toplevel using dot notation in GHDL

### LINKED ISSUE #229: GHDL Backend Support for Toplevel in Library
VHDL designs that heavily use libraries will often specify the top level in FuseSoC with a dot notation, such as "my_lib.top_design". A number of tools including ModelSim and Vivado seem to support this common notation, but it's [not supported by GHDL](https://ghdl.readthedocs.io/en/latest/using/InvokingGHDL.html#elaboration-e):

> Name of the units must be a simple name, without any dot.

GHDL instead suggests using something like
```
ghdl -m --work=my_lib --workdir=./my_lib top_design
```

Given use of this format with other tools it feels like it would be worth what should be a fairly simple workaround in the GHDL backend. Is this of interest, or do I need to see about getting this supported in GHDL?

GCHQDeveloper560:
Fixes #229.

Happy to drop this if it interferes with the ongoing work from @m-kru


olofk:
Happy to add this, but it seems tests are failing for some reason. Perhaps the reference Makefile needs an update? 
Another thing that I noticed was this line https://github.com/olofk/edalize/pull/231/commits/0bedac7ab3b00ff1d73014163c77bf012f6075d6#diff-3f3a6ee59de71dc0c247eb1ecb26135e1acb1bc8ba926513231b7a93f11f4317R23 I'm not good enough with ghdl's ways of working, but does this mean that you're compiling all files belonging to the same lib in one go? On one hand I guess that's nice, but on the other hand, does this mean we're changing the compile order? Not sure we want to do that

GCHQDeveloper560:
The failing test was because I'd forgotten Edalize still supports Python 3.5, and I used a regular dict rather than ``OrderedDict``. I've been bitten by that before, so you would think I would have learned my lesson!

All of the files for a library are being imported at the same time, but I don't think we're introducing problems with compile order since GHDL figures it out in this flow. I certainly won't claim to be a GHDL expert, but my understanding [from the docs](https://ghdl.readthedocs.io/en/latest/using/InvokingGHDL.html#import-i) is that the import/make flow being used by this backend is specifically designed to avoid having to know file order. From the description of the make (``-m``) step used after the import:

> The purpose of this command is to be able to compile a design without prior knowledge of file order. In the VHDL model, some units must be analyzed before others (e.g. an entity before its architecture). It might be a nightmare to analyze a full design of several files if you don’t have the ordered list of files. This command computes an analysis order.

olofk:
Right. Sounds like ghdl is taking care of things then

Picked and pushed. Thanks for the contribution! I took the liberty to merge the two last patches

PR #234, opened 2021-04-15
GHDL requires strings for boolean generics, not integers

GCHQDeveloper560:
This was briefly mentioned in #131 but didn't make it into that commit.

A design with boolean generics will currently fail in elaboration with
an error like ``'value: '0' not in enumeration 'boolean'``. Edalize
currently uses its default of 1/0 for true/false, but GHDL expects the
true/false strings on the command line.


imphil (approved):
Thanks!

PR #235, opened 2021-04-15
Avoid warning from deprecated NumPy float alias

GCHQDeveloper560:
Tests using Python 3.7+ report the following warning:
```
  /home/runner/work/edalize/edalize/.tox/py/lib/python3.7/site-packages/edalize/quartus_reporting.py:112:
DeprecationWarning: `np.float` is a deprecated alias for the builtin
`float`. To silence this warning, use `float` by itself. Doing this will
not modify any behavior and is safe. If you specifically wanted the
numpy scalar type, use `np.float64` here.
  Deprecated in NumPy 1.20; for more details and guidance:
https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    freq.str.split(expand=True)[0].astype(np.float).to_dict()
```

### DIFF
--- a/edalize/quartus_reporting.py
+++ b/edalize/quartus_reporting.py
@@ -24,12 +24,6 @@ except ImportError as e:
     raise e
 
 try:
-    import numpy as np
-except ImportError as e:
-    logger.exception(import_msg, "numpy")
-    raise e
-
-try:
     import pandas as pd
 except ImportError as e:
     logger.exception(import_msg, "pandas")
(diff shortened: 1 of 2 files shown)


imphil:
`np.float()` is used a second time in this file further down. If you remove that all dependencies on numpy are gone and you can remove the numpy import in the beginning of the file as well as in https://github.com/olofk/edalize/blob/master/setup.py#L69 (please double-check with grep).

GCHQDeveloper560:
Nice catch! Glad you're paying more attention than I am!

imphil (approved):

imphil:
Thanks, that's a nice simplification!

PR #238, opened 2021-04-22
Nextpnr support for xilinx and fpga_interchange variants

kowalewskijan:
This PR supersedes the [`nextpnr-xilinx` support](https://github.com/olofk/edalize/pull/193) PR.

The previous PR is quite outdated and has some flaws regarding conda packages. Current version is more general and adds support for `xilinx` and `fpga_interchange` variants.


edalize/templates/yosys/yosys-script-tcl.j2
```diff
@@ -26,6 +26,11 @@ verilog_defaults -add {{ incdirs }}
 verilog_defaults -pop
 
 {{ synth_command }} {{ synth_options }} -top {{ top }}
+
+{% for command in additional_commands %}
+{{ command }}
+{% endfor %}
+
```
olofk, line 33: We already have the tclSource file type which allows us to inject commands. However, those commands will run before synthesis and it looks like you need to run these additional commands after synthesis. I wonder if we can come up with some clever way here. I have two ideas, but I'm not sure if either of them are very good.
1. Add support for defining custom attributes to files in the EDAM file. This would e.g. allow us to tag which tcl scripts to run before and after simulation
2. Define a post-synthesis hook in the main yosys script. This would allow users to register their own functions to be run after synthesis

The first alternative is more generic and I have been thinking along these lines for other use cases. However, it will require a fair amount of work in Edalize (and FuseSoC). The second one will make tcl scripts that use this hook Edalize-specific, but I'm still inclined to go this way.

Or perhaps it's best to just do it the way you propose. I'm just a bit careful about adding additional mechanisms for injecting commands

edalize/symbiflow.py
```diff
@@ -52,17 +54,17 @@ def get_doc(cls, api_ver):
...
                         "desc": "Additional vpr tool options. If not used, default options for the tool will be used",
                     },
                     {
-                        "name" : "environment_script",
-                        "type" : "String",
-                        "desc" : "Optional bash script that will be sourced before each build step."
+                        "name": "environment_script",
+                        "type": "String",
+                        "desc": "Optional bash script that will be sourced before each build step."
                     },
```
olofk, line 68: I would like to remove the enviromnent_script options and use the EDALIZE_LAUNCHER mechanism that I added for some other backends a while ago which is more generic

tests/test_icestorm/minimal/test_icestorm_0.tcl
```diff
@@ -17,6 +17,8 @@ verilog_defaults -add -defer
 verilog_defaults -pop
 
 synth_ice40  -top top_module
+
+
 write_blif test_icestorm_0.blif
```
olofk, line 22: Where do these empty lines come from? Could it be a jinja snippet that should use `-%` instead of `%` somewhere?

olofk:
Thanks for the updated PR. I haven't finished reviewing it yet but had a couple of comments on the things I read so far. One general question though. Is nextpnr-xilinx considered more production-ready now? IIRC it was very much work in progress last time we looked at it

mithro:
nextpnr-xilinx is still considered very experimental. 

**However**, if nextpnr-xilinx isn't supported by anything then it doesn't get any users, which means it doesn't become production ready. 

We are also using edalize in FPGA-Tool-Perf to understand how Vivado, VPR and nextpnr-xilinx compare to each other (and thus actually how "experimental" nextpnr-xilinx actually is).

mithro:
Oh, btw. We hope that in the future the separate nextpnr-xilinx, nextpnr-ice40 and nextpnr-ecp5 tools are all replaced by the nextpnr-interchange backend. That is going to take a while and we definitely want to support both so we can compare.

olofk:
Thanks @mithro . As was discussed last time I'm ok with supporting experimental tools, preferably marking them experimental. My worry comes if they need to add specific API to support this tool, because I want to avoid exposing things in the API that might quickly over time. But I think that should be alright here. Just need to tackle the issues one by one

edalize/templates/yosys/yosys-script-tcl.j2
olofk: You know what, I think the best long-term solution is to allow replacing the script template with a user-defined one. Yosys is becoming this swiss-army knife and I want to support as many reasonable use-cases as possible. So before we can merge this PR I want to do that work on yosys. I'll try to get it done ASAP.

olofk: I'm not sure I understand. Is it the paths to the json and xdc files you want to parameterize? If so I see three ways to do that. Either supply a TCL template file that reads some env vars, generate the template with correct paths before handing it to Edalize or make the TCL template call something externally that provides this information. Would any of those alternatives work. Not really sure how parameterization would work otherwise.

The exception is if you need some design-specific parameters but in that case I think we should make those available as variables, like I did with $top

kowalewskijan: Thanks for suggestions. I tested the approach with env vars and it works for me. I'll implement `yosys_template` solution to this PR.

kowalewskijan: I've reworked the code to support Yosys templates. I will rebase the branch when the `yosys_template` branch will be merged to master.

edalize/symbiflow.py
kowalewskijan: I've done some clean up and removed this option along with `SOURCE_OPTS` from makefiles.

tests/test_icestorm/minimal/test_icestorm_0.tcl
kowalewskijan: As we move to yosys templates this is outdated/removed

edalize/templates/yosys/yosys-script-tcl.j2
olofk: Great. Thanks for trying it out

edalize/nextpnr.py
```diff
@@ -0,0 +1,168 @@
...
+                    },
+                    {
+                        "name" : "nextpnr_as_subtool",
+                        "type" : "bool",
+                        "desc" : "Determines if nextpnr is run as a part of bigger toolchain, or as a standalone tool"
+                    },
+                ],
+                "lists" : [
+                    {
+                        "name" : "nextpnr_impl_options",
```
olofk, line 54: I just checked the other backends using nextpnr and the parameter is just called `nextpnr_options` in all of them

edalize/nextpnr.py
```diff
@@ -0,0 +1,168 @@
...
+                "members" : [
+                    {
+                        "name" : "arch",
+                        "type" : "String",
+                        "desc" : "Target architecture. Legal values are *xilinx*"
+                    },
+                    {
+                        "name" : "output_format",
+                        "type" : "String",
+                        "desc" : "Output file format. Legal values are *fasm*"
```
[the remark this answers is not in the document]
olofk: Ok, if I understand correctly, output_format does not really belong in nextpnr because a user can't change the output format. In my eyes, this should be hardcoded in the symbiflow part instead

edalize/nextpnr.py
```diff
@@ -0,0 +1,168 @@
...
+        xdc = None
+
+        for f in src_files:
+            if f.file_type in ["synthJson"]:
+                synth_design = f.name
+            elif f.file_type in ["bba"]:
+                chipdb = f.name
+            elif f.file_type in ["device"]:
+                device = f.name
+            elif f.file_type in ["xdc"]:
```
olofk, line 120: If we want to reuse the nextpnr class in apicula, trellis and icestorm (which I think we should do), then we need to be more flexible here since neither nextpnr_ice40, nextpnr_ecp5 or nextpnr_gowin has any knowledge of chipdb and xdc files. I'm starting to think that we perhaps should either begin with pulling out the nextpnr support from the existing backends and then add the symbiflow stuff on top of that. Or, put all this directly in the symbiflow class and extract nextpnr from all backends later

edalize/nextpnr.py
```diff
@@ -0,0 +1,168 @@
...
+            logger.error('Missing required "package" parameter for "fpga_interchange" arch')
+        package = package.split("-")[0] if arch == "fpga_interchange" else None
+
+        family = self.tool_options.get("family", None)
+        if family is None and arch is "fpga_interchange":
+            logger.error('Missing required "family" parameter for "fpga_interchange" arch')
+        if family not in getattr(self, "families") and arch is "fpga_interchange":
+            logger.error("Unsupported family: {}".format(family))
+
+        schema_dir = self.tool_options.get("schema_dir", None)
```
olofk, line 147: If I understand correctly, schema_dir is system-specific rather than design-specific. I'm trying to the largest extent possible to not require users to supply system-specific paths in EDAM as that tends to lead to hard-coded paths. Is there some way we can query the system for the path to these schemas instead?

olofk (changes requested):
Thanks for the updates. It's starting to look very good, but there are still some small and large changes to be done. The biggest question being if we should perhaps begin with making a common nextpnr class from the existing nextpnr users and then put the symbiflow additions on top of that

olofk:
Actually, I looked now at the nextpnr code in the existing backends and there is really nothing that can be share among them, so I think it's best to just move the nextpnr stuff into the symbiflow class for now

edalize/nextpnr.py
```diff
@@ -0,0 +1,168 @@
...
+                "description" : "Open source Place and Route tool targeting many different FPGAs",
+                "members": combined_members,
+                "lists": combined_lists
+            }
+
+    def configure_main(self):
+        (src_files, incdirs) = self._get_fileset_files()
+
+        yosys_synth_options = self.tool_options.get("yosys_synth_options", "")
+        yosys_template = self.tool_options.get("yosys_template", None)
```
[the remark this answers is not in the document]
kowalewskijan: fixed, thanks

edalize/nextpnr.py
```diff
@@ -0,0 +1,168 @@
...
+
+        file_table = []
+
+        output_format = self.tool_options.get("output_format", "fasm")
+        arch = self.tool_options.get("arch")
+
+        if not arch in getattr(self, "archs"):
+            logger.error('Missing or invalid "arch" parameter in "tool_options"')
+
+        synth_design = None
```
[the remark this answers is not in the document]
kowalewskijan: removed

edalize/nextpnr.py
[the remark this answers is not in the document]
kowalewskijan: I've removed this option -> `.fasm` is hardcoded in makefile templates

edalize/nextpnr.py
kowalewskijan: Actually I renamed this to `pnr_options` since this is a common part for both VPR and Nextpnr

edalize/nextpnr.py
kowalewskijan: I've merged Nextpnr into SymbiFlow class (`configure_nextpnr` method). I believe we can leave this as is?

edalize/nextpnr.py
kowalewskijan: I reworked this part to check if the `INTERCHANGE_SCHEMA_PATH` env variable is present -> if not it raises runtime error for `fpga_interchange` variant.

kowalewskijan:
I reworked the [symbiflow test](https://github.com/olofk/edalize/pull/238/files#diff-3f1ee8a2776ad358ed316d32689eb9a5465c7ee4247464c5e6a3d92a4c251888) (it was quite outdated) and additionally I added tests for `nextpnr-xilinx` and `fpga_interchange`.

edalize/symbiflow.py
```diff
@@ -52,31 +59,138 @@ def get_doc(cls, api_ver):
                     {
                         "name": "pnr",
                         "type": "String",
-                        "desc": 'Place and Route tool. Currently only "vpr" is supported',
+                        "desc": 'Place and Route tool. Currently only "vpr"/"vtr" and "nextpnr" are supported',
                     },
                     {
-                        "name": "vpr_options",
+                        "name": "pnr_options",
```
olofk, line 65: I think we should keep vpr_options and add a separate nextpnr_options instead for two reasons
1. It will break builds for users already using vpr_options
2. It's easier for users if they can specify correct options for both vpr and nextpnr and then just change the `pnr` option to switch between the different tools

edalize/symbiflow.py
```diff
@@ -52,31 +59,138 @@ def get_doc(cls, api_ver):
...
+                                    }
+                                }
+                }
+
+        yosys = getattr(import_module("edalize.yosys"), "Yosys")(yosys_edam, self.work_root)
+        yosys.configure()
+
+        # Nextpnr configuration
+        arch = self.tool_options.get("arch")
+        if arch not in getattr(self, "archs"):
```
olofk, line 109: Nitpick: Use `if arch not in self.archs:`

edalize/symbiflow.py
```diff
@@ -52,31 +59,138 @@ def get_doc(cls, api_ver):
...
+                break
+
+        if target_family is None and arch is "fpga_interchange":
+            logger.error("Couldn't find family for part: {}. Available families: {}".format(part, ", ".join(getattr(self, "fpga_interchange_families"))))
+
+        chipdb = None
+        device = None
+        constr = None
+
+        for f in src_files:
```
olofk, line 137: This will always pick the latest file. At least for xdc files it's not uncommon to have several files. If that's not supported by the backend we should at least warn users that only one of them will be used

olofk (changes requested):
We're getting very close now. Think these are my final comments. Awesome that you updated the tests as well. Highly appreciated

edalize/symbiflow.py
kowalewskijan: fixed

edalize/symbiflow.py
kowalewskijan: You're right - reverted to `vpr_options` / `nextpnr_options`

olofk:
Thanks for the update. Almost there. Code looks good but tests fail when I run locally because `INTERCHANGE_SCHEMA_PATH` is expanded in the makefile which ends up in a different path on my system. I would suggest using $(INTERCHANGE_SCHEMA_PATH) in the Makefile instead. This also allows portability if the Makefile is created on one machine and the tools are run on another. Perhaps we should even move the check to the makefile, like we do [here](https://github.com/olofk/edalize/blob/master/edalize/modelsim.py#L13). In that case we need to make sure the check is only added for fpga_interchange, I guess

kowalewskijan:
I've fixed the issue with the `schema_dir` - let's use the `INTERCHANGE_SCHEMA_PATH` directly in the makefile as you suggested.

olofk:
Believe it or not, we are done now :) Thank you for the contributions and your patience

PR #239, opened 2021-05-05
Symbiflow further enhancements

kowalewskijan:
This PR consists of following changes regarding the symbiflow toolchain:
* Add fasm2bels feature - this feature enables to generate e.g. `.dcp` files from symbiflow workflow output, which can be loaded by Vivado.
* Add default yosys scripts for Nextpnr workflows - this part covers utilizing the `yosys_template` which is set to default file provided in `edalize/templates` for certain workflow (if user-specified `yosys_template` is none).

Regarding the default yosys scripts:

The idea is to have default scripts for certain workflows which have customized TCL. In general these TCL scripts are always the same for Nextpnr workflows but still they differ from the very default `yosys-script-tcl.j2`. However I'm not sure if these scripts should be rendered or not - without rendering these scripts won't appear in the build directory as they're just loaded directly from given path - so tests for Nextpnr workflows would require to have TCLs removed along with `.mk` files.


kowalewskijan:
@olofk could you please take a look?

olofk:
I believe we can drop this now that we have the f4pga backend in place. Please reopen otherwise.

PR #240, opened 2021-05-08
Adding slang support to edalize

tianrui-wei:
[slang](https://github.com/MikePopoloski/slang) offers system verilog language services and is capable of many things. In particular, a complete Systemverilog preprocessor and linter.
This commit adds slang support to edalize


mithro:
+1

tianrui-wei:
Another thing I have in mind for this is that it adds some sort of export ip functionality, where you can specify some file sets and slang could output a single clean file instead with no includes attached. This makes it easy to integrate fusesoc into flows that requires unsupported tools.

olofk:
Thanks @tianrui-wei This is great. Just wanted to let you know that I'm a bit short on time right now and have other things I need to review first

tianrui-wei:
No problem @olofk :P Happy to wait until you can spare some time.

tianrui-wei:
Hi @olofk , I rebased my branch, did some cleanup and added some new features in the last few days, could you perhaps take another look?

tianrui-wei:
ping @olofk

tianrui-wei:
ping @olofk

Jbalkind:
@olofk could you take a look at this? We would really like it to land for some openpiton stuff we're doing

olofk:
This looks find. Thank you for the work and apologies for taking so long. I also want to mention that we are moving to a new API called the flow API and it would be great to have a tool class for the new API as well, but let's start with this

tianrui-wei:
Thanks Olof, let me know if you need me to submit another PR for the new API

olofk:
Yes. That would be great. I think slang support makes a lot more sense with the new API also, because then we can use it as a frontend. Unfortunately, the flow API is not yet enabled in FuseSoC, so if you are using FuseSoC I recommend we wait until FuseSoC 2.0 is released. I'm just releasing a first release candidate for it now so hopefully shouldn't be too long.

Issue #210, opened 2021-01-20
How to limit CPU count for parallel builds


sjalloq:
Hi there,

I randomly noticed that when running fusesoc for Verilator that it uses all available CPUs for a parallel build. This seems to be due to the following lines in verilator.py in edalize.

```python
        # Do parallel builds with <number of cpus> * 2 jobs.
        make_job_count = multiprocessing.cpu_count() * 2
        args = ['-j', str(make_job_count)]
```

Is there no way to limit the number of jobs?  I'm running on a shared server and can't consume all the cores.  I think this is a dangerous default to be using.

If I want to add an option to provide a user override, can you suggest the best way of doing this and I could have a go at providing a PR.

olofk:
Oh, I don't think we ever considered the case when someone is afraid of using too many cores. It was always the opposite. Yes, we need an override. 
 I wonder if it's bad enough so that we should default to a single job, or keep it as is for backwards compatibility.

I would propose adding a new `--jobs` parameter to the verilator backend. Here's an example of how to add a parameter https://github.com/olofk/edalize/commit/00d5dd158f56559450838bb440a92d5273ffb47a Once added you should be able to run `fusesoc run --target=yourtarget yourcore --jobs=<somenumber>` and it should appear there if it uses the verilator backend. Now there is some confusion here. I'm not sure if the -j parameter is used when creating the verilator model or when the verilated source are compiled to an executable. Would be good to figure out. Maybe we want the same number of jobs for both

shareefj:
@olof, I've just started running Verilator builds with FuseSoC again and have hit this issue.  I was trying to remember where I'd asked the question but a quick search found this.

Where did we get to?  I think the PR where you mentioned this ticket moved from using all threads to using all physical cores.  Still not much use for a job running in a cluster environment.  Can we add a new `tool_option` to limit this or perhaps use the environment variables set by various job schedulers?  For example, I'm currently using SGE and it sets the `NSLOTS` variable to the number of cores your job is assigned.  From memory the same is true when using LSF.

shareefj:
@imphil You commented on #247 so any thoughts on this solution?

imphil:
Maybe the simplest option is to not have edalize call make with any `-j` argument and document that the user can use the `MAKEFLAGS` environment variable to get parallel builds, e.g. `export MAKEFLAGS=-j8; fusesoc ...`.

shareefj:
Yep, that would work and is simple.  @olofk ?

[...]

PR #247, opened 2021-06-18
Don't overcommit CPU on Verilator builds

imphil:
With recent CPUs and hyper-threading, performing 2x as many builds as CPU
(virtual) cores are available is more likely to slow down the build process
due to frequent context switches than it is to actually speed things up.


olofk:
Good point. I was basing the number on some old (probably Gentoo) guide about optimum numbers but that is likely to be outdated by now. Defaulting to number of CPUs probably makes sense to do instead. However, we also have another related issue https://github.com/olofk/edalize/issues/210 that makes me wonder if we should default to 1 instead of trying to be clever.

PR #248, opened 2021-06-18
Fix Sphinx parse error in Vivado documentation

imphil:
The CAPI2 documentation and the edalize documentation are built
automatically from contents in the backends, and descriptions must
therefore be valid reStructuredText. Fix a case where this wasn't true.

Sphinx reported (in a FuseSoC build) due to the `*` not being followed
by another `*`:

```
capi2.rst:661:Inline emphasis start-string without end-string.
```


olofk:
I wonder if we still have issues somewhere. The rtd docs only show part and jobs as the available options for the vivado backend

imphil:
Edalize (as I now found out) doesn't even use this documentation, I filed https://github.com/olofk/edalize/issues/252 to dig deeper.

Issue #250, opened 2021-06-30
Command Generator Doesn't Support Phony Targets


GCHQDeveloper560:
The new command generator added in 7ef76b08b5e9b9e889f6da3c358aa35f5b64dea3 as part of #243 doesn't support phony targets. Looking at the Vivado rework as an example, almost all of its targets including `synth`, `syn`, `build-gui`, and `pgm` aren't real files that are being generated. The old Makefile template wasn't wasn't strict about this. It used to mark `pgm` as phony, though this was lost in 2fce13111de2e7d071d2ddd08598fae5d4fe6a68. It didn't mark the other phony targets.

This seems unlikely to cause a real problem since a user would need to be working in the build directory and create files with these names to break the build. The fact most of Vivado's phony targets weren't properly marked even before this change would seem to support this conclusion.

However, phony targets feel like a useful concept that I believe is supported by other command executers like Ninja, so it should be more broadly applicable than Makefiles.

GCHQDeveloper560:
As discussed in #251 the command generator should probably also support cases where multiple commands need to be run to generate a target.

I'm less sure whether variables should be supported, and if so whether or how make's multiple variables types should be supported. On one hand use of variables is a good practice, but on the other hand they're perhaps less important in machine-generated Makefiles unless they may be modified by hand later.

olofk:
Very good point about the phony targets. I want to avoid adding Makefile-specific things because, as you noted, I want to eventually support other flow executioners (first ninja, but also simpler like just as list of commands in bat/sh file and more advanced like SLURM and other cluster workload managers). Phony targets seem safe though and are a known thing at least in ninja.

Variables are more tricky. I have thought about it and it might end up inevitable to add support, but hopefully we can avoid it somehow. Since we are dealing with autogenerated makefiles I think it's a reasonable trade-off to expand variables inside Edalize compared to having the more elegant makefiles we used to have. Somewhat related to this is that we are long term aiming to support partial recompilation where only the part of the design actually changed will need to be recompiled. This will probably make a mess of the makefiles to the point where they could probably no longer be regarded as human readable

imphil:
> I want to avoid adding Makefile-specific things because, as you noted, I want to eventually support other flow executioners (first ninja, but also simpler like just as list of commands in bat/sh file and more advanced like SLURM and other cluster workload managers).

Do you have a doc describing the ideas?

olofk:
[...]

PR #251, opened 2021-06-30
Command Generator Enhancements

GCHQDeveloper560:
This is an initial attempt to support phony targets as discussed in #250. It seems likely that something like the `Target` class will be needed rather than just a string as is currently used, but it seems awkward to have `self.EdaCommands.Target` all over. I'm open to alternatives!


GCHQDeveloper560:
I took a first pass at converting Quartus to use the command generator. First, it has a case that uses multiple commands to generate a target, so I've added this to the command generator.

The Quartus Makefiles also uses variables. I thought about adding a `add_variable` method to the command generator, but while the Quartus case just uses simply expanded variables other backends use other types, making this less trivial. For now I've just appended variables to the header, which is likely poor form. I'm certainly open to other suggestions on how to handle this. The [`ninja_syntax` Python API for Ninja](https://github.com/ninja-build/ninja/blob/master/misc/ninja_syntax.py) just uses a variable method, but I also think it only has one type of variable!

GCHQDeveloper560:
The failing tests are unrelated to this work. See #254 for a reporting test that works with the new Pandas 1.3.

olofk:
Thanks for this. I agree that it looks a bit clumsy to have `self.EdaCommands.Target` everywhere. Could we perhaps add a flag instead to `add_command` to mark the rule as generating phony targets. That won't work if the rule creates a mix of phony and regular targets but perhaps that's not a real issue. Would need to dig into that, or perhaps you know?

Regarding variables, as far as I can tell they made a deliberate choice in ninja to downplay the use of variables so that they are just simple aliases. This makes me suspect that a) we need to be careful if we want to make this portable between command executioners and b) perhaps we can get around the need for variables in most (all?) situations. The ones I'm most worried about are the cases that rely on environment variables. Come to think of it, does that even work on Windows as it is implemented today? No idea

olofk:
Taking another look at this, long overdue. Wouldn't it be possible to solve this by simple creating an extra target, like we do here https://github.com/olofk/edalize/blob/master/edalize/tools/icetime.py#L60

That doesn't create a phony target per se, but we could either regard all targets without a real command as phony, or if that doesn't work, add a phony plag to the add function. Would that solve your issue?

Also, do we want to pull in the first panda commit as is?

PR #254, opened 2021-07-07
Fix failing tests with Pandas 1.3

GCHQDeveloper560:
The previous syntax for checking a column of identical values fails with
Pandas 1.3. The new check works with 1.3 and older versions.


olofk:
Thanks for this. Picked and pushed!

PR #255, opened 2021-07-10
quartus: Introduce globalassign parameter

trabucayre:
Sometime, it's required to add some `set_global_assignment`: to have access to SPI flash or to use some function pins as `REGULAR GPIO`.
This PR add `globalassign` in argtype list to complete the tcl with global assignment key/value

### DIFF
--- a/edalize/quartus.py
+++ b/edalize/quartus.py
@@ -2,6 +2,7 @@
 # Licensed under the 2-Clause BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-2-Clause
 
+from collections import OrderedDict
 import logging
 import os.path
 import os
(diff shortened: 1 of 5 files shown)


trabucayre:
Check failures are not related to this PR but with an issue about pandas.
These are fixed by #256

olofk:
Generally, I prefer to have as few special options as possible for the backends. This one would only be applicable for the quartus backend as I understand it. The common method of adding extra directives like this is to write them in a tcl file which has the file_type set to tclSource. The backend will then source the tcl file when it is creating the quartus project file. If you look at e.g. [here](https://github.com/fusesoc/blinky/blob/master/blinky.core#L61), you can see how we use this to set the pinout for a particular board.

So unless I am missing something, I don't want to merge this PR as it should be handled by tclSource instead

trabucayre:
Yes it's true it's possible to add corresponding lines in a tcl. And it make sense to limit/avoid options specific to keep most generic behaviour.

PR #256, opened 2021-07-14
tests: fix test_reporting for ise

trabucayre:
With panda 1.3.0 `pandas.core.series.Series.all()` return a `bool`
So, for ise's tests `assert df["Clock Signal"].all() == "clk_BUFGP"` and `assert ddr_io["IO Standard"].all() == "MOBILE_DDR"` fails.

This PR replace `all()` by using a `set` to reduce values to one element and compare this set with the expected value.

### DIFF
--- a/tests/test_reporting.py
+++ b/tests/test_reporting.py
@@ -293,7 +293,7 @@ def test_picorv32_ise_spartan6_resources(picorv32_s6_data):
     # Check Control Set Information table
     df = rpt["Control Set Information"]
 
-    assert df["Clock Signal"].all() == "clk_BUFGP"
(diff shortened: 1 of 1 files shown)


olofk:
Thanks for this. I just noticed this and #254 which I believe are addressing the same issue. @trabucayre @GCHQDeveloper560, any preferences for which PR to go for?

trabucayre:
Yes. Both address same issue. It's weird I'm sure to have done my "homework".
Maybe the fastest is the best. But seems logical to apply first open PR.

olofk:
I went for the other one as both solutions look very similar. Thanks for the PR still

PR #261, opened 2021-08-11
Add Surelog/UHDM support as yosys plugin

mtdudek:
Signed-off-by: Maciej Dudek <mdudek@antmicro.com>

### DIFF
--- a/edalize/icestorm.py
+++ b/edalize/icestorm.py
@@ -42,8 +42,10 @@ class Icestorm(Edatool):
             {'yosys' : {
                 'arch' : 'ice40',
                 'yosys_synth_options' : yosys_synth_options,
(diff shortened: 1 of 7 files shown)


kamilrakoczy:
Note: our custom version fork of yosys is still required for UHDM support, but now UHDM frontend is used as a separate plugin to yosys that is dynamically loaded.

saw235:
Hi @mtdudek ,
Can the sv2v flow be added to symbiyosys tool as well?

olofk:
Is this still needed with the new flow API?

kamilrakoczy:
I tried to use the new flow API using ``fusesoc``, but couldn't get it to work. Does the new flow API is already integrated with ``fusesoc``? If yes, are there any examples of correct ``core`` file?

PR #264, opened 2021-08-24
add support for prjoxide (nexus device)

trabucayre:
nexus devices are partially supported by nextpnr (nextpnr-nexus) and prjoxide. This is an alternative for radiant

This patch add a new class and update nextpnr class to support nexus devices.


olofk:
This looks great to me. Many thanks and sorry for the delay. I just completely missed it. Since this was done before the big reformatting I had to do a bunch of manual work to get it merged correctly. Could you please take a look to make sure I didn't mess anything up?

olofk:
Pretty sure this one is done and we just forgot to close it. Please reopen otherwise

trabucayre:
Yep. Applied since 7 month :)

PR #268, opened 2021-09-03
Libero - Add flexible TCL script hooks

carlosedp:
This allow adding TCL scripts on pre-synthesize, pre-pnr and
pre-bistream generation for the Libero backend.

In the .core file, this would be defined like:

```yaml
  polarfireeval:
    files:
      - rtl/corescore_polarfire_eval_clock_gen.v: { file_type: verilogSource }
      - rtl/corescore_polarfire.v: { file_type: verilogSource }
      - data/polarfire_eval.pdc: { file_type: PDC }
      - data/script_pre_synth.tcl: { file_type: tclSourcePresynth }
      - data/script_pre_pnr.tcl: { file_type: tclSourcePrepnr }
      - data/script_pre_bitstream.tcl: { file_type: tclSourcePrebistream }
```

Multiple files of each kind could be used.

Signed-off-by: Carlos de Paula <me@carlosedp.com>


carlosedp:
Ping @olofk ... News on this?
Cc. @green4free

carlosedp:
Rebased after the merge of #274. @olofk PTAL.
Thanks!

carlosedp:
Ping @olofk ... :)

olofk:
I'm not super excited about introducing these new file types but I understand the need to distinguish between these. My first question would be how Libero itself handles this. Is there some built-ink hook API we can use?

More long-term I see a need in other flows as well for specifying which tool in a toolchain that should consume particular files. To solve the general problem I think we need to introduced custom attributes to files in core description files. In this case it could look something like `some_tcl_file.tcl : {file_type : tclSource, stage : preSynth}`.

But if this PR solves an immediate need I could be convinced to merge it with the expectation that these file types might be deprecated in a not too distant future

BEforlin:
Hello @olofk @carlosedp ,

I see this has been silent for a while, and I wonder if there was any progress in the direction of this commit?

I was forced to implement something very similar (albeit not as polished) for my own build scenario in Libero. I can try to reply to some of @olofk's comments, but I am no expert.

> I'm not super excited about introducing these new file types but I understand the need to distinguish between these. My first question would be how Libero itself handles this. Is there some built-ink hook API we can use?

As far as I know, there isn't one. The build commands are discrete, even within Edalize’s backend, so I don’t see how they would provide hooks.

> More long-term I see a need in other flows as well for specifying which tool in a toolchain that should consume particular files. To solve the general problem I think we need to introduced custom attributes to files in core description files. In this case it could look something like `some_tcl_file.tcl : {file_type : tclSource, stage : preSynth}`.

I couldn’t find anything related to this in the documentation, so I assume it hasn’t moved forward. Is there still interest in pursuing this? From my perspective, breaking the backend file into smaller chunks would be very beneficial for Libero, given that the tool is essentially a series of TCL commands.

Let me know if I can contribute to an up to date version of this.

PR #269, opened 2021-09-06
Output Vivado synth log to stdout

GregAC:
Previously only the implementation job would output to stdout. The
synthesis job only output to a seperate log file. It seems Vivado only
outputs to stdout from jobs that are explicitly waited on with
'wait_on_run'.

### DIFF
--- a/edalize/templates/vivado/vivado-run.tcl.j2
+++ b/edalize/templates/vivado/vivado-run.tcl.j2
@@ -4,6 +4,11 @@ set_property STEPS.WRITE_BITSTREAM.ARGS.BIN_FILE true [get_runs impl_1]
 # Vivado will raise an error if impl_1 is launched when it is already done. So
 # check the progress first and only launch if its not complete.
 if { [get_property PROGRESS [get_runs impl_1]] != "100%"} {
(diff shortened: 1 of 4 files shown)


edalize/templates/vivado/vivado-run.tcl.j2
```diff
@@ -4,6 +4,8 @@ set_property STEPS.WRITE_BITSTREAM.ARGS.BIN_FILE true [get_runs impl_1]
 # Vivado will raise an error if impl_1 is launched when it is already done. So
 # check the progress first and only launch if its not complete.
 if { [get_property PROGRESS [get_runs impl_1]] != "100%"} {
+  launch_runs synth_1{{ jobs }}
```
imphil, line 7: Can you add a comment why you launch the jobs separately? This information gets lost easily when doing the next refactoring.

imphil (commented):
Thanks @GregAC! LGTM with one small comment.

imphil (approved):
LGTM, thanks a lot @GregAC.

Leaving this open for a day or two for @olofk to have a look, and will merge then.

Jbalkind:
Thanks for adding this - it was driving me nuts too :)

PR #270, opened 2021-09-09
Fix bug in RivieraPro script

dawidzim:
Hey!
In this pull request I fix bug which allows VHDL files compile under VLOG in common compilation. After this fix, SV and V files go to VLOG and VHDL go to VCOM


olofk:
Thanks. Picked and pushed

PR #271, opened 2021-09-15
Add VTR backend

jgoeders:
I'm working on adding a back-end to run the VTR tool (https://verilogtorouting.org/)


mithro:
FYI - @kgugala / @acomodi / @mtdudek / @kowalewskijan

jgoeders:
Since multiple small PRs is easier, I'll start with this PR which provides a basic backend, and will add more functionality through subsequent PRs.

olofk (changes requested):
Thank you for this contribution. I agree that it's best to take it in smaller steps and this is a great first step (bonus points for adding tests!). There are still things I want to change. Most notably the xml file handling

olofk:
@mithro My ambition is to support both use cases. People with virtual hardware or vtr devs will likely know where they have their architecture files. This is fine. We just need to provide an option for them to supply this file. For other users (like me) I'm hoping to avoid this for reasons cited above. They still need to know something about their target so that we can pick the correct description file, and to me it seems like the chip identifier is the most reasonable thing to use here. This however requires some kind of translation that is hidden to the users, which is what I attempted with https://gist.github.com/olofk/0f3911e7fa1028cc4c445bb2039ba06e . This approach might very well not work for other reasons but it would be good to figure out how to proceed

olofk:
Also, if there is a large mass of users who would supply these xml files anyway, then we can start by supporting that method and add the chipname-to-xml mapping later

mithro:
> @mithro My ambition is to support both use cases.

I agree!

> This however requires some kind of translation that is hidden to the users, which is what I attempted with https://gist.github.com/olofk/0f3911e7fa1028cc4c445bb2039ba06e . This approach might very well not work for other reasons but it would be good to figure out how to proceed

We could "extend" the `class Vtr(Edatool):` with a `class VtrWithSymbiFlowDeviceFiles(Vtr):` that finds and provide the SymbiFlow created device files. (Maybe with composition rather than inheritance?)

olofk:
So, a couple of follow-up questions for @jgoeders et.al. after digging through the vtr docs a bit.
1. Looking at the run_vtr_flow.py script I'm a bit confused about the arguments. It says `usage = "%(prog)s circuit_file architecture_file [options]"` but all the examples look like `%(prog)s arch.xml circuit.v -v 2`, i.e. architecture_file **before** circuit_file (assuming I got the nomenclature right).
2. Is `run_vtr_flow.py` the entrypoint we should use? We do have the option to let Edalize take over the responsibility of this wrapper script and call the tools directly if we prefer. Another route is to teach run_vtr_flow.py about EDAM so that we can feed it an EDAM file directly. The reason for bringing this up is that in my experience all these wrappers tend to limit the flexibility of the underlying tools and also tend to change a bit more frequently over time so the closer we can get to the real binaries, the less moving parts and more flexibility we have

@mithro I think the most straight-forward solution for the symbiflow device files is to implement that as a FuseSoC generator if we expect things to change in a not too distant future. Once everything is in place I can cook up an example

mithro:
> Yes, another approach would be to instead make edalize backends for the various tools tools (yosys, odin, abc, ace, vpr, etc) and then if the "stitching them together" mechanism isn't going to be a feature of edalize, then we could do that in run_vtr_flow.py and call into edalize for each stage. I would be fine with that as well.

I think this is the right long term strategy? @kgugala, What do you think?

EDALize should be responsible for controlling the various tools and potentially pulling together the tools into a flow. Eventually `run_vtr_flow.py` ends up being a thin wrapper around calling EDALize?

olofk:
@jgoeders Great to hear we're on the same page. I have actually done a bit of work on this over the past few weeks. You can find a PoC where I have converted the icestorm backend to this new architecture [here](https://github.com/olofk/edalize/tree/esr2). I need to clean it up a little bit before pushing it to main but I think you can get the idea what it should look like. Perhaps it could even get you started creating the tool wrappers for e.g. odin, abc and vpr. Having those would be helpful to determine whether this new architecture holds up, and we could also merge them one by one to keep the size of the PR smaller

olofk:
FYI @jgoeders I have pushed the initial version of the new flow API to the master branch so we should be in a good spot to add the vtr flow now

jgoeders:
OK, great, thanks for the update!  I'm pretty tied up with other projects at the moment, so likely won't be able to get to it for a while.  I have new students coming on at the end of April, and I hope to make it a summer project for one of them.

PR #273, opened 2021-09-22
[xcelium] updated xcelium flow to a working state

rasmus-madsen:
the previous state of the xcelium flow was not working correctly.
I have created a new flow based on the existing flow - adding a lot more freedom for the user.

before there was a onestop for all flow - where compile and simulation was always executed.
the new flow breaks this into the natural steps of compile/elaborate
and another step for the simulation which is very useful if anywant is running a regression sweep or just want to update a c library or similar and don't want to be forced to recompile the full RTL code.

I also created the option of adding a prefix to the run command.
which is useful when using submission queues.

let me know what you guys think.

### DIFF
--- a/edalize/xcelium.py
+++ b/edalize/xcelium.py
@@ -4,6 +4,7 @@
 
 import os
 import logging
+import sys
 
 from edalize.edatool import Edatool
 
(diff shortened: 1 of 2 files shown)


rasmus-madsen:
@imphil @olofk 
any feed back on this?

edalize/xcelium.py
```diff
@@ -76,55 +71,55 @@ def get_doc(cls, api_ver):
...
         tcl_build_rtl  = open(os.path.join(self.work_root, "edalize_build_rtl.f"), 'w')
+        tcl_build_c    = open(os.path.join(self.work_root, "edalize_build_c.f"), 'w')
 
         (src_files, incdirs) = self._get_fileset_files()
         vlog_include_dirs = ['+incdir+'+d.replace('\\','/') for d in incdirs]
 
         libs = []
+
+        print("\n DEFINES::", file=sys.stderr)
+        print(self.vlogdefine, file=sys.stderr)
```
olofk, line 95: Use the logger instead of printing to be inline with other backends

edalize/xcelium.py
```diff
@@ -76,55 +71,55 @@ def get_doc(cls, api_ver):
...
-                        {'name' : 'xmsim_options',
+                         'desc' : 'prefix for using submission servers like bjob or nc'},
+                        {'name' : 'run_options',
                          'type' : 'String',
-                         'desc' : 'Additional run options for xmsim'},
-                        {'name' : 'xrun_options',
-                         'type' : 'String',
-                         'desc' : 'Additional run options for xrun'},
+                         'desc' : 'Additional run options for xrun -R'},
                         ]}
```
olofk, line 83: The new options look cleaner, but I'm worried about current users of the xcelium backend because removing the old options will break things for them. Or do you mean that the xcelium backend didn't work at all before? In that case I guess there is less risk anyone will complain

edalize/xcelium.py
```diff
@@ -76,55 +71,55 @@ def get_doc(cls, api_ver):
...
+                        {'name' : 'elab_options',
                          'type' : 'String',
-                         'desc' : 'Additional options for compilation with xmvhdl'},
-                        {'name' : 'xmvlog_options',
+                         'desc' : 'Additional run options for xrun -elaboration'},
+                        {'name' : 'xrun_prefix',
                          'type' : 'String',
-                         'desc' : 'Additional options for compilation with xmvlog'},
-                        {'name' : 'xmsim_options',
+                         'desc' : 'prefix for using submission servers like bjob or nc'},
```
olofk, line 79: Drop this and use the EDALIZE_LAUNCHER environment variable instead. There is an upcoming refactoring that, among other things, will standardize this for all the backend flows

olofk (commented):
This looks mostly fine except for a few smaller things. With that said however, I'm closing in on a larger refactoring (as described somewhat in https://github.com/olofk/edalize/wiki/Edalize-(Slight-return) and which you can find a PoC in https://github.com/olofk/edalize/tree/esr2. This means that depending on how fast I will land this refactoring, I might need a larger redesign of this code.

edalize/xcelium.py
rasmus-madsen: unless someone would use the generated make as standalone without the generated file list - the previous backend did not work.
so I think it is safe to assume that no one used it.

edalize/xcelium.py
rasmus-madsen: I will look into it

edalize/xcelium.py
rasmus-madsen: thank I will do that.

PR #274, opened 2021-09-28
Add floor plan pdc to libero target

green4free:
I added a src file filter to the libero target for floor planning PDC files. Libero requires these commands to be in different PDC files from IO constraints.
I'm not sure if the file_type FPPDC follows naming convention, so some input on that would be nice.

To make synthesis of a SmartFusion2 project created and built using FuseSoC work, I needed to change a path in `edalize/templates/libero/libero-project.tcl.j2`, so that commit is also included here, but that should maybe be an issue instead.
From what is this path supposed to be relative to? @carlosedp
My current thinking is that it might be a Libero version problem, I'm using Libero v2021.1


edalize/templates/libero/libero-project.tcl.j2
```diff
@@ -38,7 +38,7 @@ puts "---------- Executing User TCL script: {{src_file|tcl_file_filter|replace("
 
 # Configure Synthesize tool to use the generated Synplify TCL script
 configure_tool -name {SYNTHESIZE} \
-        -params {SYNPLIFY_TCL_FILE:../../{{name}}-syn-user.tcl}
+        -params {SYNPLIFY_TCL_FILE:../{{name}}-syn-user.tcl}
```
carlosedp, line 41: Usually this path is used with two ".." because the `tcl` file is in the base and Libero reads from `base/prj/synthesis` so the tcl file is two levels below.

carlosedp (commented):
In addition to adding the file to the filters, you also need to update the templates, adding where the files should go and also to the test case.

In the test, you add a file (actually just a name) in `edalize_common.py` section `FILES = [`.

Then add to the libero test case and is generated templates if they match-up.

edalize/templates/libero/libero-project.tcl.j2
carlosedp: I really don't know why it read from one folder below only... maybe a version difference but I doubt they changed this from version to version.

green4free:
Ok. So do I understand it correctly that I add the file name to the FILES list in edalize_common, then run pytest(which will fail?), then look at the generated files and if they look correct, update the reference files in test_libero, and then run pytest again which should now pass?

edalize/templates/libero/libero-project.tcl.j2
green4free: I can create small simple test core and compare between libero v2021.1 and v12.5.

imphil:
@green4free If you'd like to update all reference files in one go (and then manually check them), you can also use `GOLDEN_RUN=1 pytest`.

edalize/templates/libero/libero-project.tcl.j2
green4free: I have tested it against both v2021.1 and v12.5 now, and there is a difference between them. v12.5 references relative paths to the SYNPLIFY_TCL_FILE against {project root}/synthesis, whereas v2021.1 references it against the {project root}.

Should I add a version setting to the libero target? Which version should that then default to? v12.5 to keep current behavior, or the latest version?

carlosedp: Damn, why they change these things. Is it possible to identify which version is running or any internal variable holding the root project path. Maybe instead of referencing thru relative paths (../../blabla), use a ({project_root_or_something}). Maybe both versions have the same variable name.

If not possible, I think we could keep the previous version as default since they might be the most used one.

carlosedp:
> Ok. So do I understand it correctly that I add the file name to the FILES list in edalize_common, then run pytest(which will fail?), then look at the generated files and if they look correct, update the reference files in test_libero, and then run pytest again which should now pass?

Take a look at my last PR where I add the new files. Check the files and sections I changed... it's mosly adding the file to edalize_common, the filters to libero.py, the .j2 templates specifying where they should go and running the tests with GOLDEN_RUN as @imphil mentioned to generate the test output (and checking if they are correct).

Ref. https://github.com/olofk/edalize/pull/268/files

carlosedp:
This looks great!
@olofk @imphil would you guys take a look at it?

green4free:
Now I have fixed the test for floor plan PDC files. And added a tool option for if you are using a version of Libero SoC v2021 or later.
Looking at the yosys target it seemed that you could set the type of an option to 'bool', but that gave me a Parse Error:
image
So I implemented it as a String.

carlosedp:
I don't know if this might work but there is an option on project settings that defines the root directory for the project.

image

I couldn't find any examples but got this from the Microsemi "Tcl Commands Reference Guide Libero SoC v12.5".

It would avoid having the version flag... worth a try @green4free ?

green4free:
I can try that.

green4free:
@carlosedp, I was not able to make that to work. The `project_general_settings` command is not included in the v2021.1 version of the TCL guide.
http://coredocs.s3.amazonaws.com/Libero/2021_1/Tool/libero_soc_tcl_cmd_ref_ug.pdf
And libero segfaults when I try to use it.

Another solution might be to include the path as an absolute path instead of a relative one, both versions of the TCL guide specifies that the SYNPLIFY_TCL_FILE path can be provided absolute.

carlosedp:
Ah ok... I was just trying to find something to make the implementation easier.

LGTM.

carlosedp:
Great!
Can you review this @olofk?
Thanks!

carlosedp:
Ping @olofk  😁

edalize/templates/libero/libero-project.tcl.j2
olofk: Welcome to the EDA world. These people are insanely inconsistent so it's best to rely as little as possible on their conventions unfortunately :/ As I'm not a Libero user myself I can't offer much help here unfortunately but I'm happy to support whatever you come up with.

olofk:
This looks fine to me. About the naming convention, I'm not super excited about adding a specific file type for this. The long-term plan is to introduce attributes to files, so that you can e.g. have a pdc file with the `floorPlan` attribute set. This will also help with the issue where multiple tools in a toolchain all want to read separate tcl scripts. But! This is still in the future and I don't think I will realistically have it implemented anytime soon, so let's go with FPPDC for now, with a disclaimer that I might deprecate this at some point in the future

PR #276, opened 2021-10-06
Test the execution of more than one hook

imphil:
Add a test to ensure that two hooks are successfully executed. Also fix another related test and improve the code style.


olofk:
Note to self to never write tests that are supposed to fail without checking why they fail :) Thanks for finding out and fixing. Slightly  dubious to use /usr/bin/touch since that shouldn't work for people running tests on windows, I guess. But let's go for it anyway

PR #279, opened 2021-10-29
Update CI

umarcor:
This PR brings several updates to CI:

- It allows running the workflow for any branch or PR.
- Fail-fast is set to false (defaults to true).
- The style is updated and symbols/icons are added.
- Python 3.10 is added.
- CRON and workflow_dispatch events are added/enabled.

macos and windows jobs are temporarily disabled because Edalize seems to fail on those platforms. The workflow in branch `master` is currently green because the system is always Ubuntu (the matrix is ignored): see https://github.com/olofk/edalize/blob/master/.github/workflows/ci.yml#L12.


.github/workflows/ci.yml
```diff
@@ -2,29 +2,43 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
   pull_request:
-    branches: [ master ]
+  schedule:
+    - cron: '0 0 * * 5'
+  workflow_dispatch:
```
imphil, line 8: Can you explain the motivation for these changes?

pyproject.toml
```diff
@@ -1,7 +1,10 @@
+[tool.tox]
```
imphil, line 1: Can you explain why this change is beneficial? Embedding an ini file as string in another ini file doesn't look like an improvement to me.

imphil:
Thanks for your pull request @umarcor! I'd like to understand the motivation for some of the changes better, can you have a look at the comments? Thanks!

.github/workflows/ci.yml
umarcor: The scheduled/CRON event ensures that CI is executed once a week, despite the codebase not being updated. CI did not run in the last 11 days, since the last commit was pushed until I opened a PR. Since edalize has dependencies, it is desirable to run CI periodically in order to catch regressions/breaking changes.

The workflow_dispatch event allows triggering the workflow manually through the web GUI, or through the GitHub API.

pyproject.toml
umarcor: `pyproject.toml` is a configuration file used by multiple projects, instead of having one configuration file for each. Particularly, apart from `tox`, I'd like to propose a PR for using `black`. Both `tox` and `black` support `pyproject.toml`. See, for instance, VUnit (https://github.com/VUnit/vunit/blob/master/pyproject.toml). See also https://github.com/carlosperate/awesome-pyproject.

imphil: I would agree if tox would "natively" read pyproject.toml. But embedding ini within an ini is a step backwards IMO. I'd ask you to revert that. Having a different configuration file for a different tool isn't a problem in any way.

.github/workflows/ci.yml
imphil: Can you add a comment to that effect to the file?

umarcor: Done.

pyproject.toml
umarcor: Done.

.github/workflows/ci.yml
```diff
@@ -2,29 +2,49 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
   pull_request:
-    branches: [ master ]
+  schedule:
```
imphil, line 6: Let's make those comments a bit more helpful and explain *why* the constructs are used.

```suggestion
  # Run CI daily even without code changes to ensure tests pass with 
  # updated dependencies.
  schedule:
```

.github/workflows/ci.yml
```diff
@@ -2,29 +2,49 @@ name: CI
...
+    - cron: '0 0 * * 5'
+  # See:
+  # * https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#scheduled-events
+  # * https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07
+  workflow_dispatch:
```
imphil, line 11:
```suggestion
  # Allow triggering a CI run from the web UI.
  workflow_dispatch:
```

imphil (commented):
We're almost there with two comments to go.

And you can fold the "[CI] Add refs to documentation about 'schedule' and 'workflow_dispatc" commit into the "[CI] Add CRON and workflow_dispatch events" commit.

umarcor:
Just squashed and force-pushed. I successful, I think it is ready to merge.

imphil (approved):
Thanks!

PR #280, opened 2021-10-30
Update docstrings

### LINKED ISSUE #278: Documentation format not following Python standards (thus not appearing in Sphinx).
Current form of documentation (doc is before the method)
```python
    """ Get tool version
    This gets the Vivado version by running vivado -version and
    parsing the output. If this command fails, "unknown" is returned
    """
    def get_version(self):
        ...
```

Correct Python "docstring" format (doc is the first triple quoted string side the function);
```python
    def get_version(self):
        """ Get tool version
        This gets the Vivado version by running vivado -version and
        parsing the output. If this command fails, "unknown" is returned
        """
        ...
```

See https://www.python.org/dev/peps/pep-0257/

This means that these docstrings are not automatically included in the Sphinx documentation at https://edalize.readthedocs.io/en/latest/source/edalize.html#module-edalize.vivado

umarcor:
Close #278.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -110,7 +110,7 @@ except ImportError:
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+# html_static_path = ['_static']
 
 # Custom sidebar templates, must be a dictionary that maps document names
 # to template names.
--- a/edalize/edatool.py
+++ b/edalize/edatool.py
@@ -70,7 +70,8 @@ else:
 
 # Jinja2 tests and filters, available in all templates
 def jinja_filter_param_value_str(value, str_quote_style="", bool_is_str=False):
-    """ Convert a parameter value to string suitable to be passed to an EDA tool
+    """
+    Convert a parameter value to string suitable to be passed to an EDA tool.
 
     Rules:
 
@@ -353,7 +354,7 @@ class Edatool(object):
 
     def render_template(self, template_file, target_file, template_vars = {}):
         """
-        Render a Jinja2 template for the backend
+        Render a Jinja2 template for the backend.
 
         The template file is expected in the directory templates/BACKEND_NAME.
         """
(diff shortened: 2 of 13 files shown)


imphil:
About the first commit, "[Doc] Update Sphinx configuration file": Why are you removing the comments? They are actually helpful, I find.

Generally, please include a motivation *why* you're making changes in the PR description and the commit message: it makes it much easier to review. (The *what* is typically visible in the code itself, so that needs less of a description.)

umarcor:
> Why are you removing the comments? They are actually helpful, I find.

The configuration file was auto-generated long ago, and it included the description of the options back then. Since then, the available options and their functionality might have changed. Therefore, I believe it is better to point to the updated documentation (http://www.sphinx-doc.org/en/master/config), rather than trying to manually keep the comments in the `doc.conf` file up to date.

doc/conf.py
```diff
@@ -182,23 +97,11 @@
...
-#
-# epub_uid = ''
-
-# A list of files that should not be packed into the epub file.
 epub_exclude_files = ['search.html']
 
 
 # -- Extension configuration -------------------------------------------------
 
-# Take class documentation from both __init__(), and the class docstring.
```
imphil, line 203: This comment is intentionally added, let's keep it.

imphil (commented):
> The configuration file was auto-generated long ago, and it included the description of the options back then. Since then, the available options and their functionality might have changed.

Readability of such a config file matters a lot -- having to look up every option is a step backwards (it's not like most people write Sphinx configs every day and remember the options).

I'm fine with removing the commented out options, but I'd ask you to keep at least a short description of the less options where the behavior isn't obvious. 

To your point of keeping the description in sync with upstream Sphinx: if the meaning of an option changed radically, it would most likely be renamed anyways -- you cannot keep backwards compat by keeping a name and changing the meaning.

Besides that, the other changes are fine. Thanks!

umarcor:
> Readability of such a config file matters a lot -- having to look up every option is a step backwards (it's not like most people write Sphinx configs every day and remember the options).

I'd say that whoever edits the config file should open the updated documentation, precisely because we don't use that file frequently. Anyway, since those comments are not the purpose of this PR, I removed that change. I just commented the `html_static_path` which produces a warning at build time.

imphil (approved):
LGTM, thanks!

PR #282, opened 2021-10-30
[GHDL] Dot notation for specifying my_lib.top_unit is supported

umarcor:
See ghdl/ghdl#1718, which was fixed a day after merging #231.

### DIFF
--- a/edalize/ghdl.py
+++ b/edalize/ghdl.py
@@ -89,9 +89,10 @@ class Ghdl(Edatool):
         ghdlimport = ""
         vhdl_sources = ""
 
-        # GHDL doesn't support the dot notation used by other tools (e.g.
-        # my_lib.top_design) for the top level so work around this if the user
-        # has specified the top level in this manner.
+        # GHDL versions older than 849a25e0 don't support the dot notation (e.g.
+        # my_lib.top_design) for the top level.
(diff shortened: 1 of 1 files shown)


imphil:
This PR removes the comment, but keeps the workaround added in #231 in place. Shouldn't we either keep the comment (in a modified form) for older versions of GHDL, or backout the whole change and only support newer GHDL versions?

umarcor:
@imphil, looking at the codebase, it was not clear to me whether it is desirable to revert the workaround. Splitting the library and the primary unit is the recommended solution, and keeping it allows not having to detect the version of GHDL (it works with both older and newer versions). The dot notation feature is interesting when users have to manually write the CLI commands, because it saves time. However, when CLI commands are autogenerated (such as done by edalize), I don't think the dot notation adds any value.

imphil:
I agree that it makes sense to keep this workaround for the time being. In that case, please don't remove the whole comment (since it explains what the code does). Instead, you can change it to something like "GHDL versions before XZY don't support ...".

umarcor:
I rewrote the comment, instead of removing it.

imphil (approved):

imphil:
Thanks, LGTM!

Issue #25, opened 2019-05-06
Add backend to write *.f files


imphil:
Many EDA tools support digesting file lists in *.f files. Having a (pseudo-)backend which generates these files would serve as stopgap for EDA tools which we don't support natively so far, and opens the door to more exploration.

Feature list (to be discussed):
- Optionally add defines, parameters and libraries to the file list (some tools support this, other don't)
- Optionally output absolute paths

Design considerations:
- Can we enable fusesoc to use this backend without needing to add this new backend to each core file?

sjalloq:
@imphil @olofk,  I was just about to ask some questions related to this as I'm looking at doing this for our project.  Can we resurrect this thread.

So far I've got a 'filelist' backend that generates a VC file (.f/.vc) from the EDAM file list.  It seems easy to generate a list of absolute file paths and include directories and write them out to the work directory.  But what I wanted to do was allow the user to configure where the file should be written, for example, output the .f file parallel to the parent .core file.  However, it doesn't look like the information for this is passed to Edalize.  Is that correct?  Or can I get the path to the parent .core file somehow?

@olofk , what was your point about conditionals?  I don't understand the issue here as I've not got a huge amount of experience looking at the internals.

Shareef.

sjalloq:
Here's my first pass which will hopefully stimulate some discussion:  https://github.com/sjalloq/edalize/commit/08c07fd42df7a787c13196a900a6b52d7d9d1a85

It requires an update to FuseSoC too:  https://github.com/sjalloq/fusesoc/commit/4db03432acddfcc4a0c4c605782f62afdf852f38

sjalloq:
@olofk , is it possible to force `--no-export` from within a core file?

yrrapt:
I too would be interested in this feature.
@sjalloq created a modified version we used on a previous project together which can be found [here]( https://github.com/SystematIC-Design/fusesoc).
But having the feature as part of the upstream tool would be preferred.

olofk:
Sorry about the lack of feedback here. I would like to think a bit differently about the problem actually. The main use case here seems to be to run tool flows that accept .f files but don't have an Edalize backend.

What I would like to do is to split this into two problems.
1. An EDAM to .f converter
2. A simple way to create custom flows

Looking at the first problem, I don't think this would necessarily have to be an Edalize backend. It can be a small utility that can be called on the command line or instantiated as a python class. This would allow stand-alone usage.

[...]

PR #289, opened 2021-12-10
write_filelist: adding new filelist backend.

sjalloq:
This is a first pass at the filelist backend.  Still looking for feedback from the original discussion but thought a PR might kick start that.  No test updates yet.

See https://github.com/olofk/edalize/issues/25 for original discussion.

### DIFF
--- a/edalize/filelist.py
+++ b/edalize/filelist.py
@@ -1,0 +1,175 @@
+# Copyright edalize contributors
+# Licensed under the 2-Clause BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-2-Clause
+
+import os,sys,re
+import logging
+
+from edalize.edatool import Edatool
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+class Filelist(Edatool):
+
+    # Currently only +define+ is supported in the VC file. Some tools allow parameters to be passed but the syntax is tool specific.
+    argtypes = ['vlogdefine']
+
+    @classmethod
+    def get_doc(cls, api_ver):
+        if api_ver == 0:
+            return {
+                'description': 'The filelist backend writes out a VC file to be used in tool flows not yet supported by Edalize',
+                'members': [
+                    {
+                        'name': 'outpath',
+                        'type': 'String',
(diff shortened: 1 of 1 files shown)


olofk:
As you know, I had another preferred way of doing this, but I have also come to realize that it's much better to get this in than waiting for another fictive solution that doesn't exist yet. I would appreciate adding tests though, but this looks fine and I understand you are using this yourself successfully, so let's pull it in as is.

Thank you for your contributions.

shareefj:
OK, thanks.  I can take a look at adding tests and if anyone else has any feedback on its usage then I can rework this.  This has diverged somewhat from our own branch so I'll double check it still all hangs together correctly.

On the topic of your preferred way, happy to help out in any spare cycles I get.  I remember you'd replied to a comment somewhere on how you saw it working, but if you put that in a branch and/or a new ticket then perhaps we can work on it in the background.

shareefj:
Ah yes, it's not going to work as is.  @olofk - you didn't want to push the other half of this request which was to add `top_core` to the Fusesoc EDAM arguments.  This works for us as we're still using our own branch: https://github.com/idex-biometrics/fusesoc/commit/05695284cf5e26128dcfcb53951eef37b670190f

davidp135:
Great, thanks Il give it a try on my own branch. @olofk is this something that will go into the main branch of Fusesoc at some point?

shareefj:
@davidp135  Just for some background.  Olof had a reservation that adding a new key would somehow break backwards compatibility but I couldn't see it at the time and the conversation went cold.  So this proposal was shelved and we were going to look at a script that could post process the EDAM YAML or something.  I forget the details.  There's a ticket with a short discussion somewhere.

I think this PR was merged following me showing someone else an example but we both forgot that there was a corresponding Fusesoc PR that he didn't want to put in.

Given the fix to Fusesoc is a single line you should be able to patch that on top of latest.  I might even merge olofk/fusesoc/master into my branch at some point.  But we probably need to revisit this sooner if people think it might be useful.

davidp135:
Great thanks @shareefj  I have applied the patch and am now getting successful flilelist generation.

PR #290, opened 2021-12-10
feat_board_file_support: adding support for BD files

sjalloq:
- added support for the "bd" file_type for Vivado Block Designs. In the same
   way as there is XCI support, users can now provide a list of .bd files.
 - added two tool_options entries for board_part and board_repo_paths. The first
   allows a board_part property to be set in the project and the latter allows
   a local path outside of the tool installation to be used for board files.

### DIFF
--- a/edalize/templates/vivado/vivado-project.tcl.j2
+++ b/edalize/templates/vivado/vivado-project.tcl.j2
@@ -1,9 +1,22 @@
 # Auto-generated project tcl file
 
+{% if tool_options.board_repo_paths -%}
+# Must be set before creating the project otherwise it is ignored.
+# Allows the user to provide local path to board files.
+set_param board.repoPaths {
+  {%- for path in tool_options.board_repo_paths %}
+  {{ path }}
+  {%- endfor %} }
+{%- endif %}
+
 create_project {{ name }} -force
 
 {% if tool_options.part -%}
 set_property part {{ tool_options.part }} [current_project]
+{%- endif %}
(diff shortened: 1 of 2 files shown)


sjalloq:
This pull request requires https://github.com/olofk/fusesoc/pull/533

sjalloq:
This is a work in progress but I wanted to post it to get any early feedback.

olofk:
Great! Having better support for bd files is an often requested feature. I hope we can find a way to not having to copy board files directories though. I'm always a bit sceptical towards having paths like that in the core description files, because they are system-specific rather than core-specific. I haven't used the board directory feature much but I have a vague memory that you can set the board paths directly in some vivado configuration file. Would that work instead so that we can keep configuration of board dirs outside of fusesoc?

olofk:
I think it's using `set_param board.repoPaths [list "some/different" "board/paths"]` in `$HOME/.Xilinx/Vivado/Vivado_init.tcl` I'm thinking of. Is that an alternative?

sjalloq:
The ability to copy the board files isn't a required step, it's optional.  You can just put them in your $HOME, however, I'm running in a cluster environment where multiple users will be using the flow.  I need to version control everything and have the flow control where files are stored.

olofk:
Let's go with this with then. Still hope we can come up with a way to avoid storing system-specific paths in core description files, but for now it solves a problem. Thanks for your contributions. Picked and pushed

PR #294, opened 2022-01-08
Format tests

umarcor:
This PR fixes the failing 'Format' job in CI.

### DIFF
--- a/tests/mock_commands/nextpnr-nexus
+++ b/tests/mock_commands/nextpnr-nexus
@@ -2,10 +2,10 @@
 import os
 import sys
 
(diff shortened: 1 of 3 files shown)


imphil:
I don't think we can get away with Python 3.6 being removed, and certainly shouldn't do that in a PR that also combines other things. Python 3.6 is still standard in Ubuntu 18.04, which will stay with us for a while longer in the EDA world.

umarcor:
I updated this PR to only include format fixes.

imphil (approved):
Thanks, much appreciated!

PR #295, opened 2022-01-09
Deprecate and remove Python 3.6 (EOL since 23 Dec 2021)

umarcor:
Close #294.
Coming from [#294](https://github.com/olofk/edalize/pull/294#issuecomment-1008122511).

This PR bumps the required Python version to `>=3.7`, since Python 3.6 was EOL in december 2021. Furthermore, it seems that Edalize does require Python 3.7 already. According to CI tests, Python 3.6 is failing since a month ago, when Icestorm backend was converted to the new API: https://github.com/olofk/edalize/commit/78584db87769a87dacac92d11f0f2dd2ac0723f8.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,6 @@ jobs:
         #- { icon: 🍎, name: macos }
         #- { icon: 🧊, name: windows }
         pyver:
-        - '3.6'
         - '3.7'
         - '3.8'
         - '3.9'
--- a/setup.py
+++ b/setup.py
@@ -83,6 +83,6 @@ setup(
         "reporting": ["pyparsing", "pandas"],
     },
     # Supported Python versions: 3.6+
-    python_requires=">=3.6, <4",
+    python_requires=">=3.7, <4",
     scripts=["scripts/el_docker"],
 )


imphil:
I'm against removing Python 3.6 support at this point, and instead vote for fixing the breakage (https://github.com/olofk/edalize/commit/78584db87769a87dacac92d11f0f2dd2ac0723f8 shouldn't have been committed without PR/CI passing). 

Python 3.6 is still used as system Python in Ubuntu 18.04, SLES 15, and RHEL 8, which also provide security updates to it -- this makes the upstream EOL date irrelevant for those distributions. edalize is used in the EDA world, where rather old setups are very common.

If we have a strong reason to revisit we certainly should do so, but I don't see maintainability of edalize impacted at the moment by also supporting Python 3.6.

What are your thoughts @olofk?

olofk:
If it had been up to me we would probably still have py2.7 support :) One of the build machines I use at work still run Ubuntu 14.04 and last time I worked for a larger client they had plenty of setups older than that.

At some point we need to draw the line though but I would argue that RHEL8, SLES15 and Ubuntu 18.04 are systems we should readily support.

I messed up the 3.6 thing but it's probably some easy syntax thing that can (and should) be fixed. Good point that it happened because I didn't open a PR first. Will think of that in the future

imphil:
I'm now looking into the Python 3.6 breakage. Looks slightly odd at first sight.

PR #296, opened 2022-01-10
Add a __init__.py file to the tools subpackage

imphil:
Subpackages need to have an __init__.py file. Add one to the `tools`
subpackage.

Without the `__init__.py` file the `ModuleSpec` has `loader = None`
set on Python 3.6, but not on Python 3.7. I couldn't find
documentation on this change, but it certainly breaks Jinja2's
`PackageLoader()`.


olofk:
Thanks. How come that there aren't any complaints about the flow subpackage? Is that because we don't have any jinja things looking there?

imphil:
> How come that there aren't any complaints about the flow subpackage? Is that because we don't have any jinja things looking there?

Most likely. It's not really Jinja failing here, it's the Python import machinery behaving differently (which Jinja's `PackageLoader()` utilizes). I haven't really been able to find documentation on something that changed between 3.6 and 3.7 and would explain it, but then I only spend an hour or so and wasn't interested digging even further :-)

olofk:
Fair enough. Let's add one to the flows subpackage when we get the chance to do it as well

PR #298, opened 2022-01-17
Add Symbiflow Quicklogic targets. Pass vendor and part as env var.

carlosedp:
This PR adds some additional targets to Symbiflow to generate binary, header, openocd and jlink files.

Also it allows passing vendor and part via Makefile to allow the Edalize launcher to select the correct Docker image.

Sample launcher script: https://gist.github.com/carlosedp/ac36e9052bfa23f84b2f42f1a05ee1bd

This will also be sent as a PR soon.


carlosedp:
The new targets still depend on upstreaming some code from Quicklogic to Symbiflow repository due to scripts update.

I could separate this into two PRs if needed (one with targets and another with the variables).

olofk:
Thanks @carlosedp! Let's go with this. Picked and pushed

carlosedp:
Awesome ;)

PR #299, opened 2022-01-18
Improve launcher script with new containers

carlosedp:
This launcher script supports multiple Xilinx and Quicklogic devices.

It depends on PR #298 as it requires the vendor and part to be exposed as environment variables.

Implements containers from https://hdl.github.io/containers/

### DIFF
--- a/scripts/el_docker
+++ b/scripts/el_docker
@@ -1,37 +1,137 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 import os
+import shutil
 import subprocess
 import sys
+import logging
+import shlex
 
-containers = {
-    "flow.tcl": "edalize/openlane-sky130:v0.12",
-    "yosys": "hdlc/yosys",
-    "nextpnr-ice40": "hdlc/nextpnr",
-    "nextpnr-ecp5": "hdlc/nextpnr",
-    "icepack": "hdlc/icestorm",
-    "ecppack": "hdlc/prjtrellis",
-    "icetime": "hdlc/icestorm",
-    "icebox_stat": "hdlc/icestorm",
-    "ghdl": "ghdl/ghdl:buster-llvm-7",
-    "verilator": "verilator/verilator",
-}
+logger = logging.getLogger(__name__)
+symbiflow_init = "bash -lec {}"
+
+# fmt: off
+containers = [
+    {'tool': 'ghdl',                      'image': 'ghdl/ghdl:buster-llvm-7',                         'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'verilator',                 'image': 'verilator/verilator',                             'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'flow.tcl',                  'image':  'edalize/openlane-sky130:v0.12',                  'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'mill',                      'image': 'adoptopenjdk:8u282-b08-jre-hotspot',              'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'sbt',                       'image': 'adoptopenjdk:8u282-b08-jre-hotspot',              'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'yosys',                     'image': 'hdlc/yosys',                                      'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'nextpnr-ice40',             'image': 'hdlc/nextpnr',                                    'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'nextpnr-ecp5',              'image': 'hdlc/nextpnr',                                    'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'nextpnr-nexus',             'image': 'hdlc/nextpnr',                                    'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'icepack',                   'image': 'hdlc/icestorm',                                   'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'ecppack',                   'image': 'hdlc/prjtrellis',                                 'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'icetime',                   'image': 'hdlc/icestorm',                                   'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'icebox_stat',               'image': 'hdlc/icestorm',                                   'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'icepll',                    'image': 'hdlc/icestorm',                                   'init': '', 'vendor': '', 'part': ''},
+    {'tool': 'ecppll',                    'image': 'hdlc/prjtrellis',                                 'init': '', 'vendor': '', 'part': ''},
+    # For Quicklogic EOS S3
+    {'tool': 'ql_symbiflow',              'image': 'gcr.io/hdl-containers/conda/symbiflow/eos-s3',    'init': symbiflow_init, 'vendor': 'quicklogic', 'part': 'ql-eos-s3'},
+    {'tool': 'symbiflow_synth',           'image': 'gcr.io/hdl-containers/conda/symbiflow/eos-s3',    'init': symbiflow_init, 'vendor': 'quicklogic', 'part': 'ql-eos-s3'},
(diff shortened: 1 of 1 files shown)


carlosedp:
I still would like to test this more when Quicklogic Symbiflow changes hit upstream as I describe here: https://gist.github.com/carlosedp/ed17d934c94338cbb2f9b29dd076bf1a

olofk:
Thanks @carlosedp. It would be great to get this added but I think we need to figure out if it can be solved more easily. I had no idea that symbiflow provided different containers depending on which device is being used. That definitely makes things a bit messier. Ideally I would like to decompose SymbiFlow so that Edalize directly calls yosys, VPR/Nextpnr and the bitstream generation tools directly because there's a lot of things that become hidden with the Symbiflow wrapper scripts. But that requires a bit more work so hopefully we can find some intermediate solution.

I would like to start with figuring out why there are different containers for each Xilinx part and if we can find a better way to handle that

carlosedp:
I think @umarcor can give a better explanation but the containers are built like this to avoid having a single container that is more than 23GB unpacked (the Xilinx 7 series databases are huge).

olofk:
Ah ok, so it's the database files. Then I propose that those are handled outside the container. I just realized that I might already have solved that problem when I was doing some experiments with SymbiFlow a couple of months ago. I created a small script that downloads and caches the database files https://gist.github.com/olofk/0f3911e7fa1028cc4c445bb2039ba06e and prints out the path to them. Maybe we can use that and have a symbiflow container that doesn't contain any database files instead?

umarcor:
> I had no idea that symbiflow provided different containers depending on which device is being used.

SymbiFlow as an organisation does not provide containers. The containers being used here are provided in hdl/containers.
This is to mean that those containers are *not* the officially endorsed solution in SymbiFlow.
Precisely, installing Conda inside containers is disruptive in comparison to all other tool containers from hdl/containers. 
That was done as a very temporary solution to have something usable in CI as a ready-to-use environment, and to allow projects such as edalize launchers, openflow or EDAA to consume those tools through container wrappers.

As you said, the complexity of SymbiFlow (and OpenROAD) is they are composed by a dozen smaller tools, at least, and not all of them are prepared for being individually packaged and distributed. Some are huge tools, some are rather small scripts. Therefore, while hdl/containers is slowly enhanced to add each tool individually, the all-included Conda solution is provided.

> I would like to start with figuring out why there are different containers for each Xilinx part and if we can find a better way to handle that.

You (users) can use `conda/symbiflow/xc7`, which includes definitions for all devices except xc7a200t. See https://hdl.github.io/containers/ToolsAndImages.html#symbiflow-conda. However, that's 18GB when the container is run, compared to 6-9GB for the per-device containers. So, the main reason is size, which has impact in traffic and provisioning time in CI.

Anyway, there are two solutions to have everything included in a single image:

- Build your own image based on `conda/symbiflow/xc7`, which adds the xc7a200t data. That cannot be done in hdl/containers at the moment, due to GitHub Actions' limitations. However, we expect to start using the GCP in the following months.

- Enhance edalize and the launcher to support pre-downloading the definitions to a local directory (or to a volume) and then bind that location/volume to the container. This is particularly interesting to help reverse-engineering efforts by keeping data local and tools on containers.

umarcor:
> Maybe we can use that and have a symbiflow container that doesn't contain any database files instead?

That is container `conda/symbiflow/xc7/toolchain`.

olofk:
Thanks for the info @umarcor. I will probably try to use the bare container then and fetch database files on demand

olofk:
@umarcor and @carlosedp I'm having an issue now that I can't work out. Launching symbiflow in the containers make them complain about missing prjxray-config. Have you encountered this? The file is in the container but not in the path so I wonder if I'm missing some step

olofk:
Ok, I see the discussion in https://github.com/hdl/containers/issues/49 now. Trying to do things piece by piece and missed the init part of the revised el_docker

olofk:
I made some feeble attempts to improve the situation by adding a script that calls symbiflow_* through bash instead of launching bash as a command-line argument to the container.

After spending an entire evening going crazy over argument quoting I ended up with a dockerfile that looks like this
```
FROM gcr.io/hdl-containers/conda/symbiflow/xc7/a100t:latest as build
ADD symbiflow_synth /usr/local/sbin/symbiflow_synth
ADD symbiflow_pack /usr/local/sbin/symbiflow_pack
```

a symbiflow_synth script that looks like this
```
#!/bin/sh
bash -lec "/usr/local/bin/symbiflow_synth '$*'"
```

That works for symbiflow_synth, but the corresponding script for symbiflow_pack
```
#!/bin/sh
bash -lec "/usr/local/bin/symbiflow_pack '$*'"
```

gives me `Unexpected command-line argument '-d'`. It all feels like a big hack to get around this damn conda thing and I hope to strip out the conda stuff eventually. But if we can get this working so that you can just launch the container with symbiflow_synth like the other ones, then we have come one step forward at least

umarcor:
@olofk:

1. You can change the entrypoint in your Dockerfile, and set it to `bash -lec` (maybe `ENTRYPOINT ['bash', '-lec']`?). That avois having to add fragile wrappers, and it also avoids @carlosedp's `init` solution.
  - However, I would recommend @carlosedp's solution, because it makes easier to support containers with or without profile sourcing requirements.
  - Note that the default shell in hdl/containers images is `bash`, not `ash`, `dash`...

2. If bash wrappers are used, you might put `le` in the shebang, to avoid using `-c` (which makes escaping/quoting fragile). See https://github.com/hdl/packages/blob/main/.github/workflows/containers-conda-symbiflow.yml#L96. Then, you don't need to modify the entrypoint in the dockerfile and you don't need `init`.

  - Alternatively, you can activate the conda environment explicitly, as @carlosedp did in https://github.com/hdl/containers/pull/26/commits/a0ae2bce2b8faf8fbdc6a644d6e918295760441c#diff-19620fbf3bbcba01ef7d42a38960c97b03769697bdd704e8ceafee32eaf99fa2R3:

```bash
#!/bin/bash

source "/opt/conda/etc/profile.d/conda.sh"

exec "$@"
```

olofk:
Fantastico @umarcor! I chose to do like @carlosedp's original docker image and activate conda. It means maintaining a separate image, but that's fine. It's temporary (famous last words :)) I hope to eventually shed a few layers from the symbiflow scripts so that Edalize calls yoys, vpr, etc. directly.

olofk:
Ideally, I would like to have the database files outside of the containers and rewrite the symbiflow backend and a lot more things, but I realize that would take too much time and this PR solves the problem for now. So let's pull in this (and https://github.com/olofk/edalize/pull/298) and we can improve things later.

Thanks for your work!

carlosedp:
Yay!

carlosedp:
I think in the future, @umarcor could create an API publishing the available containers and the script can query that when it receives the called utility.

umarcor:
@carlosedp since containers are published to registries, and there is an specification for registry APIs, hdl/containers does have an API to query the available containers already. See, for instance, https://github.com/hdl/containers/tree/explorer/utils/HDLCE.

Moreover, the dockerfiles used for each image, and which images are built in each CI run, are defined in a YAML file. See https://github.com/hdl/containers/blob/main/utils/pyHDLC/config.yml. See also the dataclasses and helpers in https://github.com/hdl/containers/blob/main/utils/pyHDLC/__init__.py.

I expect to used that info in order to generate the [Graphs](https://hdl.github.io/containers/dev/Graphs.html) automatically, which would allow to programmatically query image dependencies. In other words, to programmatically get the steps required in a pipeline in order to build all the dependencies of some specific image. However, there is no ETA.

umarcor:
Refs:

- https://docs.docker.com/registry/spec/api/ for docker.io
- https://cloud.google.com/container-registry/docs/reference/docker-api for gcr.io

carlosedp:
That's very true, I forgot I could query them directly... gonna think on an evolution to the script. Thanks!

carlosedp:
@olofk Edalize is not finding the el_docker script by itself just by specifying `EDALIZE_LAUNCHER=el_docker`. Is it a bug or do I need do put the full path to it?

PR #301, opened 2022-01-28
Fixed quotation for vhdl generics for Xilinx Xsim

Steinegger:
Overrides fail without quotation marks, the PR fixes this issue.
Resolveses olofk/fusesoc#543

Related Xilinx support article: https://support.xilinx.com/s/article/64118

### DIFF
--- a/edalize/xsim.py
+++ b/edalize/xsim.py
@@ -128,7 +128,7 @@ XSIM_OPTIONS  = {xsim_options}
             gen_param.update(self.generic)
             gen_param_args = " ".join(
                 [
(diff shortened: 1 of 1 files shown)


olofk:
Thanks for the patch. The change looks fine, but the reference files for the test suite must be updated as well to reflect this change. The easiest way to do this is to run the test suite with the environment variable GOLDEN_RUN set, e.g. run `GOLDEN_RUN=1 pytest` from the Edalize root. This will update the files and you can see which test suite files you need to add to the patch

olofk:
I took the liberty to update your patch with the corrected reference files. Thank you for your contribution! Picked and pushed

PR #303, opened 2022-02-02
Port Vivado to new Edalize API

acomodi:
This PR ports the current Vivado Edalize support to use the new API based on the clear separation between `flows` and `tools`.

Vivado is enabled to support both a full Vivado flow or a yosys+Vivado one.

Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>


edalize/vivado.py
```diff
@@ -46,11 +44,6 @@ def get_doc(cls, api_ver):
                         "type": "String",
                         "desc": "FPGA part number (e.g. xc7a35tcsg324-1)",
                     },
-                    {
-                        "name": "board_part",
-                        "type": "String",
-                        "desc": "Board part number (e.g. xilinx.com:kc705:part0:0.9)",
-                    },
                     {
```
olofk, line 47: Suspecting a rebase mistake here

olofk:
Awesome! Thanks for this. Think there's a rebase error for board_part but other than that it looks perfectly fine

edalize/vivado.py
acomodi: Yep, fixed

shareefj:
@olofk, it looks like this PR has missed the `board_repo_paths` list from the `tool_options`.  Does `TOOL_OPTIONS` allow the same differentiation of `lists` vs `members` as the old flow?

Readon:
In this patch the vivado flow has its own build_pre and build_post which do nothing but inherit the methods from Edatool.
So the scripts defined as pre_build hooks does not work anymore.
``` yaml
  target:
    hooks:
      pre_build : [ip_repo, update]

  scripts:
    ip_repo:
      cmd : [sed, -i, '0,/set_property/!b;//a\set_property ip_repo_paths ../src [current_project]', xxx.tcl]
    update:
      cmd : [sed, -i, '0,/ip_repo/!b;//a\update_ip_catalog -rebuild', xxx.tcl]
```

PR #304, opened 2022-02-02
Add initial and basic VPR support with the new Edalize API

acomodi:
This PR adds a basic support for VPR for the new Edalize API and calls the tool directly, without passing through wrappers.

I believe this is quite similar to https://github.com/olofk/edalize/pull/271, and currently it still requires the XML architecture file to be provided by the user. As per the discussion in #271, I believe that in the long term we should enhance VTR to have a pckg-config-like feature, that allows to have a `share` dir from which VTR automatically selects the correct arch, RR graph, and other files required for the build.

I think that as a first step we may allow the user to pass the paths to these files, and once VTR will have the `pkg-config` feature, we can fix the VPR support in Edalize accordingly.

Signed-off-by: Alessandro Comodi <acomodi@antmicro.com>


edalize/flows/vpr.py
```diff
@@ -0,0 +1,31 @@
...
+        ("yosys", ["vpr"], {"output_format": "blif"}),
+        ("vpr", [], {}),
+    ]
+
+    FLOW_OPTIONS = {}
+
+    def __init__(self, edam, work_root, verbose=False):
+        Edaflow.__init__(self, edam, work_root, verbose)
+
+    def build_tool_graph(self):
```
olofk, line 25: I believe this function can be dropped since it just calls the parent class

edalize/vpr.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    @classmethod
+    def get_doc(cls, api_ver):
+        if api_ver == 0:
+            return {
+                "description": "The VTR backend performs Packing, Placement, Routing & Timing Analysis.",
+                "lists": [
+                    {
+                        "name": "board_repo_paths",
```
olofk, line 33: Merge mistake?

olofk:
I agree about the XML arch files. Let's go with this for now and we'll solve how to find them programatically later. FYI, I have started a project on this already with the working name [pdk-config](https://gist.github.com/olofk/a31e94c834e3de6cb20c5424cd27e7bb) focusing on symbiflow and openlane to begin with.

Added a few comments above, and then there's one more thing that needs to be fixed. Each tool must create a new EDAM struct with its outputs so that the next tool in the chain can pick it up. See [here](https://github.com/olofk/edalize/blob/master/edalize/tools/nextpnr.py#L72) for an example of how nextpnr does this

edalize/flows/vpr.py
acomodi: fixed

edalize/vpr.py
acomodi: fixed

PR #306, opened 2022-02-11
ISE backend: add support for specifying the JTAG chain index

codepainters:
This fixes a problem that I've encountered while trying to run `fusesoc` Blinky on Digilent Spartan-S3E Starter board.

Namely, this Xilinx Impact batch doesn't work (failing with device ID mismatch):

```
setMode -bscan
setCable -port auto
addDevice -p 1 -file ./blinky.bit
program -p 1
quit
```
... because with more devices on the JTAG chain (`XCF04S` PROM and `XC2C64` CPLD in my case) it seems necessary to either specify the whole chain or scan it. 

With this change the chain is first scanned with `identify` and a file is then assigned to the device with `assignFile`:
```
setMode -bscan
setCable -port auto
identify 
assignFile -p 1 -file ./blinky.bit
program -p 1
quit
```

I've also added `board_device_index` parameter allowing for specifying chip JTAG chain location (defaults to 1).

Note: I'm not sure about impact (no pun intended) of this change on other boards, I've no hardware to check it. But I expect no issues.

### DIFF
--- a/doc/edam/api.rst
+++ b/doc/edam/api.rst
@@ -154,14 +154,15 @@ yosys_synth_options List of String        Options for Yosys Synthesis
 ise
 ~~~
 
-================ ===================== ===========
-Field Name       Type                  Description
-================ ===================== ===========
-family           String                FPGA family e.g. *spartan6*, *virtex5*
-device           String                Device identifier e.g. *xc6slx45*
-package          String                Device package e.g. *csg324*
-speed            String                Device speed grade e.g. *-2*
-================ ===================== ===========
+================== ===================== ===========
+Field Name         Type                  Description
+================== ===================== ===========
+family             String                FPGA family e.g. *spartan6*, *virtex5*
+device             String                Device identifier e.g. *xc6slx45*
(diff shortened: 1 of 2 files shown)


olofk:
Thank you for the thorough explanation. I'm always a bit hesitant towards adding features around device programming because I really want to get that out of Edalize and into some separate library. But it seems like a solid addition and fixes your problem so let's go with it. Thanks!

olofk:
Ok, one small issue complaint from the CI. The source formatting tool `black` wants some changes. Could you update to make black happy? I think you need to run `pre-commit install` from the root of your Edalize repo. Then it will run syntax checks every time you try to commit something

codepainters:
Thanks for the hint, indeed I missed some code style requirements :)

imphil:
@codepainters Amending commits and force-pushing is best practice, so you did everything right! Thanks!

olofk:
Agree! That's just the way I want it too :) Thank you for your contributions. Picked and pushed!

PR #307, opened 2022-02-14
Mistral backend for edalize

infphyny:
This Mistral backend is based on oxide.  Generating bitstream work <del> but pytest failed. </del>

<del > Current state: mistral_test is incomplete.  2 test pass: test_mistral_multiple_qsf and test_mistral_no_device, 2 test failed: test_mistral and test_mistral_minimal. I seek for advice and/or help to complete test_mistral. </del>

<del> My understanding is the initial call for pytest need to set GOLDEN_RUN to copy script in test_mistral folder. Unfortunately not all script are copied. What missing is yosys.cmd and nextpnr-mistral.cmd.  </del>

<del>Here is the error from pytest </del>: 
```
----------------------------- Captured stdout call -----------------------------
yosys -l yosys.log -p 'tcl edalize_yosys_template.tcl'
nextpnr-mistral -l next.log --device 5CSEBA6U19I7 --json test_mistral_0.json --rbf test_mistral_0.rbf
----------------------------- Captured stderr call -----------------------------
Info: Initialising bels...
Info: Initialising routing graph...
Info:     imported 2739969 wires and 25465239 pips
ERROR: Failed to open JSON file 'test_mistral_0.json'.
0 warnings, 1 error
make: *** [Makefile:9 : test_mistral_0.rbf] Erreur 255
=========================== short test summary info ============================
FAILED tests/test_mistral.py::test_mistral - RuntimeError: '['make']' exited ...
FAILED tests/test_mistral.py::test_mistral_minimal - RuntimeError: '['make']'...
======================== 2 failed, 102 passed in 29.97s ========================

```


infphyny:
There are at least two things that need to be assessed. 
1) The optional [mistral-cv](https://github.com/olofk/fusesoc/issues/552#issuecomment-1037155354) command is not implemented.
2) In addition of the required device option for nextpnr-mistral, there should also have a required family option for yosys. The mistral backend work because the default family option for yosys is cyclonev when arch = synth_intel_alm . This need modification of the yosys backend and maybe tcl script template.

olofk:
Cool! Thanks for working on this. I think we can do a first iteration without mistral-cv and add that later. Not 100% sure I get the family issue. Is the command something like `synth_intel_alm -family cycloneiv -other options`? I see a commented line where you extend the yosys_options with the family stuff. Is there some issue with doing it like that?

infphyny:
Exactly, when synth_intel_alm  is set as architecture, -family setting can be cyclonev (default), ariav or cyclone10gx . So hard coding cyclonev family settings in mistral backend is working. 
Even if family settings is not required for mistral backend, I think it is a good idea to add it just in case yosys default family setting change in the future.  

That line of code does the trick.
`
    "yosys_synth_options": ["-family","cyclonev"] + self.tool_options.get("yosys_synth_options",[])  ,
`
 What was in comment before family settings was over written in the test_mistal  but was correct in test_mistral_minimal.
 Now family option show in tcl script for both test.

I will commit my change soon.

olofk:
I was rather thinking of doing it the opposite way and not specifying anything so that it's up to the user to set it through yosys_synth_options e.g. in their FuseSoC core description files.

I see your point about hard coding it to cyclonev in case it changes upstream, but I prefer to hard code as few things as possible in the backends to follow upstream as close as possible.

So if we drop "yosys_synth_options": ["-family", "cyclonev"] I'm ready to pull it in. Is that ok with you?

infphyny:
I agree with you to drop "yosys_synth_options": ["-family", "cyclonev"]. I will apply change and commit soon. Thanks.

olofk:
That looks perfectly fine! Thank you for the contribution. Picked and pushed!

PR #308, opened 2022-02-14
Add Project XRay backend currently supporting Kintex7

carlosedp:
This PR adds support to Project XRay supporting Xilinx series 7 FPGAs.
Currently I'm testing with Kintex 7 and a QMTech board.

The work is based on https://github.com/kintex-chatter/xc7k325t-blinky-nextpnr and is in draft (although working) state.

The idea is to open this for discussion.


umarcor:
@carlosedp note that the "backend" is nextpnr in this case (VTR is supported as well). Project XRay is the "architecture documentation/database", equivalent to IceStorm, Project Trellis or Project Nexus. See:

- https://github.com/hdl/containers/blob/main/debian-bullseye/nextpnr.dockerfile
- https://github.com/hdl/containers/blob/main/.github/workflows/nextpnr.yml
- https://github.com/hdl/containers/blob/main/debian-bullseye/prjtrellis.dockerfile
- https://github.com/hdl/containers/blob/main/.github/workflows/prjtrellis.yml

carlosedp:
Ah yes, my idea here is to create the Project Trellis equivalent for XRay where Yosys backend is used by default but since nextpnr-xilinx is kinda different from the standard nextpnr, I decided now to keep it in the xray file instead of "plugging" the nextpnr backend.

Also the build and databases is a bit different so I created the Dockerfile myself. I believe it's still early since I used forked versions from @hansfbaier that did the amazing work starting this.

PR #310, opened 2022-03-16
Restore missing board_repo_paths

shareefj:
The recent rewrite of the Vivado backend seems to have dropped support for board_repo_paths.


shareefj:
This fix seems to work.  What I was seeing was that the `vivado-project.tcl.j2` template wasn't rendering the `board_repo_paths`.

I've just added a test to catch this as well.

PR #311, opened 2022-03-29
Fix formatting/CI check

imphil:
Commit f764e7efa3e4b86553930e708db35b0bf7cc7953 broke the black formatting
CI check. Fix that.

Also update black to a version that works with recent version of the Click dependency.


olofk:
All good! Thanks for the fix

PR #313, opened 2022-04-08
Handle unset file_type in edalize/tools

rswarbrick:


olofk:
Thank you! I've been a bit too careless about that :/ Picked and pushed

rswarbrick:
Thanks! It's super-obvious when it goes wrong, but really hard to spot until then :-)

PR #316, opened 2022-05-06
Limit CPU usage in Verilator build #210

shareefj:
Verilator builds previously used all available CPUs by calling Make with the -j<cpu_count>
argument. This is fine when running on your local machine but is not fine when running the
build in a professional environment where you are submitting the job to a cluster.

This update removes the -j<cpu_count> argument from the build step.  Users wishing to use
N CPUs for a build should instead pass the MAKEFLAGS environment variable when calling
fusesoc:

  export MAKEFLAGS=-jN; fusesoc ...

### DIFF
--- a/edalize/verilator.py
+++ b/edalize/verilator.py
@@ -3,7 +3,6 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 import logging
-import multiprocessing
 import os
 import logging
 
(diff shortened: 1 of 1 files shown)


imphil (approved):
LGTM, thanks for the PR. I wish we had a way to better document that. 

@olofk can you ensure this change makes it into the release announcement? (There's no changelog for edalize AFAIK.)

@rswarbrick FYI

rswarbrick:
This looks sensible to me. @imphil, thanks for the heads-up!

edalize/verilator.py
```diff
@@ -207,13 +206,9 @@ def build_main(self):
         if not "mode" in self.tool_options:
             self.tool_options["mode"] = "cc"
 
-        # Do parallel builds with <number of cpus>
-        make_job_count = multiprocessing.cpu_count()
-        args = ["-j", str(make_job_count)]
-
         if self.tool_options["mode"] == "lint-only":
             args.append("V" + self.toplevel + ".mk")
```
mpriestleyidex, line 210: This line is now broken - `args` is undefined (and unused)

shareefj: Blimey, helps if you add the file....

olofk:
Sorry for the delay. Was sure I had already merged this one. Many thanks for your contributions. Picked and pushed!

PR #318, opened 2022-05-18
Gatemate backend

trabucayre:
This PR add support for cologneChip's GateMate toolchain: this one is based on **yosys** for synthesis and **p_r** for PnR.

Since **p_r**'s input is a verilog file the first commit introduce support to write synthesis output as verilog file when `output_format` is set to **verilog**. To do so and to avoid potential conflict beween verilog source and synthesis output a new parameter **output_name** is introduce, when not set the original behaviour (`name.output_format`) is used. `yosys-script-tcl.j2 ` is modifed too to use *output_name* instead of forging again ouput name.

Note: **p_r** is currently windows only, linux version has not been already released. It's required to use **wine** to use it. To avoid to explicitly call wine in edalize backend I have a script in my $PATH similar to:
```bash
#!/bin/sh
wine /somewhere/cc-toolchain-win/bin/p_r/p_r.exe $@
```
\cc @pu-cc


trabucayre:
I will fix yosys format issue ASAP when everything is OK for everyone with this new support (in another commit)

olofk:
Many thanks! This looks great and I look forward to test it myself. The only thing that I'm sceptical about is hard coding the output file in the yosys template instead of using `$name`, but I guess it makes no difference in practice and perhaps just avoid one more abstraction layer.

Thank you for your contributions. Picked and pushed!

trabucayre:
Thanks!
For hardcoding output file in `ys`: I have hesitate a bit for this solution but if user select to use current directory as build directory and if this one contains a file with `$name.v`! yosys output will ovewrite source so it's more easy/secure to "forge" the name at `yosys.py` level to be sure to avoid this wrong side effect.

PR #321, opened 2022-06-06
Add fix for ISE path handling on Windows

panther03:
By default, the ISE backend crashes when configuring a project on Windows, because the Tcl scripts don't handle Windows '\' path separators. Here's an example of the error when building the `blinky` project for the `pipistrello` target (which uses Xilinx ISE by default):
`INFO: Building
xtclsh fusesoc_utils_blinky_1.1.1.tcl
ERROR:TclTasksC:xfile_070: File(s) "./..src♀usesoc_utils_blinky_1.1.linky.v" cannot be found
    while executing
"xfile add ..\src\fusesoc_utils_blinky_1.1.1\blinky.v"
    (file "fusesoc_utils_blinky_1.1.1.tcl" line 21)
make: *** [Makefile:10: fusesoc_utils_blinky_1.1.1.xise] Error 1
←[1;31mERROR: Failed to build fusesoc:utils:blinky:1.1.1 : '['make']' exited with an error: 2←[0m`

This patch simply enables the option in _get_fileset_files to replace the path separators by default, for ISE projects.
Even on Windows, the '/' separator works just fine with Tcl.

### DIFF
--- a/edalize/ise.py
+++ b/edalize/ise.py
@@ -151,7 +151,7 @@ quit
                 )
             )
 
-        (src_files, incdirs) = self._get_fileset_files()
+        (src_files, incdirs) = self._get_fileset_files(force_slash = True) # ISE tcl doesn't like '\', so '/' is forced
 
         if incdirs:
             tcl_file.write(


olofk:
Aha. ISE too? Shouldn't be too surprised. At this point I think pretty much _all_ the tools prefer regular slash even on Windows. Thank you for your contributions.

Unfortunately it doesn't pass one of the CI tests because the code needs to follow the conventions set up by `black`. I think the way to get this fixed is to run `pre-commit install` in the root of your checked out Edalize repo. This will then perform some basic checks and automatically reformat the source code.

Could you do that and submit an updated patch?

olofk:
I took the liberty to fix up the formatting so that we could pull this one in. Thank you for your contribution

PR #324, opened 2022-06-21
Update spyglass.py

jgbhatia:
The run_main runs the spyglass tool. There are two targets in the make file. args = ["-i"] option is required to ignore the error code from the first design read target and run the lint target.

### DIFF
--- a/edalize/dc_shell.py
+++ b/edalize/dc_shell.py
@@ -1,0 +1,111 @@
+# Copyright edalize contributors
+# Licensed under the 2-Clause BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-2-Clause
+
+import logging
+import re
+from collections import OrderedDict
+
+from edalize.edatool import Edatool
+
+logger = logging.getLogger(__name__)
+
+
+class Dc_shell(Edatool):
+
+    _description = """ Synopsys (formerly Atrenta) DC_shell Backend
+
+This module generates the file list for dc_shell. 
+"""
+
+    tool_options = {
+        "lists": {
+            "dc_options": "String",
+        },
+    }
+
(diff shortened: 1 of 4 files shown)


olofk:
Hi @jgbhatia. Thank you for this patch. I'm not completely sure how the spyglass backend works but if I understand correctly it tries to run multiple goals and if any of them fails, it will not run the other ones. I wonder if everyone wants to have it this way or if we should add an option so that users can select which behavior they want. Perhaps @imphil can give some input here?

Your fix looks fine, but I think we can make it simpler. spyglass doesn't handle plusargs (AFAIU) so we only need to set the `-i` flag, not parse plusargs in `run_main`.

jgbhatia:
Hi @olofk, thanks for your response. When I configured fusesoc for my design setup, I found that the setup was created for spyglass but the spyglass did not run. It was because the run_main routine was not present. After adding the run_main routine, I hit another issue. The makefile has two targets, run-goal-design_read & run-goal-lint_lint_rtl. It always used to run the design_read goal and exit with an error. Hence I added -i option so that make can ignore the error and execute the next goal.
The errors in the design_read goal can be due to various reasons. One  of the most common reason for the error is due to the use of standard cells or hard macros. The fix for this issue required use of read_file -type gateslib command, which I could not find. So -i option was the easiest one to work around the error.
With some more debugging, I found that this behavior is related with the "set_option enable_pass_exit_codes yes" option. I am not sure about the history behind it. But setting this option to no, removing it from makefile and removing -i argument makes spyglass run as expected. This  code is not yet checked in into git. Based on the discussion, I can push the changes in the repository for review and merge.
We can discuss the support for gateslib file type in a separate thread.

jagriti-bhatia:
I have pushed the changes as mentioned by @jgbhatia

PR #325, opened 2022-07-01
F4PGA Flow: restructured tests

Pocketkid2:
This PR includes (and thus requires) #322 

Now tests both VPR and NextPNR flows for F4PGA. VPR uses Yosys the way Symbiflow did, and NextPNR uses Yosys the default way. Also checks for openocd config file. These tests are intended for Xilinx 7 series boards as that is what the new F4PGA module currently supports.


Pocketkid2:
Closing this PR as its parent PR has been superseded

PR #326, opened 2022-07-01
Example documentation for F4PGA

Pocketkid2:
This PR includes (and thus requires) #322 and #325

First, a minor fix adding the pnr tool option to the EDAM API page

Second, a new page is created where examples of certain flow scripts can be put. Two examples are given for F4PGA and some basic explanation, one with the default setting (VPR for place and route) and another for using NextPNR.


Pocketkid2:
Closing this PR as its parent PRs have been superseded

PR #328, opened 2022-07-06
Refactor and introduce get_tool_options in flow

olofk:
Before this change, there was no meaningful way to use flow options
because it was not possible to query or set tool options for tools
outside of the ones in the static FLOW list.

This patch separates the process of applying options into two steps.
A user can first use get_flow_options to get flow options and then
use get_tool_options with the preferred flow options set, to get
all tool options for a configured flow.

The redefinition of get_flow_options to only return flow options
(and not also tool options like previously) is an API break but will
hopefully not cause too much trouble.

### DIFF
--- a/edalize/flows/vivado.py
+++ b/edalize/flows/vivado.py
@@ -12,30 +12,53 @@ class Vivado(Edaflow):
 
     argtypes = ["vlogdefine", "vlogparam"]
 
-    FLOW = [
-        ("yosys", ["vivado"], {"arch": "xilinx", "output_format": "edif"}),
-        ("vivado", [], {}),
-    ]
+    FLOW_DEFINED_TOOL_OPTIONS = {
+        "yosys": {"arch": "xilinx", "output_format": "edif"},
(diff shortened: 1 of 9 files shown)


edalize/flows/vivado.py
```diff
@@ -12,30 +12,55 @@ class Vivado(Edaflow):
...
+
+        if flow_options.get("synth") == "yosys":
+            flow.append("yosys")
+        flow.append("vivado")
+
+        return cls.get_filtered_tool_options(flow, cls.FLOW_DEFINED_TOOL_OPTIONS)
+
+    def configure_flow(self, flow_options):
+        flow = []
+        if edam.get("flow_options", {}).get("synth", {}) == "yosys":
```
Pocketkid2, line 47: I believe this edam reference should be self.edam and that will fix your error

olofk: Thanks for taking a look. This is fixed, together with some other things that made the tests fail

PR #329, opened 2022-08-01
openroad: adding `openroad-flow-script`

joennlae:
Can be seen in action here:
https://github.com/joennlae/halutmatmul/blob/master/hardware/halut_top.core


olofk:
Very interesting. I haven't found the time to do a proper review though, but please ping me if I don't get to it

davideschiavone:
@olofk , this is an amazing contribution, thanks @joennlae !

I am actually testing it in here https://github.com/esl-epfl/x-heep/pull/181 and it seems to do a very well job

olofk:
I pushed the patches now that were related to openroad, but left the ones about the export and cocotb backends.

There are a number of changes I ideally would like to see in this backend, but I think we can go through them when we convert this to the flow API instead. Thank you for your contribution!

PR #331, opened 2022-08-08
Refactor VPR for f4pga/symbiflow usage (includes fixvpr branch change)

Pocketkid2:
This contains the changes to VPR needed for the F4PGA flow to work. I've included the fixvpr change as well so that you don't have to merge both. I also decided to put the variable names inside this file instead of the f4pga flow for the present, as the user isn't supposed to be changing them anyway and it simplifies the f4pga flow.

### DIFF
--- a/edalize/tools/vpr.py
+++ b/edalize/tools/vpr.py
@@ -29,6 +29,10 @@ class Vpr(Edatool):
             "type": "str",
             "desc": "Path to target architecture in XML format",
         },
+        "gen_constraints": {
+            "type": "bool",
+            "desc": "If true, activates the symbiflow gen_constraints python script as part of the VPR process",
+        },
         "vpr_options": {
             "type": "str",
             "desc": "Additional options for VPR.",
@@ -67,16 +71,21 @@ class Vpr(Edatool):
         """
         super().configure(edam)
 
-        src_files = []
-        incdirs = set()
-
-        file_netlist = []
+        netlist_file = ""
         timing_constraints = []
 
-        for f in src_files:
-            if f.file_type in ["blif", "eblif"]:
-                file_netlist.append(f.name)
-            if f.file_type in ["SDC"]:
+        for f in self.files:
+            file_type = f.get("file_type", "")
+
+            if file_type in ["blif", "eblif"]:
+                if netlist_file:
+                    raise RuntimeError(
+                        "VPR only supports one netlist file. Found {} and {}".format(
+                            netlist_file, f["name"]
+                        )
+                    )
+                netlist_file = f["name"]
+            if file_type in ["SDC"]:
                 timing_constraints.append(f.name)
 
         arch_xml = self.tool_options.get("arch_xml")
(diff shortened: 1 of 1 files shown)


Pocketkid2:
@olofk take a look as soon as you can

edalize/tools/vpr.py
```diff
@@ -90,27 +99,68 @@ def configure(self, edam):
...
+        depends = netlist_file
         targets = self.name + ".net"
-        command = ["vpr", arch_xml, self.name + ".blif", "--pack"]
+        command = ["vpr", arch_xml, netlist_file, "--pack"]
         command += sdc_opts + vpr_options
         commands.add(command, [targets], [depends])
 
-        depends = self.name + ".net"
+        # Run generate constraints script of flag set to true
+        if self.tool_options.get("gen_constraints", False):
```
olofk, line 109: `if self.tool_options.get("gen_constraints"):` is enough since `get` will return `None` which is a false value. (No big deal though. Just FYI). Also `of` should be `if` in the comment above

edalize/tools/vpr.py
```diff
@@ -90,27 +99,68 @@ def configure(self, edam):
...
+            depends = self.name + ".net"
+            targets = "${IOPLACE_FILE}"
+            commands.add(
```
olofk, line 112: I must confess I don't really understand where all these `${SOMETHING}` are defined? I kind of suspect they are supposed to be defined by the f4pga flow but that won't work if vpr is run stand-alone. Let's see for each of these if they can be expanded instead. I assume e.g. that `${OUT_EBLIF}` should be `netlist_file`, or did I misunderstand? Should `${NET_FILE}` be `self.name+.net`? I also think we need to offer the user some way to set `PART`. Should that be a tool option? Or can we read that somehow e.g. directly from the netlist or so?

olofk:
@Pocketkid2 Back from the wilderness now. Need to understand better how the env vars ares supposed to be used. Made some comments in the diff

edalize/tools/vpr.py
Pocketkid2: Fixed both in 6c43fcf

edalize/tools/vpr.py
Pocketkid2: @olofk It's true that this will cause a problem if the gen_constraints feature is run with VPR not being run through the F4PGA flow, but there is no other easy way to have this feature work properly the way F4PGA needs it. 

Some of those variables could be based on the already known stuff like you said, but there are also several variables that are unique to F4PGA and are traditionally set as environmental variables in the Makefile which means they would have to be set in the F4PGA flow, based on the install parameters. 

Previously I had this gen_constraints feature as an array parameter instead of a boolean parameter, where the user is forced to pass in an array of the variables, which prevents them from being hardcoded inside VPR, but it does look and feel a bit more awkward because you have to assume the size of the array. Which option do you think is best?

Pocketkid2: I should add that the PART variable comes from the main F4PGA flow, and is set by the board the user is targeting
 https://github.com/olofk/edalize/blob/7c3d2e01803c7744a89c26de8c0d3836eceb2489/edalize/flows/f4pga.py#L179

olofk: I understand that they have traditionally been set as environment variables, but is there a good reason for that? I really don't want to make the edalize backend unnecessarily complicated when we have an excellent opportunity to clean up this mess a bit. So please go through them to see which ones are actually needed. I'm confident it will be a much smaller subset than what's currently there.

When you say "based on the install parameters", do you mean that these are absolute paths to files within the the f4pga intallation? If so, that's a questionable design choice and I would highly recommend f4pga to do something about it, but if we have to, we can make it work. There are other backends that require certain environment variables to be set, see .e.g (https://github.com/olofk/edalize/blob/master/edalize/openfpga.py#L67). The important thing here is then to raise an error so that the user know they need to set these variables themselves (like in the openfpga example)

Pocketkid2: > I understand that they have traditionally been set as environment variables, but is there a good reason for that? I really don't want to make the edalize backend unnecessarily complicated when we have an excellent opportunity to clean up this mess a bit. So please go through them to see which ones are actually needed. I'm confident it will be a much smaller subset than what's currently there.

I can go through and get a minimized list of needed variables, but one of the reasons I had so many is because it simplifies the repetition used in each recipe. Do you think regular Make variables would work fine for that or should we just remove variables altogether and just plug the values in everywhere at Makefile generation?

Pocketkid2: > When you say "based on the install parameters", do you mean that these are absolute paths to files within the the f4pga intallation? If so, that's a questionable design choice and I would highly recommend f4pga to do something about it, but if we have to, we can make it work. There are other backends that require certain environment variables to be set, see .e.g (https://github.com/olofk/edalize/blob/master/edalize/openfpga.py#L67). The important thing here is then to raise an error so that the user know they need to set these variables themselves (like in the openfpga example)

There are F4PGA environmental variables, but some of them are set manually by the user at install time. The F4PGA architecture file directory structures are also not consistent across architectures, which I recently noticed and pointed out to the F4PGA people who are working to fix that. I think your suggestion is probably the best way forward, where we assume that the variables shown in the F4PGA install documentation are there, and throw errors if they aren't.

Pocketkid2:
@olofk What should my next step be? I can revise this change and move those variables back into the main f4pga flow if you think that would be better. Either way, the current f4pga flow requires the python scripts to run at the VPR stage, and with the modular "slight return" I figured the simplest way was to have that happen inside the VPR tool.

Pocketkid2:
@olofk I've just rewritten the VPR change so that the values have to be passed in from the parent flow. This should help abstract out the magic values and make it easier to fix the need for environmental variables by having the paths and values figured out in the f4pga flow class.  Take a look and let me know what you think and if we can pull something like this

PR #332, opened 2022-08-14
add check to ensure -pvector option is used only when output_format is edif

Seyviour:
This pull request edits the Yosys command generator to ensure that the "-pvector bra" option is only included for edif files.

According to the Yosys docs, the "-pvector bra" option applies only to edif files i.e when the issued command is [write_edif](http://yosyshq.net/yosys/cmd_write_edif.html). Except I am missing something (entirely possible since I'm a bit new to this),  "-pvector bra" should not apply when the command is [write_json](http://yosyshq.net/yosys/cmd_write_json.html). In my experience, this made the write_json command fail.

The other changes are necessary implications of the first, and were generated by running the test suite with the  "GOLDEN_RUN" environment variable set.

### DIFF
--- a/edalize/yosys.py
+++ b/edalize/yosys.py
@@ -130,7 +130,9 @@ class Yosys(Edatool):
             "synth_options": " ".join(self.tool_options.get("yosys_synth_options", "")),
             "write_command": "write_" + output_format,
             "output_name": default_target,
-            "output_opts": "-pvector bra " if arch == "xilinx" else "",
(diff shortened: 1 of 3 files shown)


olofk:
I think this looks perfectly fine. Thank you for your contribution. Picked and pushed!

Seyviour:
> I think this looks perfectly fine. Thank you for your contribution. Picked and pushed!

That's great! It's my pleasure to contribute.

PR #338, opened 2022-09-09
Use $(EDALIZE_LAUNCHER) envvar when calling xtclsh

hipolitoguzman:
The ``$(EDALIZE_LAUNCHER)`` environment variable was missing from the calls to ``xtclsh`` in the code that generates the ``Makefile``(s), so ISE docker images could not be used (the custom launcher was silently ignored)

The calls to ``impact`` seem to be a different beast since they are managed by the ``_run_tool()`` function, which (from the tests I have made locally) also doesn't seem to read the ``$(EDALIZE_LAUNCHER)`` variable. I suppose this would also happen with the other EDA tools that are invoked by the same function (such as ``radiantc``, ``quartus_pgm``, ``sby`` or ``vsim``), but I'm not familiarized enough with the codebase to really look into it.


olofk:
Thank you for this. Patch looks perfectly fine.

And as you have already seen, many of the current backends are a bit ad-hoc and require a lot of different patches to support EDALIZE_LAUNCHER.

There is however a new API called the flow API, which will have native support for this. Eventually all backends will be converted to use the flow API but there's still some work remaining and the ISE backend has not yet been converted

Issue #140, opened 2020-04-18
Interfacing edalize and bazel


mithro:
[bazel is Google's build system](https://bazel.build/). Currently it primarily targets things like C++ and Java server building, but it would be nice if it had strong support for interfacing to HDL toolchains.

I'm unsure what the best way for edalize to be interfaced to bazel, should;
 1) edalize generate bazel build files (in a similar manner to generating Makefiles)?
 2) bazel build files (which can support Python) call into edalize?
 3) edalize be used inside bazel in some way?

Ideally bazel would not have to duplicate all the code in edalize for doing things like parsing and understanding the output of various EDA tools.

mithro:
FYI - @hzeller

imphil:
@mithro what would be the goal of such an integration? Specify all source files in bazel description files, and use edalize to 
- call the EDA tools?
- create a "run configuration"/bazel snippet describing how to run the tool?
- something else?

mithro:
> @mithro what would be the goal of such an integration? Specify all source files in bazel description files, and use edalize to
> 
> * call the EDA tools?
> * create a "run configuration"/bazel snippet describing how to run the tool?
> * something else?

This is what I opened this issue to try and figure out! Should bazel be using edalize to call the EDA tools? Should edalize be generating configuration for bazel? Or is there something else better here?

fangism:
FYI - @snsokolov

olofk:
Hi @mithro . Thanks for raising the topic. I had similar discussions with the Meson devs last year. I know about Bazel, but have never used it so I can't really comment on the technical side. But I have been having some general high-level ideas about this on my mind for a long time. As the way things are now I believe the best way to start with is for Bazel to treat Edalize as a monolithic build step where you input and EDAM file togeteher with verilog, vhdl, constraints and all the other stuff that the EDA tools need and as a result you get a simulation model, FPGA image, netlist, or whatever you have set up your Edalize flow to produce.

Parallel to this I'm planning to exchange the current makefiles with ninja to improve the connection between the internal stages of edalize

In the longer run I'm considering ways to make Edalize even more modular and treat it as a series of more fine-grained processing steps. The current backends would then become pre-defined flows that just pull in the processing steps needed. But it would also allow to more easily create custom flows. And once we have come to that stage, it might make sense for Bazel-based builds to define these flows in the Bazel way instead, so that Bazel has full visibility of the artifacts produced between stages.

[...]

PR #346, opened 2022-10-02
Add support for user defined docker flags

solsjo:
Add support for supplying additional flags to docker such as mounting additional volumes.

This is useful e.g. if building via bazel and requesting a build using docker. The need stems from that el_docker is executed in the bazel sandbox. In the bazel sandbox all input sources only exists as symlinks. and docker won't have access to the real source files, unless the repo root is also added as a mounted volume.

This is oneway of enabling this.

Example project here:
https://github.com/solsjo/rules_edalize

### DIFF
--- a/scripts/el_docker
+++ b/scripts/el_docker
@@ -110,6 +110,7 @@ if [c for c in containers if c["tool"] == toolname]:
         logger.error("ERROR: Tool {} not found in container list.".format(toolname))
         exit(1)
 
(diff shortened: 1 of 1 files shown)


solsjo:
This pull-request aims at enabling support for interfacing bazel, see #140 and https://github.com/solsjo/rules_edalize

olofk:
Very nice! I set up the mounts in el_docker to be compatible with the way that FuseSoC sets up the build directory, but I hadn't thought about the case when Edalize is used directly. Thank you for your contribution. Picked and pushed!

PR #347, opened 2022-10-07
F4PGA Flow: The flow itself, plus some additional tweaks

Pocketkid2:
I've tried to keep this PR to a manageable size and I hope I've done okay. I have plenty to say and am totally open to your thoughts and comments and questions and critiques to make sure I'm going about this in the right direction. My intention is to have this and future PRs entirely supersede #322 and its branches (#325, #326, and #327).

First, the changes to utils.py are relatively simple and helped me fix a bug that was occuring at Makefile generation time relating to the command variables.

Next, the changes to tools/yosys.py:
- Reduced the number of arguments for F4PGA to just one, the part JSON file. This is used in the TCL script and is therefore unique to F4PGA flow, whereas the XDC file could be detected in the file list if we know to look for it.
- Make sure yosys template script isn't created unless it actually needs it (meaning there isn't a yosys template specified)
- Allow for a different log file name for F4PGA (not strictly necessary, I may change this later)

Then, the changes to tools/vpr.py:
- Rename the logfiles for each individual stage so that you can have individual log files for each stage (pack, place, etc.) since VPR apparently doesn't do that already. We may want to consider adding this as a generic feature for all uses of VPR, although I made it specific to F4PGA for this PR because it breaks the VPR tests.
- Refactor the generate constraints to use the new single script, and thus instead of passing an array of scripts and filenames we use the array to pass the arguments to the one script instead and take care of the rest.

And finally, we have the F4PGA Flow itself and its 'front-end' Edatool. For the moment I'm defaulting to Xilinx 7-series and VPR so that I don't clutter this PR, but those will be configurable in the future. (NextPNR is the next step I have planned after this)


olofk:
Awesome! Thanks for this and for the detailed explanation. I hope to begin review this soon

olofk:
Just picked d1fe937d4ea770e43aa4481, picked and squashed 503999f036caec681765ac49a455e3f39a012d7f + 93e0663b8c86bdee8211cc0e3812d3acb7790c38, picked and squashed f78fcd865c65e26bdda486568723da0d0eff5542 + a5cdb9fcd7cc20d63aa37de4a9d73775f863936a + f692966d11bfefacf48b3d79ae49a58f7fb5ccd1 and picked 7b4176ec57523194d77ae558805115c131caceb3

I.e. all the commits dealing with utils, yoss, vpr and el_docker. Will review the rest next time I get the chance

olofk:
I think moving the log files using mv is a bit shady, but let's figure out a better way later

olofk:
Other than that I really enjoyed the various cleanups. Well done

olofk:
I was more hoping we could pass some flag to vpr to set the output file name instead of it being hardcoded.

edalize/flows/f4pga.py
```diff
@@ -0,0 +1,220 @@
...
+            "desc": "Targeting architecture for Yosys. Currently supported is 'xilinx' and that is the default if none specified.",
+        },
+        # Optional, defaults to VPR
+        "pnr_tool": {
+            "type": "str",
+            "desc": "Place and route tool. Valid options are 'vpr'/'vtr' and 'nextpnr'. Defaults to VPR.",
+        },
+    }
+
+    # Standard F4PGA command line arguments to VPR
```
olofk, line 38: This is a _lot_ of default options. Perhaps that's what's needed, but remember that it is usually harder to override a hardcoded default, because we need to keep backwards compatibility, so if any of these are commonly modified by the user, then I suggest just thinking through if they should be removed from the list.

Or perhaps vpr supports overriding an argument by specifying it twice on the command-line. In that case it's not a problem

edalize/flows/f4pga.py
```diff
@@ -0,0 +1,220 @@
...
+
+        # Set target
+        # toplevel = self.edam["toplevel"]
+        self.name = self.edam["name"]
+        self.commands.set_default_target(f"{self.name}.bit")
+
+        # Set up nodes
+        synth_tool = "yosys"
+        pnr_tool = "vpr"
+        if "pnr" in flow_options and flow_options.get("pnr_tool") in ["nextpnr"]:
```
olofk, line 101: Is this a typo? Looks like it first checks for `pnr` and then `pnr_tool`. Given my comment above to just change to `pnr` I suggest simplifying it to  `pnr_tool = self.flow_options.get("pnr", "vpr")` to fall back to vpr if no tool is specified, and then

    if not pnr_tool in ["vpr", "nextpnr"]:
        raise(RuntimeError(f"Unsupported P&R tool: '{pnr_tool}'")

I realize now that we should also support setting pnr to `none` if we just want to do synthesis, but we can add that later

edalize/flows/f4pga.py
```diff
@@ -0,0 +1,220 @@
...
+            pnr_tool = "nextpnr"
+
+        self.device = flow_options.get("device")
+        if not self.device:
+            print("F4PGA flow error: missing 'device' specifier")
```
olofk, line 106: Use `raise(RuntimeError("F4PGA flow error: missing 'device' specifier"))` instead of print+return for this and the ones below

olofk:
I left some comments inline. I think that should be all

edalize/flows/f4pga.py
Pocketkid2: F4PGA sets all these options exactly, so I wasn't planning on allowing them to be configurable unless F4PGA decides to allow that. If you really think it's important I could write a system that would allow options to be passed in via tool_options or flow_options and then those would override the defaults, but I personally think that's a lot more effort than it would be worth right now.

edalize/flows/f4pga.py
Pocketkid2: Should be fixed now in #[48d4d21](https://github.com/olofk/edalize/pull/347/commits/48d4d2182391051a50e31a1fb6933a35211f7334)

edalize/flows/f4pga.py
Pocketkid2: Should be fixed now in #[48d4d21](https://github.com/olofk/edalize/pull/347/commits/48d4d2182391051a50e31a1fb6933a35211f7334)

edalize/flows/f4pga.py
```diff
@@ -0,0 +1,220 @@
...
+        "part": {"type": "str", "desc": "Example: 'xc7a35tcpg236-1'"},
+        # Required (used by VPR)
+        "chip": {"type": "str", "desc": "Example: 'xc7a50t_test'"},
+        # Optional, defaults to xilinx
+        "arch": {
+            "type": "str",
+            "desc": "Targeting architecture for Yosys. Currently supported is 'xilinx' and that is the default if none specified.",
+        },
+        # Optional, defaults to VPR
+        "pnr_tool": {
```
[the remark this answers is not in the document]
Pocketkid2: Fixed in #[48d4d21](https://github.com/olofk/edalize/pull/347/commits/48d4d2182391051a50e31a1fb6933a35211f7334)

Pocketkid2:
> I left some comments inline. I think that should be all

Thanks, I was able to make the fixes. I've rewritten the issue tracker that we had going [here](https://github.com/olofk/edalize/issues/323) to reflect the trajectory of the new PR. As soon as this one gets merged I'm going to close the previous PRs.

olofk:
And with that we're done! Squashed and merged. Thank you for your contributions. You don't happen to have som instructions on how to get the necessary prerequisites so that I can try it myself? :)

Pocketkid2:
> You don't happen to have som instructions on how to get the necessary prerequisites so that I can try it myself?

@olofk [This](https://f4pga-examples.readthedocs.io/en/latest/getting.html) is the guide that I recommend following. It will set up a conda profile for the install that you'll want to activate before doing anything f4pga related.

PR #348, opened 2022-10-17
F4PGA: Allow VPR options to be configured at the flow level

Pocketkid2:


olofk:
Sorry for the delay. All good

PR #353, opened 2022-10-28
F4PGA: Basic NextPNR-Xilinx support

Pocketkid2:
The main F4PGA project was designed for VPR as the place&route tool, so there are some differences in how the flow operates when compared to VPR. 

First, you must provide a chip database file that you can generate yourself from the NextPNR-Xilinx project. (In the future, we want to think about having this packaged with the tools and loaded based on what board/chip is specified in the flow options, but we probably need to communicate with the F4PGA people about that because I don't think we want to package those files inside Edalize)

Second, the custom F4PGA Yosys tcl script appears to have been designed with VPR in mind so unless it is modified or another made then we must jettison it for NextPNR-Xilinx and let the default Edalize Yosys template take over, targeting a JSON netlist. At least, this is what my own tests have shown. Ideally we would want F4PGA to use its tcl script for NextPNR too but at the moment I haven't gotten it to work. The good news is that the default Edalize Yosys template appears to work just fine.

Lastly, NextPNR-Xilinx uses one command to go from the JSON netlist all the way to FASM, whereas with the VPR flow the Edalize defaults to having each stage with its own command and then F4PGA provides a wrapper that makes it easy to generate the FASM. I haven't looked at the documentation of NextPNR-Xilinx very thoroughly to figure out if it can be broken into stages like VPR but that could be something I look into.

### DIFF
--- a/edalize/flows/f4pga.py
+++ b/edalize/flows/f4pga.py
@@ -97,9 +97,9 @@ class F4pga(Edaflow):
 
         # Set up nodes
         synth_tool = "yosys"
-        pnr_tool = flow_options.get("pnr", "vpr")
-        if not pnr_tool in ["vpr", "nextpnr"]:
-            raise RuntimeError(f"F4PGA flow error: invalid P&R tool: {pnr_tool}")
+        self.pnr_tool = flow_options.get("pnr", "vpr")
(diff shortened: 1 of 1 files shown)


olofk:
Yes! Thanks for this. Looks good

PR #345, opened 2022-10-01
Add modelsim option prerun_tcl

ttobsen:
This PR adds a new option for Modelsim which helps to execute Tcl scripts before simulation.

A very important use-cae is to disable the IEEE warnings. In Modelsim, this can either be done using the modelsim.ini or via Tcl commands. This Tcl commands have to be run right before simulation and not while compiling. This PR adds an option to specify a Tcl script which will be executed right before simulation starts.

PR #354, opened 2022-10-31
Change Modelsim options order

ttobsen:
This is the followup PR from discussion in #345.

The place of the Modelsim EXTRA_OPTIONS has changed, so configuration scripts can be added before simulaton is running.


olofk:
Perfect. Thanks

PR #355, opened 2022-11-07
tool openlanes missing file_type systemVerilogSource

shuckc:
In common with other tools, openlanes backend should be passed both `verilogSource` and `systemVerilogSource` files.


olofk:
Oops. Thanks for the fix

PR #357, opened 2022-11-10
Add questa formal

ttobsen:
Add Siemens Questa Formal tooling.


olofk:
Thanks! I would like to mention that we are moving all tools to a new API called the flow API and ideally we would like to see a tool class in edalize/tools that also implements support for this tool as well. But we can start by pulling this in

PR #358, opened 2022-11-21
Vunit dependencies

nopeslide:
added dependency information for vunit

- added fields to edatool File class
  - "fileset": fileset which included the src file
  - "core": name of the core which included the src file
  - "depend": name list of core dependencies
- added core dependency and core file map to vunit jinja env
  - "depend": mapping of core name to core name of dependencies
  - "core_files": mapping of core name to src file objects
- reworked vunit template
  - save return value of added files added to libraries
  - generate mapping of core names to vunit file objects
  - specify core dependencies as dependencies between core files

### DIFF
--- a/edalize/edatool.py
+++ b/edalize/edatool.py
@@ -400,10 +400,17 @@ class Edatool(object):
 
     def _get_fileset_files(self, force_slash=False):
         class File:
-            def __init__(self, name, file_type, logical_name):
+            def __init__(
+                self,
+                name,
+                file_type,
+                logical_name,
+                core=None,
+            ):
                 self.name = name
                 self.file_type = file_type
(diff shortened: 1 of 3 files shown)


nopeslide:
depends on
olofk/fusesoc#593

nopeslide:
WIP because I do not know if this is the intended way to add such a feature

olofk:
I merged this, but I'm having issues. It doesn't pass the test suite anymore. It looks like there's a hard requirement that the "dependencies" fields exist. I'm afraid we need to make sure it works even without this, because not all producers of EDAM files will set this

olofk:
Did some attempts to fix this, but I don't understand the code well enough, so I'm removing this fix for now until we can figure out the proper fix

PR #242, opened 2021-05-20
Adding desing compiler support

timsaxe:
A first pass at adding design compiler support.  This is a fairly simple shell that sets up a few variables, creates a tcl script to analyze the sources, and passes off control to a tcl script that will do the heavy lifting.  I have it running on the core-v mcu from open hardware, so I know that it is usable.

PR #359, opened 2022-12-05
add support for Synopsys DC and Cadence Genus

davideschiavone:


davideschiavone:
Hi @olofk, this PR replaces this one https://github.com/olofk/edalize/pull/242
and it adds support for Synopsys Design Compiler.

The FuseSoc file we used is for example this one: https://github.com/esl-epfl/x-heep/blob/main/core-v-mini-mcu.core#L339

We recently made a chip using FuseSoc and the edalize extensions for Design Compiler. 

Let me know what's missing to merge it in

edalize/design_compiler.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @classmethod
+    def get_doc(cls, api_ver):
+        if api_ver == 0:
+            return {'description' : "The design_compiler backend executes Synopsys design_copiler to build a gate-level netlist",
+                    'members' : [
+                        {'name' : 'script_dir',
+                         'type' : 'String',
+                         'desc' : 'Path to Syopsys scripts (e.g. /home/user/project/synopsys/scripts)'},
+                         {'name' : 'dc_script',
```
Jbalkind, line 37: Perhaps this would be better as either a tclSource file or a "new filetype" along with the main sources?

davideschiavone: you mean replacing "String" with tclFile?

shareefj: No, he means using the tclSource file type in the core file and then copying any accompanying files over.  
 E.g.: https://github.com/lowRISC/opentitan/blob/master/hw/top_earlgrey/chip_earlgrey_cw310.core

edalize/generic.py
```diff
@@ -0,0 +1,270 @@
...
+A core (usually the system core) can add the following files:
+
+- Standard design sources (Verilog only)
+
+- Constraints: unmanaged constraints with file_type SDC, pin_constraints with file_type PCF and placement constraints with file_type xdc
+
+"""
+
+
+class Generic(Edatool):
```
shareefj, line 28: This probably shouldn't be called Generic?  What tool flow is it executing?  And I think @olofk has got a claim on the Generic name already:  https://github.com/olofk/edalize/compare/master...generic_tool  (if only he'd merge it so it can be used ;-) )

edalize/design_compiler.py
davideschiavone: correct me if I am wrong, but wouldn't this assume that there is a kind of ordering of calls to those scripts?

calling a "root" script like it is now, allows people to define their flow in that script, and get rid of the boring, lengthy, and error-prone "analyze" script to compile all the files. This one is very well done by FuseSoc.

But making a synthesis flow one-fit-all does not sound very versatile to me.

But you reminded me that actually, it makes sense to support the tcl files by sourcing them in the analyze script.

Of course, if you have suggestions I would be happy to work on them.

As today the Makefile only calls the "root" script, and then from that script you call all the rest, that's why the TCL file was not actually included in the analyze script. But why not. Let me do it for completeness reasons.

davideschiavone: done in next commit @shareefj @Jbalkind

edalize/generic.py
davideschiavone: actually, I haven't realized that this file was here... as you see it's a cross-people contribution and I did not have everything under control :D sorry, removing it

davideschiavone:
changed the title of the PR as now we also added Cadence Genus support. I will take care of the CI ASAP

edalize/templates/design_compiler/design-compiler-read-sources.tcl.j2
```diff
@@ -0,0 +1,9 @@
+# Auto-generated project tcl file to read all sources
+
+{% if incdirs -%}
+set search_path [list {{ incdirs|join(' ') }}]
```
Semora, line 4: I use fusesoc and I used this tool to generate DC analyze scripts. I need to add timing libraries to search path, but I don't want to mention those in Core file. (Timing libraries path may vary for different servers or users)
However, this line overwrites search_path and I can't set search_path in my scripts. (I can but it won't work :sweat_smile: )
I suggest to concatenate incdirs with previous value of search_path: 
```suggestion
set search_path [concat [list {{ incdirs|join(' ') }}] $search_path]
```
(I did test this solution and it worked for me)

olofk: That sounds like a good idea. @davideschiavone, can you change that?

olofk:
I'm really happy to see this work. There are some things I would like to see changed, but I think we can pull this in for now and fix those things when we port it from the legacy tool API to the new flow API instead. I will amend the patches slightly to remove the permissions changes to the files and then split them up into a DC and a Genus commit. Hope that is ok with everyone.

olofk:
I cleaned up the commits and fixed the formatting now so that it passes the CI. I don't think I have introduced any functional changes, but I'm not entirely sure anymore. So before I push this to the main branch, could someone (@davideschiavone perhaps?) take a look at the https://github.com/olofk/edalize/tree/dcgenus branch, because that is what I am planning to push.

davideschiavone:
@olofk done :)

Semora:
> I cleaned up the commits and fixed the formatting now so that it passes the CI. I don't think I have introduced any functional changes, but I'm not entirely sure anymore. So before I push this to the main branch, could someone (@davideschiavone perhaps?) take a look at the https://github.com/olofk/edalize/tree/dcgenus branch, because that is what I am planning to push.

@olofk 
As @davideschiavone approved https://github.com/olofk/edalize/pull/359#discussion_r1099748413 please apply this change to https://github.com/olofk/edalize/tree/dcgenus too.
Tnx

davideschiavone:
@olofk - I should have fix the format

PR #362, opened 2022-12-29
[fix] fix the build command for symbiyosys

### LINKED ISSUE #257: SymbiYosys Backend Doesn't Support Running All Tasks
The documentation in the SymbiYosys backend indicates that all tasks should be run if there's no setting for `tasknames`:
```
   symbiyosys:
     tasknames:
       # A list of task names to pass to sby. Defaults to empty (in which case
       # sby will run each task in the .sby file)
       - my_proof
```
However, leaving this unset produces the following `sby` error:
```
ERROR: Exactly one task is required when workdir is specified.
```
It appears that for this to work the `-d` option for `sby` should be removed in favour of using the default output directories.

Henkru:
The current `run` step of the symbiyosys backend generates a `sby` command which includes the hardcoded `-d build` argument. However, this command line argument is incompatible with multiple tasks (see #257).

If the `-d build` argument is specified, the output folder of sby is forced to `build/<core>/<target>-symbiyosys/build`. Otherwise, the following path is used: `build/<core>/<target>-symbiyosys/test_<taskname>`.

This fix removes the hardcoded `-d` argument and introduces a new `extra_options` tool option for the symbiyosys backend, which can be used to pass additional arguments for the `sby` command (e.g. `-d build`, if necessary).

The example usage of the `extra_options`:
```
   symbiyosys:
     tasknames:
       # A list of task names to pass to sby. Defaults to empty (in which case
       # sby will run each task in the .sby file)
       - my_proof
     extra_options:
       # A list of extra command line arguments to sby. Defaults to empty.
       - -d
       - build
```

Closes #257.

### DIFF
--- a/edalize/symbiyosys.py
+++ b/edalize/symbiyosys.py
@@ -25,6 +25,10 @@ Example snippet in CAPI2 format:
        # A list of task names to pass to sby. Defaults to empty (in which case
        # sby will run each task in the .sby file)
        - my_proof
+     extra_options:
+       # A list of extra command line arguments to sby. Defaults to empty.
+       - -d
+       - build
 
 The SymbiYosys tool expects a configuration file telling it what to do. This
 file includes a list of all the source files, together with various flags which
@@ -108,8 +112,10 @@ You can reproduce the example above with something like
         "lists": {
             # A list of tasks to run from the .sby file. Passed on the sby
             # command line.
-            "tasknames": "String"
-        }
+            "tasknames": "String",
+            # A list of extra arguments to the sby command.
+            "extra_options": "String",
+        },
     }
 
     def __init__(self, edam=None, work_root=None, eda_api=None, verbose=True):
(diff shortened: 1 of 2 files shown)


edalize/symbiyosys.py
```diff
@@ -108,8 +112,10 @@ class Symbiyosys(Edatool):
         "lists": {
             # A list of tasks to run from the .sby file. Passed on the sby
             # command line.
-            "tasknames": "String"
-        }
+            "tasknames": "String",
```
Henkru, line 115: Is this `tool_options` some legacy stuff? Other edatool classes are using `TOOL_OPTIONS` with a different structure.

olofk:
Thanks. This sounds good to me. Hard coded parameters should be avoided as far as possible. I would like to rename `extra_options` to `sby_options` for symmetry with other tools.

edalize/symbiyosys.py
olofk: This backend is part of the legacy tool API. The classes under the `tools` and `flows` directories belong to the new flow API. Unfortunately most tools haven't been ported over to the new API yet.

olofk:
At this point I would prefer dropping this and focus on migrating the sby backend to the Flow API instead.

PR #364, opened 2022-12-30
[fix] yosys: use default target name for Makefile

### LINKED ISSUE #363: yosys: wrong target name in Makefile, if output_name tool option is used
The `yosys` edatool does not add the custom output name to Makefile when the `output_name` tool option is defined. For example the following options

```yaml
yosys:
  arch: ice40
  output_name: test.json
```
generates the following `Makefile` where the target contains the wrong name:
```makefile
#Auto generated by Edalize

all: test_yosys_0.blif

test_yosys_0.blif: edalize_yosys_template.tcl
        $(EDALIZE_LAUNCHER) yosys -l yosys.log -p 'tcl edalize_yosys_template.tcl'
```

Henkru:
This fix the issue where `f"{self.name}.{output_format}"` is always used as a target name in the Makefile even if the custom output name is defined in the tool options.

Closes #363


olofk:
This looks perfectly fine. Thank you for your contribution. Picked and pushed

PR #366, opened 2023-01-04
Added support for sandpiper-saas, a TL-Verilog Interpreter

shariethernet:
[Sandpiper-saas](https://pypi.org/project/sandpiper-saas/) Edition runs [Redwood EDA](https://redwoodeda.com/)'s [SandPiper](https://redwoodeda.com/products)™ [TL-Verilog](https://tl-x.org/) compiler as a microservice in the cloud to support low-overhead and zero-cost open-source development using commercial-grade capabilities.

This pull request adds support for using sandpiper-saas within Edalize. Coupled with integrations with other tools using flow API, this would allow Edalize to be used for TL-Verilog-based Digital Design/Verification Flows.

A simple example of using sandpiper-saas within Edalize is provided in this [repo](https://github.com/shariethernet/edalize_sandpiper_example)


olofk:
I'm pulling this in as it is, but it would be great if we could also port it to the new flow API. That would make it much more useful since we can then easily hook it up as a frontend to any flow

PR #375, opened 2023-03-30
Fix Quartus version detection for optional patches

patrickrst:
Hi! I added an optional group in the version detection regex for when a patch has been installed on top of Quartus.

Example version string: `Version 21.4.0 Build 67 12/06/2021 Patches 0.03 SC Pro Edition`

And some more examples of patch version: https://www.intel.com/content/www/us/en/docs/programmable/683706/22-4/software-patches-included-in-this-release.html

### DIFF
--- a/edalize/quartus.py
+++ b/edalize/quartus.py
@@ -94,23 +94,28 @@ class Quartus(Edatool):
             "edition": "Standard",
         }
         try:
-            qsh_text = subprocess.Popen(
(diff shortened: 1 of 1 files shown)


olofk:
Patch is fine. I will pull it in, but after this there will be no more changes to this version of the quartus backend. We should instead focus on porting it to the new Flow API.

olofk:
Sorry. Backing this out as it causes some weird issues with the regression tests

olofk:
And now I can't figure out how to reopen this issue...

Issue #344, opened 2022-09-29
vivado xsim support for xci or xilinx ip


skokvermon:
Hi, 
it seems that vivado xsim doesn't support generating xilinx ip for simulation. 
It is supported for synthesis

Is it a planned feature?
If not I would be willing to help implementing it.

Jbalkind:
I think we'd all love to see that if you're willing to give it a go :)

olofk:
Yes, that is a known issue. Currently, the vivado backend runs in project mode while xsim runs in non-project mode. My proposal is to convert the xsim backend to run in project mode as well. That should bring a number of other features as well. Haven't got the bandwidth to work on it myself so very happy for any help here.

PR #377, opened 2023-04-04
vivado xsim xci compatibility. compatible with vivado synthesis.

skokvermon:
This adds possibility of simulating vivado XCI in xsim.
Compatible with vivado synthesis XCI support.
Automatically calls vivado to generate code from the makefile

needs a new parameter: part which select xilinx part number

Also added a side feature in the makefile to enable vcd output.
the default makefile run is now selectable with new parameter: default_run

fixes bug #344

example

    tools:
      xsim:
        default_run: run-vcd
        part: xc7a100tcsg324-2


olofk:
I think this might be better to handle with a generator so that all simulators can benefit from this. Also, the xsim backend currently only implements the old Tool API. We should focus on porting it to the new Flow API.

PR #380, opened 2023-04-12
-quiet option removed from vsim call

pawelkudlakaldec:
-quiet option is invalid for standalone vsim command in Riviera-PRO


olofk:
Ok, this is a very simple patch so I will pull it in after the test case reference files are also fixed.

But generally, I will not allow any more changes to the backends that implement the old Tool API. We should instead focus on porting this backend to the new Flow API.

danielmlynek:
1 Will this PR be merged/closed?
2 You mean that riviera backend is not implemented in new Flow API? Do we need to implement it is there any reference - ie new api implemented for other backend that we can refer to.

olofk:
This is fixed now with a8d015b06ee7ef8500441c6d67e184e0d480330f

PR #383, opened 2023-05-18
removed absolute path to binary when calling `touch`

nopeslide:
absolute paths should not be used because they are not portable


olofk:
Whoops! Thanks. Fixed

PR #384, opened 2023-05-18
Vunit dependencies

### LINKED ISSUE #350: explicit dependencies for vunit
vunit can't resolve dependencies when using a vhdl/verilog language mix. 
Such dependencies are expressed by specifying them manually via `A.add_dependency_on(B)`.
I propose to add the dependencies between the edalize filesets as explicit dependencies in the generated `run.py`.

nopeslide:
I mitigated the problem of #358 by adding a few checks if the fields are present.
Will have to do further testing with an actual codebase, but I think this should do it.

fixes #350

### DIFF
--- a/edalize/edatool.py
+++ b/edalize/edatool.py
@@ -403,10 +403,17 @@ class Edatool(object):
 
     def _get_fileset_files(self, force_slash=False):
         class File:
-            def __init__(self, name, file_type, logical_name):
+            def __init__(
+                self,
+                name,
+                file_type,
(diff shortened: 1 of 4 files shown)


nopeslide:
Checked it against multiple of my projects and it looks good,
If there are any other tests I can run this against, please let me know.

olofk:
Getting back to this took a fair bit longer than I had hoped for, but it looks fine now. Thank you for your contributions. Picked and pushed!

PR #385, opened 2023-05-30
Libero Project TCL Updates

paul-gatewood:
I've been working on incorporating FuseSoC/Edalize into an existing project, and I made a few updates to the Libero project template to get it to build.

1. Add include directories to Libero project
2. Set root after running user TCL scripts (this allows a SmartDesign to be used as the root)
3. Skip organize_tool_files when no constraints are present (avoids a Libero error since -file is required)

### DIFF
--- a/edalize/templates/libero/libero-project.tcl.j2
+++ b/edalize/templates/libero/libero-project.tcl.j2
@@ -13,10 +13,9 @@ build_design_hierarchy
 {% endif %}
 
 # Import HDL sources and constraints
-import_files \
-        {% for src_file in src_files if src_file|src_file_filter%}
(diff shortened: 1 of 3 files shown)


olofk:
Thank you for this. The added functionality looks fine but I see no updates of the tests. Did you check that it still passes the test suite? (Running tox from the repo root should do the trick otherwise. Setting the environment variable `GOLDEN_RUN` will automatically update the reference files)

paul-gatewood:
Ah, I didn't try running the tests before. They should be passing with the latest commit.

olofk:
Yes! This looks good now. Thank you for your contributions. Picked and pushed!

PR #386, opened 2023-06-18
Updated Sandpiper-SaaS to support the Flow API and to be used as a frontend

shariethernet:
This update adds support for Sandpiper-SaaS using the Flow API. This allows sandpiper-saas to be used as a "frontend" to convert TL-Verilog to SystemVerilog/Verilog for any Flow.

An updated example of how to use sandpiper-saas with Edalize and Fusesoc in various contexts (standalone tool, frontend to vivado, in a custom flow) is provided [here](https://github.com/shariethernet/edalize_sandpiper_example)


olofk:
Very nice! And fantastic to see an example on how to use this in different modes. Well done. Picked and pushed

PR #387, opened 2023-07-18
Add support for Libero netlist constraints

paul-gatewood:
Libero has two types of netlist constraints files that are synthesis only, so they didn't work with the existing setup. Adding specific filters for synthesis and place/route simplified the changes needed to the template. I also added GOLDEN_RUN to be passed through to tox.


paul-gatewood:
Splitting this out into different requests.

PR #389, opened 2023-08-07
Added support for additional Verilator modes

sifferman:
Hello! I was having troubles getting [Verilator XML dumps](https://veripool.org/guide/latest/exe_verilator.html#cmdoption-xml-only) working with the existing Edalize version, so I added it as a verilator->mode option.

I'm not sure if better documentation should be provided for it, but I didn't want to be too verbose

Thanks!


sifferman:
Actually, now that I think about it, I should probably just add all the remaining modes:

Existing:

- --cc
- --sc
- --lint-only

New:

- [ ] --binary
- [ ] --dpi-hdr-only
- [X] --xml-only
- [ ] --E

sifferman:
Should be all ready for review!

This adds the following modes:
```plain
binary (--binary)
dpi-hdr-only (--dpi-hdr-only)
preprocess (-E)
xml-only (--xml-only)
```

I also removed `--exe` from the .vc file when in the modes "binary", "dpi-hdr-only", "lint-only", "preprocess", or "xml-only".

sifferman:
Actually, I just had a new idea for 3 additional Verilator fields:

* gen-xml: \<bool\>
* gen-dpi-hdr: \<bool\>
* xml-output: \<filename : String\>

This way, using the same run command on a single target, we could generate an XML file, and a DPI.h file.

If gen-xml was enabled, verilator could just be run once more, except with the option --xml-only. If gen-dpi-hdr was enabled, verilator could just be run once more, except with the option --dpi-hdr-only.

This could even potentially replace the dpi-hdr-only and xml-only modes. -E only prints to stdout, so I don't think this deserves its own field.

sifferman:
At least one of the following flags must be set [[ref](https://github.com/verilator/verilator/blob/cbdee5a80499aaae049e955b7b45d5d8abc5b8ef/src/V3Options.cpp#L799-L803)], and the documentation implies them to be mutually exclusive [[ref](https://veripool.org/guide/latest/exe_verilator.html)]:

`--binary`, `--cc`, `--sc`, `--dpi-hdr-only`, `--lint-only`, `--xml-only`, `-E`

Selecting more than one leads to unexpected behavior.

### Example
core snippet
```yaml
      verilator:
        verilator_options:
          - --xml-only
```
error
```plain
make[1]: Vtop.mk: No such file or directory
make[1]: *** No rule to make target 'Vtop.mk'.  Stop.
```
fixed core snippet
```yaml
      verilator:
        mode: lint-only
        verilator_options:
          - --xml-only
```
With FuseSoC, to get XML output, you must specify `verilator_options: --xml-only` and `mode: lint-only`, which seems weird. Plus there is no current way to get DPI header output because `--lint-only` overwrites `--dpi-hdr-only`, and `--dpi-hdr-only` + `--cc`/`--sc` is not allowed.

Therefore, I feel like creating a mode for each makes a lot of sense.

sifferman:
Small update:

Wilson just merged my pull request that explicitly enforces them to be mutually exclusive (verilator/verilator#4409)

Now the `--xml-only` example I gave no longer works:

> ```yaml
>       verilator:
>         mode: lint-only
>         verilator_options:
>           - --xml-only
> ```

due to

```plain
%Error: The following cannot be used together: --lint-only, --xml-only. Suggest see manual
```

sifferman:
Alright, @olofk. How does this look?

I expanded the Verilator Makefile to include the modes as callable Makefile targets:

https://github.com/olofk/edalize/blob/2e30ad1726622a6674e723cdbc32397f063c62b3/edalize/verilator.py#L37-L47

I also added the following fields `gen-xml`, `gen-dpi-hdr`, `gen-preprocess`, and have them run the corresponding Makefile targets after the usual building due to the specified mode.

https://github.com/olofk/edalize/blob/2e30ad1726622a6674e723cdbc32397f063c62b3/edalize/verilator.py#L78-L92

https://github.com/olofk/edalize/blob/2e30ad1726622a6674e723cdbc32397f063c62b3/edalize/verilator.py#L251-L257

I also added an additional mode "none" which when specified, will only do the additional builds specified with `gen-xml`, `gen-dpi-hdr`, `gen-preprocess`

https://github.com/olofk/edalize/blob/2e30ad1726622a6674e723cdbc32397f063c62b3/edalize/verilator.py#L247-L249

olofk:
Took me a while longer than planned to get back to this, but I think it looks all fine now. Thank you for your contributions. Picked and pushed!

sifferman:
Thanks, @olofk!

PR #392, opened 2023-08-24
Add a --pgm tool option to stop build stage from automatically programming

GCHQDeveloper147:
Add a `--pgm` argument to separate place & route from programming in build stages using relevant tools (Vivado, Quartus, ISE). By default these won't attempt to program a device (which can quite easily crash servers unless a number of other options are also configured), but setting `--pgm=<applicable tool>` will allow it to do so.


olofk:
This looks fine to me. I always envisioned a pgm switch as a bool, but this makes it easier to bring in alternative programming methods if needed. Thanks!

PR #394, opened 2023-10-02
diamond: Add thermal_default_clock_frequency tool option

markfeathers:
This is required to specify on some parts or synthesis fails with:

ERROR: Clock frequency cannot be 0. Please set Default Clock Frequency in thermal analysis of Strategy.


olofk:
Can this be solved instead by adding the needed line to a TCL file and include it with `file_type : tclSource` ? We generally try to avoid adding things to tool_options if they can be solved in other ways. Otherwise we would end up with too many tool options

markfeathers:
This works for me. At the time of writing I wasn't aware I could include tcl like that. Thanks!

PR #398, opened 2023-10-06
Libero Updates

paul-gatewood:
Adds support for Libero netlist constraints and for --no-export with Libero. Also fixes some issues I was having with the PR checks.

### DIFF
--- a/edalize/libero.py
+++ b/edalize/libero.py
@@ -95,6 +95,12 @@ class Libero(Edatool):
         self._check_mandatory_options()
         (src_files, incdirs) = self._get_fileset_files(force_slash=True)
         self.jinja_env.filters["src_file_filter"] = self.src_file_filter
+        self.jinja_env.filters[
+            "syn_constraint_file_filter"
+        ] = self.syn_constraint_file_filter
+        self.jinja_env.filters[
+            "pnr_constraint_file_filter"
+        ] = self.pnr_constraint_file_filter
         self.jinja_env.filters["constraint_file_filter"] = self.constraint_file_filter
         self.jinja_env.filters["tcl_file_filter"] = self.tcl_file_filter
 
@@ -168,6 +174,8 @@ class Libero(Edatool):
             "PDC": "-io_pdc {",
             "SDC": "-sdc {",
             "FPPDC": "-fp_pdc {",
+            "FDC": "-net_fdc {",
+            "NDC": "-ndc {",
         }
         _file_type = f.file_type.split("-")[0]
         if _file_type in file_types:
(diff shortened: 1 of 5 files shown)


olofk:
Thank you for this. I have some questions about the changes and the libero backend in general. I did not write that one myself, and looking at it now I'm a bit confused about certain things.

But I will start right away with pulling in your two first commits as I'm seeing the CI failures as well, that I believe these are aiming to fix.

olofk:
First two patches to fix CI are puled in now. Thanks for those!

So, the first thing I wonder now is if there is a compelling reason to ever use `import_files` instead of `create_links`? I understand that `import_files' is what has been used up until now but I'm not convinced it was done for the right reasons. Then again, I'm not a Libero user myself, so I can't know for sure. But looking at the code I believe this could simplify things a bit. What do you think?

paul-gatewood:
Of course, I'm glad I can contribute! That's a great point, I can't actually think of a use case for `import_files` that wouldn't be covered by edalize's manual export to the src directory (or maybe that's handled by fusesoc). I ran through some tests on a project of mine, and it looks like that approach works great.

edalize/libero.py
```diff
@@ -186,19 +194,39 @@ def tcl_file_filter(self, f):
             return file_types[_file_type] + f.name
         return ""
 
-    def constraint_file_filter(self, f, type="ALL"):
+    def syn_constraint_file_filter(self, f):
+        file_types = {
```
olofk, line 198: I don't really understand what this dict does. Could you explain?

edalize/libero.py
```diff
@@ -186,19 +194,39 @@ def tcl_file_filter(self, f):
...
+            "SDC": "constraint/",
+            "FDC": "constraint/",
+            "NDC": "constraint/",
+        }
+        _file_type = f.file_type.split("-")[0]
```
olofk, line 203: What is this split for?

edalize/libero.py
```diff
@@ -186,19 +194,39 @@ def tcl_file_filter(self, f):
...
         if _file_type in file_types:
-            filename = f.name.split("/")[-1]
-            if type == "ALL":
-                return file_types[_file_type] + filename
-            elif _file_type == type:
-                return file_types[_file_type] + filename
+            return f.name
+        return ""
+
+    def constraint_file_filter(self, f, type="ALL"):
```
olofk, line 219: Do we still need this function now that we have the syn and pnr versions?

olofk:
Thanks. Happy to see the simplifications after dropping the export parts (and to answer your other question, it's technically FuseSoC that does the exporting). In general it looks good, but I made some comments where I think we can actually make further simplifcations.

edalize/libero.py
paul-gatewood: From what I can tell, it might just be used to convert `file_type`s like [`vhdlSource-2008`](https://github.com/olofk/edalize/blob/main/edalize/quartus.py#L139) to just `vhdlSource`. It seems like that pattern's used in a few other toolchains, like [quartus](https://github.com/olofk/edalize/blob/main/edalize/quartus.py#L139).

olofk: That would be my guess too, but since none of the potential matches contain dashes in the file type I suspect this whole function (as well as the other ones) can just be simplified to:
```python
    def pnr_constraint_file_filter(self, f):
        if _file_type in ["PDC", "SDC", "FPPDC"]:
            return f.name
```

edalize/libero.py
paul-gatewood: Oh, that dict used to supply the path to the constraints files when they were imported, but it's obsolete now that we're always linking. I'll take a stab at cleaning that up.

edalize/libero.py
paul-gatewood: It was still used in one place for the timing constraints, but I replaced that with a simpler filter and removed this one.

olofk:
Many thanks for your contributions. Squashed and merged!

PR #400, opened 2023-10-13
Update setup.py to use Git versioning like FuseSoC

paul-gatewood:
This should update Edalize to generate its version from the tags like FuseSoC does (I pretty much just copied and pasted it).


olofk:
Sounds like a plan! Thanks. Picked and pushed

shareefj:
This PR caused me some head scratching tonight so I thought I'd post it here to notify others who may come searching.

A common workflow for creating Edalize PRs has been to:

1. Identify bug/feature
2. Fork repo
3. Create branch 
4. Fix bug or add feature
5. Push branch
6. Open PR against upstream
7. Update local requirements.txt to pull in forked branch until PR is merged
8. Revert to using vanilla Edalize once PR is merged.

However, this PR's change to use Setuptools SCM for tag management means that a forked repo defaults to a tag of 0.1 which fails the dependency checks when pip installing.  Fusesoc 2.3 requires Edalize of >= 0.4.1.

The solution is to tag your branch with the same tag that you created it from.  Unless there's some nicer way to force a tag from a requirements.txt file?  Or some other method I'm not familiar with?

Before I tagged my feature branch, the following requirements.txt file would end up installing Fusesoc version 1.8.x.

```
fusesoc
git+ssh://git@github.com/shareefj/edalize.git@feature/allow_edif_submodules#egg=edalize
```

PR #401, opened 2023-10-13
Common get_file_type function

paul-gatewood:
Since the same pattern was used in several places to remove the version from the file_type, I made a common function in `utils.py` and added a comment about its use.


olofk:
Yeah, that's better. Thanks! Hadn't seen that enumerate trick before. Took a moment to realize how it worked :)

PR #407, opened 2023-12-06
Fix efinity tool

louiecaulfield:
The top level of the efinity tool is missing, making fusesoc not find the tool.

This fixes it.

### DIFF
--- a/edalize/efinity.py
+++ b/edalize/efinity.py
@@ -1,0 +1,67 @@
+# Copyright edalize contributors
+# Licensed under the 2-Clause BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-2-Clause
+
+import logging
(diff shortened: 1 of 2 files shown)


olofk:
This is actually by design. Efinity is only implementing the new Flow API rather than the old Tool API. There are a number of benefits to this new API and I have been writing a bit about it e.g. [here](https://blog.award-winning.me/2022/01/edalize-030.html) but I'm aware it could have been better communicated.

Basically, what you need to do is to add a key `flow : efinity` to your target, and then specify all your tool- and flow-specific options under the `flow_options` key. Adding the `flow` key will tell FuseSoC to use the new flow API which ignores the `default_tool` and `tool_options` keys

olofk:
Closing this.

PR #409, opened 2024-01-09
Add cocotb verilator flow_tool option

isomoye:
This PR adds verilator as a cocotb flow_tool simulation option by adding the cocotb "extend an existing build flow" options in sim.py. It also adds a switch to the verilator makefile call to use "Vtop.mk" instead of "V 'cocotb-module'.mk".

A cocotb simulation running on verilator can then be run using just the following fusesoc simulation configuration

  ```default:
    flow: sim
    flow_options:
      tool : verilator
      cocotb_module : test_file
      timescale: 1ns/1ns
    filesets : [rtl,cocotb]
    toplevel : [toplevel]```

### DIFF
--- a/edalize/flows/sim.py
+++ b/edalize/flows/sim.py
@@ -42,7 +42,7 @@ class Sim(Generic):
                     ["--vpi",
                      "--public-flat-rw --prefix Vtop",
                      "-LDFLAGS \"-Wl,-rpath,`cocotb-config --lib-dir` -L`cocotb-config --lib-dir` -lcocotbvpi_verilator -lgpi -lcocotb -lgpilog -lcocotbutils\""
-                    ],
+                     ],
                 ),
             }
             (opt, val) = cocotb_options[tool]
--- a/edalize/flows/sim.py
+++ b/edalize/flows/sim.py
@@ -43,8 +43,7 @@ class Sim(Generic):
                 "verilator": (
                     "verilator_options",
                     ["--vpi",
-                     "--top top",
-                     "--public-flat-rw",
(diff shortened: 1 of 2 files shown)


edalize/flows/sim.py
```diff
@@ -23,6 +24,8 @@ class Sim(Generic):
     def configure_tools(self, flow):
         if self.flow_options.get("cocotb_module"):
             tool = self.flow_options.get("tool")
+            share_dir = os.popen("cocotb-config --share").read().rstrip() \
```
olofk, line 27: I wouild prefer evaluating `cocotb-config --share` directly like we do with `cocotb-config --libdir`. That makes the script more portable if we want to run it on a different machine than it was created on.

olofk:
Many thanks. A very welcome addition and I'm hoping to find some time to try it myself. I just had a smaller comment above. Happy to merge if you could fix that.

edalize/flows/sim.py
isomoye: fixed by changing the share directory evaluation to a direct call in tools/verilator.py rather than a system call. see commit https://github.com/olofk/edalize/pull/409/commits/aa7d8c31b19fb11f7f73b4ecf312dab157153e7c.

isomoye:
I fixed the format issue by following your suggestions. I agree with your assessment of using flow_options in the tool specific flow and I'm open to working on a permanent fix soon. I tested the verilator.py changes without flow_options and it looks like your suggested change fixed the crashing issue.

olofk:
There was still some formatting issues in sim.py, but I took the liberty to fix that locally. Squashed and pushed! Thank you for your contributions!

PR #412, opened 2024-01-22
Improve GHDL Tool

psi-chuffine:
This works with my simple builds, but needs additional testing and validation for libraries and more complex GHDL usage.
I will correct any problems as I encounter them.

### DIFF
--- a/edalize/tools/ghdl.py
+++ b/edalize/tools/ghdl.py
@@ -111,9 +111,9 @@ class Ghdl(Edatool):
                 # Files without a specified library will by added to
                 # libraries[None] which is perhaps poor form but avoids
                 # conflicts with user generated names
-                libraries[f["logical_name"]] = libraries.get(f["logical_name"], []) + [
-                    f["name"]
-                ]
+                libraries[f.get("logical_name", None)] = libraries.get(
+                    f.get("logical_name", None), []
+                ) + [f["name"]]
                 vhdl_sources += " {file}".format(file=f["name"])
                 depfiles.append(f)
             else:
--- a/edalize/tools/ghdl.py
+++ b/edalize/tools/ghdl.py
@@ -24,6 +24,8 @@ class Ghdl(Edatool):
     def setup(self, edam):
         super().setup(edam)
         analyze_options = self.tool_options.get("analyze_options", [])
+        run_options = self.tool_options.get("run_options", [])
+        sim_options = self.tool_options.get("sim_options", [])
 
         # Check of std=xx analyze option, this overyides the dynamic determination of vhdl standard
         import re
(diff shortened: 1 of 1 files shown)


psi-chuffine:
Apologies for the overbroad PR, I should have branched the cocotb addition.

olofk:
There seems to be issues still I tried it with an old VHDL example that I put together for other reasons a couple of years ago. Attaching the example project if that is of any use for debugging.

Run `fusesoc run --target=simold libaxis` for the old Tool API or `fusesoc run --target=sim libaxis` for the new Flow API that you are working on.

[vhdl_example.tar.gz](https://github.com/olofk/edalize/files/14052653/vhdl_example.tar.gz)

psi-chuffine:
I'm concatenating GHDL calls incorrectly. I should have a fix shortly.

psi-chuffine:
I have it building, almost, with the test set you sent.
I'm currently getting 
``` src/libaxis_0/libaxis/bench/vhdl/suv/tb_axis_sync_fifo.vhd:56:19:error: type of a shared variable must be a protected type
  shared variable words : t_mem(0 to TRANSACTIONS-1)(WIDTH-1 downto 0);
                  ^
src/libaxis_0/libaxis/bench/vhdl/suv/tb_axis_sync_fifo.vhd:56:19:note: (you can use -frelaxed to turn this error into a warning)
  shared variable words : t_mem(0 to TRANSACTIONS-1)(WIDTH-1 downto 0);
```
This is coming from `analyze_options`. The old GHDL tool picks up `analyze_options` from `libaxis_0.eda.yml`, but the new tool doesn't. Why would that be the case?
I'm also assuming that each command in the flow api philosophy should be one action, so I've broken the library imports into independent stages.

alexfanqi:
> Why would that be the case?

I think you need to add these options to TOOL_OPTIONS. tool_options are filtered through it.

alexfanqi:
> This is coming from analyze_options
I think you need to add those into TOOL_OPTIONS, tool_options are filtered through it.

alexfanqi:
Does it make sense to also support having top level entity in a library other than `work`? Right now, I get `No rule to make target 'work-obj08.cf', needed by 'run'`.

psi-chuffine:
I should be able to make that change, I set the flow to require ```f"work-obj{standard}.cf"``` to ensure the build order is handled correctly as I have always used work as my top level.

psi-chuffine:
I've still got an action here. I'll post an update when I have made the changes.

flaminggoat:
@olofk @psi-chuffine I made some changes following on from this PR to hopefully complete this feature. https://github.com/olofk/edalize/pull/462

olofk:
Many thanks. I pushed this as part of #462

PR #413, opened 2024-01-24
Add Vivado power optimization option flag

sebinho:
This proposal adds a flag to enable power optimization during implementation stage of a Vivado run.
Power optimization can have huge impacts, especially when using a device that has a PS (Processing System).

To use this new option, simply add the following flag in your `*.core` file:
```bash
flow_options :
            part           : xczu3eg-sfvc784-1-e
            power_opt      : True
```


sebinho:
It works well for me when using the Vivado FLOW. Maybe something is off when using the older tool scheme

sebinho:
I will create a new pull request so will close this one

PR #414, opened 2024-01-24
Add Vivado power optimization option flag

sebinho:
This proposal adds a flag to enable power optimization during implementation stage of a Vivado run.
Power optimization can have huge impacts, especially when using a device that has a PS (Processing System).

To use this new option, simply add the following flag in your `*.core` file:
```bash
flow_options :
            part           : xczu3eg-sfvc784-1-e
            power_opt      : 1
```

If set to 0, it will disable the power optimizations (and also if the options is not specified at all)

### DIFF
--- a/edalize/templates/vivado/vivado-run.tcl.j2
+++ b/edalize/templates/vivado/vivado-run.tcl.j2
@@ -1,5 +1,9 @@
 # Create a bin file which can be used to program the flash on the FPGA
 set_property STEPS.WRITE_BITSTREAM.ARGS.BIN_FILE true [get_runs impl_1]
+{% if tool_options.power_opt == 1 -%}
+# Enable power optimizations during implementation
+set_property strategy Power_DefaultOpt [get_runs impl_1]
+{%- endif %}
 
 # Vivado will raise an error if impl_1 is launched when it is already done. So
 # check the progress first and only launch if its not complete.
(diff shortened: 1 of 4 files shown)


sebinho:
So on my setup, it works well when using the FLOW approach. 
I am not sure why I get those checks failures. Maybe @olofk can help explaining why this fails?

Thanks

olofk:
Generally, we try to avoid adding explicit options when there are other ways to accomplish the same thing. The reason for that is there would otherwise be a million different flags which adds complexity. Instead, I propose doing the following:

Create a TCL file which contains `set_property strategy Power_DefaultOpt [get_runs impl_1]` and add that to a fileset with `file_type : tclSource`. Just tried that on a design and it seems to do the trick, but I don't have a good way to verify.

sebinho:
Thanks for the suggestion @olofk 
I tried and it does indeed work as expected. 

My only comment would be to say that the Vivado implementation step `power_opt_design` is quite common and one of the steps used very often. So I think it could be a good idea to somehow integrate this in the flow.
I know that Vivado has tons of flags that can be set, but this one seems like a big one.

Anyways, thanks a lot for your help.

olofk:
I understand your point, but one reflection I have made over the 12 years I have been working on this is that people are using these tools in such incredibly different ways. Things I thought everyone was doing turned out to of little use to others. We also used to have some assumptions built in that turned out to break code for others. That would not happen in this case, but it's the general reason why I am knowingly keeping things as lean as possible

Issue #417, opened 2024-03-08
QSys Pro 23 files are not processed correctly by the filter


GCHQDeveloper137:
I've recently had to create some new QSys files, and was surprised to see the build failed because the IP didn't get generated correctly. This was a bit of a surprised, as an existing project withb QSys in it still worked fine with the latest FuseSoC and Edalize.

On investigation, it seems that for v23, they've completely changed the format of the XML that QSys generates. After tracing through the code, it's become clear that the check inside qsys_file_filter in quartus.py is now failing to detect that this is a QSys Pro file, and so not running generation on it. To make life more complicated, they still call it version 1.0, so detecting it can't be a simple version check inside the file.

I think it should be fine to check both variants. I'll try some local changes, and if they work will put up a PR to work around this file format change. I see the fallback just assumed Standard rather than Pro - I suspect changing this to assume the same as the version of Quartus installed might be a better choice, so I might do this at the same time, unless anyone objects.

GCHQDeveloper137:
Aha, the fallback comment states that we should treat it as standard, but looking at the code it actually gets treated as the /opposite/ of the variant of the tool in use. That's clearly a very strange thing to do - if running standard the template will ignore this, and if running pro it would be filtered out so would ignore it! I'll make it default to the version in use, and update the comment to match the code.

GCHQDeveloper137:
Aha, the old behaviour was correct - we're matching versions up in here, my mistake. I'll just update the comments to be a little clearer.

olofk:
EDA tool vendors never cease to amaze me when it comes to finding new ways to create compatibility with the world around them, including their own tools.

As I don't have access to Quartus Pro myself, a PR would be most welcome if there is a fix needed.

GCHQDeveloper137:
I've got something that works fine for all the tool versions I have access to (old Standard, old Pro and new Pro). Thankfully, after updating the tests it spotted an issue with what you'd expect the new format Standard to look like, so I've fixed that. PR incoming shortly.

PR #418, opened 2024-03-08
Handle the new XML format for QSys from version 23.

GCHQDeveloper137:
It seems the format for QSys files changed significantly at some point, with at least version 23 using a completely different XML schema to before. This PR attempts version sanity checking using both formats. This should fix issue #417.


olofk:
Wow! Does this mean that Altera just moved to using IP-XACT 2014, two years after IP-XACT 2022 came out?? Not sure if I should cry or be happy that they at least didn't invent yet another format of their own.

Oh well, no way for me to test this so I trust your judgement on this one. Thanks for the fix. Picked and pushed

PR #421, opened 2024-04-02
Improve GateMate support

GMM-7550:
- pass Verilog defines and parameters, always define 'CCGM' (for conditionals)
- search Cologne Chip p_r tool in shell PATH

### DIFF
--- a/edalize/gatemate.py
+++ b/edalize/gatemate.py
@@ -71,7 +71,7 @@ class Gatemate(Edatool):
                 else:
                     ccf_file = f.name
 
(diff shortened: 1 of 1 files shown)


olofk:
Thanks! It would be great if we could also port the gatemate backend to the new flow API at some point.

olofk:
Changes merged now. Thanks for your contributions. I had to make some smaller syntax changes to pass the CI tests. [Here](https://edalize.readthedocs.io/en/latest/dev/setup.html) are some tips on how to make the tools do these checks for submitting a PR

Issue #426, opened 2024-04-12
How to read an EDIF netlist into Vivado that isn't the toplevel?


shareefj:
I'm working with some IP where they supply an EDIF netlist and I want to read this in as a submodule.  If I specify the source file as having a `file_type: edif`, then the `netlist_flow` boolean is set which in turn causes a `link_design` command to be run.  Also see #425 for a bug here.

https://github.com/olofk/edalize/blob/main/edalize/tools/vivado.py#L179

I can't say I've used a lot of EDIF netlists before so don't know when the `link_design` command is required.  In this case, if I just comment it out, then the flow completes without issue.  Is it only needed if you're trying to read in a toplevel netlist?

A simple fix looks to be to use the `tags` field in the Core file to set something like a `no_link_design` flag.

PR #427, opened 2024-04-12
feat[vivado]: fix for #426

shareefj:
This commit updates tools/vivado.py to allow the user to skip the link_design step normally used after reading in a toplevel netlist.

The use case here is when an EDIF netlist, or any other netlist, is read in as a submodule and is not the toplevel.  In this case the link_design step will fail as Vivado is expecting a netlist with the toplevel name.

In order to bypass the link_design step, use the no_link_design tag when specifying the netlist in your Core file:

filesets:
  netlist:
    files:
      - cute_a7_core/cute_a7_core.edf: { file_type: edif, tags: [no_link_design] }


olofk:
This looks good and I think tags are a good fit for this. Thanks for your contributions. Picked and pushed!

PR #433, opened 2024-05-08
Update questaformal.py

durmazh997:
Because of the seperate if conditions, args gets the empty list value when file type endswith -87 or -93.

### DIFF
--- a/edalize/questaformal.py
+++ b/edalize/questaformal.py
@@ -108,9 +108,9 @@ class Questaformal(Edatool):
                 cmd = "vcom"
                 if f.file_type.endswith("-87"):
                     args = ["-87"]
-                if f.file_type.endswith("-93"):
+                elif f.file_type.endswith("-93"):
                     args = ["-93"]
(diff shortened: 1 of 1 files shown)


olofk:
Oops! Thanks for catching that. Before merging this, I will need an update to the tests as well. You can run the test suite locally by running `tox` from the repo root. Set the `GOLDEN_RUN` environmnent variable to get some help with updating the reference files. More info here https://edalize.readthedocs.io/en/latest/dev/tests.html and let me know if you need any help.

yvishal-123:
my error is :-
**Trouble with Simulation Kernel.
make: *** [Makefile:36: run] Error 218
ERROR: Failed to run array:t1:rtl_top:1.0.0 : '['make', 'run']' exited with an error: 2**
when i am running manually that time the simulation is done and when i run the same files in fusesoc environment  i am facing the above issues.
can anyone help me to resolve the issues 
thanks in advance.

PR #437, opened 2024-06-24
Add support for official gowin tool

flaminggoat:
Example core:
```
  tang_nano_20k:
    default_tool: gowin
    description : Sipeed Tang Nano 20k
    filesets: [rtl, tang_nano_20k]
    flow: gowin
    flow_options:
      tool: gowin
      part: GW2AR-LV18QN88C8/I7
      gowin_options:
        - -multi_boot 1
    toplevel: top
```


olofk:
Thanks for adding this. Bonus points for also adding tests. Unfortunately this implementation is written for the Tool API that is about to be deprecated, and all new backends must instead support the Flow API. In practice this means adding a file in the tools subdirectory and then probably a short file in the flows subdirectory as well. I am fully aware that I haven't communicated this change very clearly and I apologize for that. Hopefully it's not too much work to change this. I think the efinity files can be useful as examples for how to write these files.

Let me know if you need assistance.

edalize/tools/gowin.py
```diff
@@ -0,0 +1,147 @@
...
+        if not self.tool_options.get("pnr"):
+            self.tool_options["pnr"] = "gowin"
+
+        if self.generic:
+            raise RuntimeError("Gowin does not support top level generics")
+
+        if self.vlogparam:
+            raise RuntimeError("Gowin does not support top level verilog parameters")
+
+        if self.vlogparam:
```
olofk, line 101: I guess this should be vlogdefine

olofk:
Very good. I think I found a typo, but looks fine otherwise. Appreciate if you could throw in some tests too. Doesn't have to be anything advanced. Check e.g. test_tool_verilator.py for reference and let me know if you have any questions.

flaminggoat:
@olofk I think that should be resolved now

olofk:
Sorry for taking so long to get back to this. It looks all fine and dandy. Picked and pushed. Thanks!

PR #438, opened 2024-06-24
Support nextpnr-himbaechel for gowin fpgas

flaminggoat:
This is required to be able to build for Gowin FPGAs such as GW2AR-LV18QN88C8/I7


olofk:
Happy to see more devices supported by nextpnr! As I wrote in https://github.com/olofk/edalize/pull/437 however, I need to have this converted to the flow API. In this case it means first updating tools/nextpnr.py, which should be very similar to what you have already done. Unfortunately the Apicula flow is not yet converted to the flow API, so we also need something similar to flows/icestorm.py in the flows directory as well.

PR #439, opened 2024-06-26
add SDC file type to radiant

johnathan-convertino-afrl:
SDC files are supported and allow for multiple constraint files vs only one with PDC.

### DIFF
--- a/edalize/radiant.py
+++ b/edalize/radiant.py
@@ -114,6 +114,7 @@ prj_close
             "systemVerilogSource": "prj_add_source ",
             "vhdlSource": "prj_add_source ",
             "PDC": "prj_add_source ",
+            "SDC": "prj_add_source ",
         }
         _file_type = get_file_type(f)
         if _file_type in file_types:


olofk:
Thanks. Could you also update the reference files accordingly so that the test suite passes? Some more info about testing and using the `GOLDEN_RUN` env var to update files automatically can be found here https://edalize.readthedocs.io/en/latest/dev/tests.html#testcases

johnathan-convertino-afrl:
Neat, I'll try to get on this soon(tm).

johnathan-convertino-afrl:
Ok a few issues, added it, tried it and it didn't seem to work. Could you give me a few pointers here? I assumed this should of been the exact same as the add for systemVerilog, but that fails.

Also Python 3.9 and greater are broken due to a issue unrelated to mine. Looks like a numpy version error.

johnathan-convertino-afrl:
Ok, so I've fixed the issue on my end. Though it looks like python 3.9 and up is broken due to a numpy version conflict. Nothing to do with this commit so I'm leaving it broken.

Radiant SDC test passes and fixed.

olofk:
Alright. Picked and pushed! Thank you for your contributions.

PR #445, opened 2024-10-04
Fixed workdir for ghdl in template creation

stdefeber:
As per ticket https://github.com/olofk/fusesoc/issues/707

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -136,7 +136,6 @@ Tada! We have simulated. As an exercise, try to just change the tool variable to
 
 Now it's time to create an FPGA image instead
 
-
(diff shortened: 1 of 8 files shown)


olofk:
Thank you for this. Generally it looks fine, but there are also some changes in tests/edalize_common.py that I don't see mentioned anywhere (and a removed line in the docs).

More problematic is that this patch is against the old tool API version of the ghdl backend. ghdl has been ported to the new Flow API, so this file will eventually be removed and all new development should go in https://github.com/olofk/edalize/blob/main/edalize/tools/ghdl.py

Issue #446, opened 2024-10-23
The Riviera pro backend of Edalize has minimal VHDL functionality


GCHQDeveloper211:
Currently, we are unable to add vcom options to Riviera pro's backend, as well as passing generic parameters to the RP simulator.

Additionally, the current version of VSIM in RP does not support the argument -quiet, therefore removing that as a default passed arg.

eminakgun:
Any updates?

PR #447, opened 2024-10-23
Add vhdl compatibility to riviera pro

GCHQDeveloper211:
This aims to resolve the #446 issue.


olofk:
The patch looks fine and all. I'm a bit reluctant however to add more features to the backends that implement the old tool API rather than the new flow API.

I will make an exception in this case, but I'm about to push Flow API support for QuestaSim, so for any further feature additions it would be great to use that as a base for a modernized rivierapro backend as well.

olofk:
A more severe issue is that it doesn't seem to pass CI tests, neither the formatting check or the actual tests. That needs to be fixed first.

olofk:
Fixed tests and formatting and pushed through 4a7afda3da9911958d25a0a77a9e3ec4e8619c29

PR #449, opened 2024-11-01
Updated SpyGlass backend to support features in Synopsys Spyglass v2023.12

amal-araweelo:
This pull request introduces enhancements to the Spyglass backend in the Edalize project. The key changes include:

**Support generalized SystemVerilog source files:**
- Updated the project file generation to support SystemVerilog without version limits (previously just SV09)

**Extending support of constraint file types**
-  Added support SpyGlass Design Constraint files (`.sgdc`)

**Support for `set_goal_option`**
- Implemented functionality for `set_goal_options`

This resolves the issue I opened (#444)


amal-araweelo:
@olofk is Edalize no longer in development, or?

edalize/spyglass.py
```diff
@@ -14,7 +14,7 @@
 
 class Spyglass(Edatool):
 
-    _description = """ Synopsys (formerly Atrenta) Spyglass Backend
+    _description = """ Synopsys Spyglass Backend (v2023.12)
```
olofk, line 17: Why is the version hardcoded in the description here?

olofk:
The patch looks mostly fine from a quick look but there are some issues, like the hardcoded version string. The test cases also aren't updated, so the tests are failing now.

I also need to inform that this backend uses the old Tool API instead of the new Flow API, and I'm a bit hesitant towards making updates to the old backends. I would much prefer if it was instead rewritten to use the new Flow API.

amal-araweelo:
@olofk wouldn't I still have to change the spyglass tool API in order to create a flow with the extended functionality? Or what do you suggest?

PR #450, opened 2024-11-08
Extend Vivado run script to support FPGAs of the Versal family

schenfab:
Building bitstreams for the Versal family was not yet supported through the run script for Vivado. This change adds support for Versal FPGA's by the following means:

- The property `STEPS.WRITE_BITSTREAM.ARGS.BIN_FILE` is only set if it exists. For the Versal family, this step has been replaced by `write_device_image` and hence the property would not apply, throwing an error
- Versal bitstream files are ending in `.bif` instead of `.bit`, which made copying the `.bit` file throw an error. Now, after completion of the run, either a `.bit` or `.bif` file will be copied to the project workroot.

The test reference files have been updated accordingly...


olofk:
Excellent! Let's get this merged. Thank you for your contributions.

Issue #453, opened 2024-11-24
[Bug Report] - Missing reporting option in edalize v0.6.0


rbrglez:
After updating to edalize version v0.6.0, it seems the reporting extra option is no longer available. This was previously installable using:

``` bash
$ pip install edalize[reporting]
```

However, with the new version, the command fails with the following warning:

```bash
WARNING: edalize 0.6.0 does not provide the extra 'reporting'
```

PR #454, opened 2024-11-24
#453: Added reporting option to pyproyect.toml

rbrglez:


olofk:
Completely missed this PR, but it seems like this is already taken care of through some other PR. Thanks anyway for the patch.

PR #459, opened 2025-02-13
Support passing `vlogparam`s to Genus

jfrimmel:
_Note_: this is written from the viewpoint of a FuseSoC user, but this applies to all usages of edalize itself.

FuseSoC already accepts `vlogparam`s for the `genus` backend/tool, but does not write them to any TCL-file or other way to access it from the user-defined Genus script. Therefore one could not easily add parameters to the top-level module when synthesizing with Genus, effectively rendering them unusable.

In order to resolve this issue, this commit creates new variable called `ELABORATE_PARAMETERS`, which is set unconditionally. Its value is constructed in such a way, that the value can be passed directly to the `elaborate`-command inside the user-supplied Genus TCL script like so:
```tcl
elaborate -parameters "$ELABORATE_PARAMETERS" "$TOP_MODULE"
```
This way, the user-supplied TCL-script can be kept generic (without the need for special-casing parameter) and FuseSoC makes use of the defined parameters.

### DIFF
--- a/edalize/genus.py
+++ b/edalize/genus.py
@@ -86,6 +86,10 @@ class Genus(Edatool):
         jobs = self.tool_options.get("jobs", None)
         jobs = "$nproc" if "all" in jobs else jobs
 
+        # Build up parameters in the format that the `elaborate` command expects
+        parameters = [f"{{ {name} {value} }}" for name, value in self.vlogparam.items()]
+        elaborate_parameters = "{ " + " ".join(parameters) + " }"
+
         template_vars = {
             "name": self.name,
             "src_files": src_files,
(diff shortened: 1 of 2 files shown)


jfrimmel:
Note, that `genus.py` already advertised support for `vlogparam`s:
https://github.com/olofk/edalize/blob/2759689ac4848e8bd54bcefd5e7bbfc949f9f356/edalize/genus.py#L30

olofk:
There are several issues with the genus interface and should be rewritten from scratch for the Flow API. I have other things I need to focus on right now though, but this is high on my TODO list.

olofk:
Ah! Missed that this was a  PR and not an issue and that you had provided a fix. I'm generally sceptic towards adding more features to the backends using the old Tool API though since I would like to focus people's energy on deprecating these old backends to reduce the maintenance burden for me. I would also like to have a test case to verify that it works both with and without parameters set.

But first of all, we need to make sure it passes the regression tests. Right now it's failing the formatting check. Make sure you have pre-commit set up for Edalize development https://edalize.readthedocs.io/en/latest/dev/setup.html

jfrimmel:
I've corrected the formatting, which is the easy part :wink: 

What would have to be done to convert ithe Genus backend to the new flow API? Are there any developer instructions?

PR #461, opened 2025-02-20
Copy PDI instead of BIF and allow multiple PDI files

stefanaltorfer:
The BIF is used by bootgen to create the PDI, the actual device image, so it makes more sense to copy the PDI. A vivado project setting for Versal, called "segmented configuration" results in multiple PDI files, so it makes sense to copy all of them.


olofk:
Let's go for this. I haven't used this flow myself but trust that it's the right thing to do.

PR #462, opened 2025-04-09
Improve GHDL tool 2

flaminggoat:
Follows on from the work on  #412 

Completes the TOOL_OPTIONS section

Adds default_library option

GHDL works with the sim flow


flaminggoat:
I think there are some issues with named libraries, I will look into that

olofk:
Many thanks. Picked and pushed

PR #465, opened 2025-06-25
slang: Added slang-tidy support to the lint flow

HU90m:


olofk:
Thanks Hugo,

This looks good except for a couple of things.
* Missing test cases
* Missing support for parameters/defines
* The default target is `slang_tidy`. This works fine if slang is used stand-alone, but if it will break if we run it e.g. as a frontend to another tool since we don't create a file called `slang_tidy` that the next tool could pick up. It means slang also always rerun even if there are no changes to the input files. I looked at the slang options to see if we could make it produce a log file or similar, but the closest thing I could find was `--diag-json`. The other option is to do something like `clang-tidy ... && touch slang.done` and then use `slang.done` as the default target. Any ideas here? I realize the same issue goes for verilator in lint-mode and should be fixed eventually.

Also, how do you turn off warnings? I'm getting a million of these `warning: [STYLE-2] port 'o_q' is not correctly suffixed with suffix: "_o"` because I'm using a style guide that has the direction as a prefix instead of a suffix :)

PR #469, opened 2025-08-12
Add processors option to quartus : #468

GCHQDeveloper211:


olofk:
I haven't checked the Quartus manuals, but could you confirm if this is the only way to change numbers of CPUs to use? It seems very strange to encode that information in the project files (although it wouldn't be the strangest decision by an EDA vendor).

PR #472, opened 2025-08-20
Change Vivado project not to be order only dep

### LINKED ISSUE #423: Vivado flow not re-run when Verilog parameter is changed
With the way the current Makefile is written, Synthesis, Implementation, etc. are not re-run when a Verilog parameter is changed in the `edam`. This is because the `.xpr` file is an [order-only prerequisite](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html) of the successive targets, and is the only file updated when the parameter is changed.

I'm really not a makefile expert. My gut tells me one of two solutions needs to be used here. One is having design-affecting parameters in their own `.tcl` file that is a normal prerequisite of the synth target (or have it be part of the existing `_synth.tcl` which is already a prerequisite there). The other is using some sort of phony target, but I can't quite see a justification for that when there is a material code change to prompt the rebuild.

I'd be willing to help with working towards a solution here, but I'm wondering if there aren't other gaps where changes in the design need to be added to this issue and we can try to solve them all at once.

Lucas-KB:
Fixes olofk/edalize#423 by making the Vivado project a normal dependency of the other targets to ensure synthesis and implementation are rerun if a source file is updated.

### DIFF
--- a/edalize/tools/vivado.py
+++ b/edalize/tools/vivado.py
@@ -219,8 +219,7 @@ class Vivado(Edatool):
             commands.add(
                 vivado_command + tcl_scripts + [project_file],
                 targets,
-                tcl_scripts,
-                [project_file],
+                tcl_scripts + [project_file],
             )
         else:
             targets = edif_files
@@ -233,8 +232,7 @@ class Vivado(Edatool):
         commands.add(
             command=vivado_command + tcl_scripts + [project_file],
             targets=[bitstream],
-            depends=tcl_scripts,
-            order_only_deps=[project_file],
+            depends=tcl_scripts + [project_file],
         )
 
         commands.add(["vivado", project_file], ["build-gui"], [project_file])
--- a/tests/test_vivado/Makefile
+++ b/tests/test_vivado/Makefile
@@ -7,12 +7,12 @@ pre_build:
 test_vivado_0.xpr: test_vivado_0.tcl sdc_file sv_file.sv tcl_file.tcl vlog_file.v vlog_with_define.v vlog05_file.v vlog_incfile vhdl_file.vhd vhdl_lfile vhdl2008_file xci_file.xci xdc_file.xdc bootrom.mem another_sv_file.sv | pre_build
 	$(EDALIZE_LAUNCHER) vivado -notrace -mode batch -source test_vivado_0.tcl
 
-test_vivado_0.v test_vivado_0.edn: test_vivado_0_synth.tcl test_vivado_0_netlist.tcl | pre_build test_vivado_0.xpr
+test_vivado_0.v test_vivado_0.edn: test_vivado_0_synth.tcl test_vivado_0_netlist.tcl test_vivado_0.xpr | pre_build
 	$(EDALIZE_LAUNCHER) vivado -notrace -mode batch -source test_vivado_0_synth.tcl test_vivado_0_netlist.tcl test_vivado_0.xpr
 
 synth: test_vivado_0.v test_vivado_0.edn | pre_build
 
-test_vivado_0.bit: test_vivado_0_synth.tcl test_vivado_0_run.tcl | pre_build test_vivado_0.xpr
+test_vivado_0.bit: test_vivado_0_synth.tcl test_vivado_0_run.tcl test_vivado_0.xpr | pre_build
 	$(EDALIZE_LAUNCHER) vivado -notrace -mode batch -source test_vivado_0_synth.tcl test_vivado_0_run.tcl test_vivado_0.xpr
 
 build-gui: test_vivado_0.xpr | pre_build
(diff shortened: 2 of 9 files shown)


craigjb:
I'm also interested in this PR for FuseSoC. It seems like changes to verilog source files aren't triggering a build. I have to pass `--clean` to `fusesoc run ...` every time. The source file in this case is coming from a generator (not cached).

olofk:
AFAIR I had to make the project file an order-only dep because it kept getting updated by Vivado which meant that everything that depended on the XPR got unconditionally rebuilt even if nothing changed. Looking at it now, however I see that the current logic doesn't really make sense. Maybe we can change the project generation script to write out a marker file and use that instead of the xpr to track when we need to rebuild?

Lucas-KB:
That makes sense.

When I tested this before, I beleive that the only thing that depended on the XPR file was the synthesis and implementation runs, but those runs check wheather or not they are complete before running. So, in practice, if nothing changed, nothing would be run.

On the other hand, in case something else depends on the XPR file, that recipe would be run again.

Your solution of adding a marker file seems good.

nicorum:
> AFAIR I had to make the project file an order-only dep because it kept getting updated by Vivado which meant that everything that depended on the XPR got unconditionally rebuilt even if nothing changed. Looking at it now, however I see that the current logic doesn't really make sense. Maybe we can change the project generation script to write out a marker file and use that instead of the xpr to track when we need to rebuild?

I think this is more of a dependency issue. The xpr shouldn’t be rebuilt by Make at all if its dependencies don’t change. I also can’t reproduce this problem with the patch.
Perhaps fusesoc/edalize updated the timestamps of unmodified dependency files in an earlier version? 

If there are file types that don’t affect the xpr, they could also simply be set as dependencies of the other targets instead of the xpr. (Maybe just TCL scripts for the XPR?)

Lucas-KB:
I am not sure I understand everything you said.

> I think this is more of a dependency issue. The xpr shouldn’t be rebuilt by Make at all if its dependencies don’t change.

That part is ok to me.

> I also can’t reproduce this problem with the patch. Perhaps fusesoc/edalize updated the timestamps of unmodified dependency files in an earlier version?

What did you try to reproduce? The idea of the patch is to ensure the synthesis and implementation are rerun every time a source file is changed. The thing is, the current behavior is to remake the project file every time any source file changes.

> If there are file types that don’t affect the xpr, they could also simply be set as dependencies of the other targets instead of the xpr. (Maybe just TCL scripts for the XPR?)

That could be a solution, but maybe we would have problems with xci and bd files. Not sure about this.

nicorum:
> What did you try to reproduce? The idea of the patch is to ensure the synthesis and implementation are rerun every time a source file is changed. The thing is, the current behavior is to remake the project file every time any source file changes.

I've been trying to understand oflofk's reasoning behind using order-only dep. He said that xpr gets updated even if nothing changes. But that's exactly what Make isn't supposed to do. And that didn't happen to me either with your patch. Or am I misunderstanding something here?


> That could be a solution, but maybe we would have problems with xci and bd files. Not sure about this.

Yepp. The question is what dependencies the XPR file should actually have. And since Vivado is a black box to us, it’s pretty hard to figure that out. But we would have the same problem with the marker file. At some point, we have to decide which file changes trigger an update to the XPR file.


But I also wonder if that isn't a bit of over-optimization. At the risk of overlooking something.
The current situation at least has the advantage that it consistently doesn't work without “--clean”. In the worst case, that then depends on a few corner cases.

Lucas-KB:
My understanding is that Vivado updates the xpr file sometimes if I just open the project, for example, and that could rerun the synthesis. But my argument is that wouldn't happen because the synthesis script used by edalize checks if the synthesis is already complete.

In my opinion I wouldn't mind recreating the project every time a source file changed. As you mention, we could easily forget some case if we add some specific file types as dependencies. It seems a better scenario than the current behavior, which is: modify something; run make; go get some coffee; come back and nothing ran because of this issue. In the projects I have worked on, the project creation itself is a lot slower than synthesis and implementation.

nicorum:
Oh, now I get it. I didn't know that Vivado sometimes randomly changes the xpr's timestamp.
I’ve now created another pull request that writes a marker file right at the end of the project generation tcl script.

But I also think that’s a bit of over-optimization. When you use `make pgm`, the Vivado activity can easily give you the impression that the bitstream has been rebuild. Especially if you're a inexperienced user. “Something’s running, after all.” I really like fusesoc, but for me the UX with Vivado is pretty bad because of this bug.

olofk:
Yes, the core problem is that Vivado updates the xpr file after each step (synthesis, placement, routing, etc) which prevents us from having a dependency chain of "(source files) -> (project tcl) -> (.bit)" or "(source files) -> (edf) -> (project tcl) -> (bit)" if we're using yosys for synthesis. Not sure if the latter flow actually works right now tbh, but either way, let's try with the marker file as in #526

Issue #474, opened 2025-10-06
Vivado: used_in_implementation property not supported for .vhd/.v files with simulation tag


rbrglez:
# Description:

When using the simulation tag on testbench files in Vivado via FuseSoc + Edalize, Vivado fails to open the design. This is because the tool attempts to set the used_in_implementation property on VHDL and Verilog source files, which is not supported.

**Error message from Vivado**:
```bash
ERROR: [Common 17-54] The object 'file' does not have a property 'used_in_implementation'.
```

# Reproduction:

Here is an excerpt from the core file I used to test this behavior:

```yaml
################################################################################
# Filesets
################################################################################
filesets:
  rtl_files:
    depend:
      - "^open-logic:open-logic-aviat:base:4.0.0"

    files:
      - "rtl/bucket.vhd"
      - "rtl/checkpoint.vhd"
      - "rtl/leaky_bucket.vhd"
    file_type: "vhdlSource-2008"

  tb_files:
    files:
      - "tb/bucket_tb.vhd"
      - "tb/checkpoint_tb.vhd"
      - "tb/leaky_bucket_tb.vhd"
    file_type: "vhdlSource-2008"
    tags:
      - "simulation"
```

# Offending code in Edalize (Tool Vivado backend):

https://github.com/olofk/edalize/blob/6f3260c67331bd300443014b024219214337da82/edalize/tools/vivado.py#L155-L157

The same issue is also present in Flow Vivado backend, but unfortunately I didn't find the offending code.

# Proposed solution:

Skip setting used_in_implementation if the file type is a source file (vhdlSource, verilogSource, etc.). These properties are not valid for HDL files in Vivado and will generate an error.

rbrglez:
I've experimented a bit with the `USED_IN` and `USED_IN_IMPLEMENTATION` properties in Vivado 2023.2.

To exclude VHDL files from the implementation flow, you should use the `USED_IN` property, not `USED_IN_IMPLEMENTATION`.

How `USED_IN` Works:

The `USED_IN` property explicitly defines which design flows a file is used in. When you set it, only the specified values (e.g., synthesis, simulation, etc.) are enabled. All other flows are implicitly disabled.

This works as expected:

```tcl
set_property USED_IN {simulation synthesis} [get_files *.vhd]
```

This doesn't work:

```tcl
set_property used_in_implementation false [get_files *.vhd]
```

[...]

PR #475, opened 2025-10-06
Fix Vivado property usage for simulation files (#474)

rbrglez:
Replace `used_in_implementation` and `used_in_synthesis` with `used_in`

Vivado does not support `used_in_implementation` for all file types (e.g., Verilog and VHDL). Replaced both `used_in_implementation` and `used_in_synthesis` with the `used_in` property, which explicitly defines the allowed design flows and works consistently across all file types.

### DIFF
--- a/edalize/tools/vivado.py
+++ b/edalize/tools/vivado.py
@@ -150,10 +150,7 @@ class Vivado(Edatool):
                         f"move_files -fileset sim_1 [get_files {f['name']}]"
                     )
                     sim_files.append(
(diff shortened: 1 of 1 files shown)


olofk:
The fix looks good, but we need to also update the tests because they are failing with this change. Easiest way to see what needs to be done is to run `GOLDEN_RUN=1 tox` from the edalize root and it will update the reference files with the expected changes. Check and add the changed files and update the PR.

Issue #480, opened 2025-11-06
flow_make_options should be defined in the Edaflow parent class


ivanvig:
The Make build runner currently retrieves its build_options from flow_make_options. However, this variable is defined only in the generic flow, which causes the option to be ignored in all other flows. To fix this, flow_make_options should be defined in the Edaflow parent class, ensuring that it is properly inherited and available to all child flows.

PR #481, opened 2025-11-11
Vivado flow now adds synth arg if pnr is none

### LINKED ISSUE #479: Vivado flow ignores pnr=none
This issue is the root cause of #471. It occurs because the Vivado flow is inheriting the edaflow build method, which does not handle the pnr flow option. As a result, when pnr is set to none, the flow fails to include the required synth argument, leading to the process not stopping at the synthesis stage

ivanvig:
Fixes #479. The idea is to override the inherited build method so it properly accounts for the pnr option and appends the synth argument when is needed, basically mirroring how it’s handled in the Vivado tool code.

### DIFF
--- a/edalize/flows/vivado.py
+++ b/edalize/flows/vivado.py
@@ -66,6 +66,13 @@ class Vivado(Edaflow):
         self.commands.set_default_target(name + ".bit")
         return FlowGraph.fromdict(flow)
 
+    def build(self):
+        (cmd, args) = self.build_runner.get_build_command()
+        pnr_opt = self.flow_options.get("pnr", "")
+        if pnr_opt == "none":
+            args.append("synth")
+        self._run_tool(cmd, args=args, cwd=self.work_root)
+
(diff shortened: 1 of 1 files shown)


mcarter-cyber:
Is this not caused by the build method in the vivado tool class?

https://github.com/olofk/edalize/blob/7f5efda7df26524db193a82996fdad43ecfc3a7b/edalize/tools/vivado.py#L280-L288

"synth" is appended to `args`, but the method returns `self.args`

ivanvig:
afaik the vivado tool build method never gets called using the flow api. 

This is the inherited build method from edaflow

https://github.com/olofk/edalize/blob/8c74627c2ba9b79df9df604a74d0948174c758e3/edalize/flows/edaflow.py#L368-L370

It gets the build command (by default make) and the build args (which currently cant be set for the make builder #480 ). And then executes that with _run_tool() which is just a wrapper for subprocess.run()

PR #483, opened 2025-11-13
Added a OSS flow, apicula, for the Gowin series of FPGAs

blueluna:
Apicula is the name of the Yosys Gowin bitstream format project.

 - Added a apicula flow. Using Yosys, Nextpnr, gowin_pack and openFPGALoader.
 - Added gowin_pack tool.
 - Added openFPGALoader tool.
 - Updated Nexpnr tool to use nextpnr-himbaechel for gowin instead of the deprecated nextpnr-gowin.
 - Added tests for the above.

Tested on Tang Nano 9K and Tang Nano 20K using version 20251110 of Yosys oss-cad-suite.


blueluna:
Happy to help.

Regarding `device` and `device_family`. My understanding is that it is not possible to decode the device family from the device code alone for some Gowin devices.
See this note in the [Apicula project](https://github.com/YosysHQ/apicula?tab=readme-ov-file#user-content-fn-1-29098bd01cb9f2c7d0a966844db703b5) and [this comment](https://github.com/YosysHQ/nextpnr/issues/1203#issuecomment-1703337781).

PR #497, opened 2026-01-28
New Tool Support: VC Formal

ArkoshEternal:
# Add [VC Formal support](https://www.synopsys.com/verification/static-and-formal-verification/vc-formal.html) to Edalize

## Feature Overview 
* Utilize similar jinja template method to that of the symbiyosys to run vcformal (vcf) applications 
* CAPI2 selected VC formal application (FPV, FSV, etc.) mandatory tool option
* Add new filetype, vcfConfigTemplate, to CAPI2 file typing system

## Verification
* Add testing for vcf tcl file generation, tests against assumed CLI command (most information contained in template .tcl file) 
* In use in fork w/ VCF 2024

### DIFF
--- a/edalize/vcf.py
+++ b/edalize/vcf.py
@@ -1,0 +1,163 @@
+# Copyright edalize contributors
+# Licensed under the 2-Clause BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-2-Clause
+
+import os
+import re
+import logging
+import jinja2
+from edalize.edatool import Edatool
+
+logger = logging.getLogger(__name__)
+
+
+class Vcf(Edatool):
+
+    _description = """Synopsys Formal Backend
+
+VC Formal is Synopsys's formal verification tool. 
+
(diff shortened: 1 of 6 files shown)


olofk:
The code looks straight-forward (A+ for adding tests) but this should be converted to the Flow API instead as we are sunsetting the old Tool API.

ArkoshEternal:
@olofk Sounds great, I'll port this to a flow API. 

I suppose this would be part of a "formal" flow, but I feel the term formal has so much loading to it that it might be a bit awkward to fit everything under a "formal" flow

I'll start working on one, but I'm wondering if you have an idea of what other tools I should implement under the flow

olofk:
Since it's a single tool, I think it can be run with the `generic` flow. That's what I use for other similar one-off tools. I think we can create a dedicated `formal` flow when we have a need for it. Like for the lint flow that sets some special options for verilator, or the sim flow where there is some shared cocotb logic.

FYI, I'm also considering making the `generic` flow the default if the user doesn't specify a flow in the target.

PR #500, opened 2026-02-13
Flow Config File Setup - Append

ArkoshEternal:
# Update Config File - Append

Modify functionality of the edatool update_config_file function to append the "contents" string to a config file if it does not exist in the file or if the file does not already exist. 

## Motivation
This change is motivated by allowing custom config files (e.g. synopsys_sim.setup) to be included in flows as a dependency. The unique pain point is using precompiled libraries with VCS, which necessitates use of the setup file instead of a vmap or equivalent from other major commercial simulators.

## Testing
Simple test added to confirm append method, which ensures "content" (at least for VCS) is the most recent config option if it needs to overwrite prior config options. Also tests to ensure we aren't re-adding content. 

This is weak to the contents not being in the same order in the config file as they are in the "content" variable, but in the case of VCS and I assume others, this should hopefully be overlooked as the config will be the most recent option

## Possible Other Bug
It should also be noted that the initial use case that spawned this, a `copyto: synopsys_sim.setup` file, worked if I ran fusesoc run ... --setup, and then ran later w/ --build/--run. The flow tool would not overwrite `synopsys_sim.setup`

I need to look more at the flow-graph to determine why that happened

### DIFF
--- a/tests/test_edatool_update_config.py
+++ b/tests/test_edatool_update_config.py
@@ -1,0 +1,28 @@
+import os
+
+from edalize.tools.edatool import Edatool
+
+
+def test_update_config_file_appends_and_dedup(tmp_path):
+    work_root = str(tmp_path)
+
+    tool = Edatool()
+    tool.work_root = work_root
+
+    fname = "cfg.txt"
+    fpath = os.path.join(work_root, fname)
+
+    with open(fpath, "w") as f:
+        f.write("initial\n")
+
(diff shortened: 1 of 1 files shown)


olofk:
I was wondering if this would become an issue when I was working on the VCS backend and, yep, apparently it did :) Let's go for this and we'll modify it if problems arise later.

PR #513, opened 2026-04-01
[doc] Flow name fixed

marnovandermaas:
I think the name of the target and the flow should be the same here.


olofk:
Oops. You're right. Thanks for that. Picked and pushed

Issue #515, opened 2026-04-29
VHDL generics are not supported in the Vivado flow


GCHQDeveloper137:
Whilst migrating our FuseSoC cores' Vivado targets from the tool API to the flow API, we found that the flow API does not seem to support parameters of type "generic". This means that any core that uses VHDL generics on the top level will not work with the Vivado flow API, but will work fine with the Vivado tool API, meaning such cores cannot be transitioned with Edalize 0.6.8.

I've done some testing, and found that if `argtypes` in flow/vivado.py is updated to include "generic", that seems to be all that's required; TCL is generated to set the generic correctly, and the resulting design is processed as expected by Vivado.

I can't find an existing issue for this problem, so have reported it now. I'll try to get a pull request up to fix this shortly.

PR #516, opened 2026-04-29
Add "generic" to the list of supported parameter types

GCHQDeveloper137:
It looks like the code to handle this is already present, as with this in place it generates the correct TCL to set the generics.

Without this change, FuseSoC cores using `flow: vivado` will print a warning for a each generic parameter for a given taget, and if an attempt is make to set that generic on the command line, the build will fail. With this change, the warnings are gone and CLI options to set generics are accepted and transformed into correct TCL, resulting in Vivado working as expected.

Fixes issue #515


olofk:
This makes sense. Thanks for fixing.

PR #518, opened 2026-05-11
Fix AttributeError in tools/vpr.py when EDAM contains an SDC file

ThVerg:
Bug: `tools/vpr.py:89` accesses `f.name` on a dict, raising `AttributeError` when an SDC file is present. Fix is to use again `f["name"]` like lines 84 and 87.


olofk:
Thanks!

PR #519, opened 2026-05-11
flows/vpr: fix 'ftdo' typo to 'fdto'

ThVerg:
The `ftdo` key in `flows/vpr.py`'s yosys flow node is silently dropped because `Edaflow` only recognises `fdto` (flow-defined tool options). As a result yosys is never told to emit BLIF output for the VPR flow.

One-character fix.


olofk:
Thanks!

PR #520, opened 2026-05-11
tools/sandpipersaas: pass dependency list to commands.add, not a string

ThVerg:
The `clean` target was registered with `depends=" "` (a single-space string). `EdaCommands.add` expects a list of dependency names; the string is wrong and was never intentional.


olofk:
Thanks!

PR #521, opened 2026-05-11
tools/surelog: fix UnboundLocalError when vlogdefine is non-empty

ThVerg:
The Verilog-defines loop appended to `verilog_params` instead of `verilog_defines`. `verilog_params` is not defined at that point (assigned a few lines later), so the loop raises `UnboundLocalError` whenever `self.vlogdefine` is non-empty.


olofk:
oops, you're right. Thanks for your contributions. Picked and pushed. Will add a test for that backend too.

PR #523, opened 2026-05-11
tools/vpr: return empty args from build(), not undefined self.args

ThVerg:
`Vpr.build()` returns `self.args`, which is never assigned. Unlike `Vivado.build()`, there is no local `args` to recover, so the fix is conservative: return an empty list. Happy to change if a different intent is preferred.


olofk:
Good catch. I think we should completely drop the function though. None of the other tools, except Vivado, even has a build function. IIRC there was some special reason for having that in the vivado build, but the one in vpr should be removed.

PR #526, opened 2026-06-15
Added marker file for Vivado project

### LINKED ISSUE #423: Vivado flow not re-run when Verilog parameter is changed

nicorum:
Use a markerfile, as suggested in #472, to fix  #423. The marker file gets updated by the vivado-project.tcl, This ensures that further synthesis are only triggered if the xpr has been updated by the project.tcl.

As a result,  a synthesis is always performed if changes are made to the design files (RTL, xdc, etc.). But no synthesis is triggered if Vivado changes the timestamp of the xpr by only opening it.

### DIFF
--- a/tests/test_vivado/board_file/vivado.cmd
+++ b/tests/test_vivado/board_file/vivado.cmd
@@ -1,3 +1,2 @@
 -notrace -mode batch -source test_vivado_0.tcl
--notrace -mode batch -source test_vivado_0.tcl
 -notrace -mode batch -source test_vivado_0_synth.tcl test_vivado_0_run.tcl test_vivado_0.xpr
--- a/tests/test_vivado/board_file/vivado.cmd
+++ b/tests/test_vivado/board_file/vivado.cmd
@@ -1,2 +1,3 @@
 -notrace -mode batch -source test_vivado_0.tcl
+-notrace -mode batch -source test_vivado_0.tcl
 -notrace -mode batch -source test_vivado_0_synth.tcl test_vivado_0_run.tcl test_vivado_0.xpr
--- a/tests/test_vivado/edif_netlist/vivado.cmd
+++ b/tests/test_vivado/edif_netlist/vivado.cmd
@@ -1,3 +1,2 @@
 -notrace -mode batch -source test_vivado_0.tcl
--notrace -mode batch -source test_vivado_0.tcl
 -notrace -mode batch -source test_vivado_0_synth.tcl test_vivado_0_run.tcl test_vivado_0.xpr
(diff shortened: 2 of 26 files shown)


tests/test_vivado/board_file/vivado.cmd
```diff
@@ -1,2 +1,3 @@
 -notrace -mode batch -source test_vivado_0.tcl
+-notrace -mode batch -source test_vivado_0.tcl
```
olofk: Why is this line duplicated here? Doesn't look right

tests/test_vivado/edif_netlist/vivado.cmd
```diff
@@ -1,2 +1,3 @@
 -notrace -mode batch -source test_vivado_0.tcl
+-notrace -mode batch -source test_vivado_0.tcl
```
olofk: Same as previously noted. Duplicated line points to something else that's not correct.

olofk (commented):
Thanks for this. The code looks fine from a quick look, but I'm a bit worried about some lingering bug that causes the duplication in the test reference scripts. Would appreciate if you could take a look at that or let me know if I misunderstood something before I merge.

tests/test_vivado/board_file/vivado.cmd
nicorum: Oh, I overlooked that. It’s because the Vivado mock doesn’t create the *.xpr/*.xpr.marker files, etc. Since we now have xpr.marker and xpr as dependencies and these aren’t created when the target is called, make tries to run the target twice (once for *.xpr and once for *.xpr.marker).
The same problem also occurs with the _synth_ target.

I’ve now modified the mock so that it actually creates the files that are supposed to be generated by the target/vivado.
It would also probably make sense to declare targets that generate more than one file as “grouped targets,” so that they would never be executed multiple times in a single make call. However, this feature is only available since make 4.3.

olofk: Ah! That makes sense. Thanks for explaining and fixing. We have soon reimplemented most of Vivado in that mock file but it 's fine :)

olofk:
Perfect. Thanks for this. Squashed and merged!

PR #529, opened 2026-06-22
feat(xcelium): add the --gui option for Xcelium backend

### LINKED ISSUE #528: feat(xcelium): add the --gui option for the Xcelium backend
Currently it is not possible to invoke the GUI of the Cadence Xcelium simulator because the `xrun_options` options are passed only to the compilation/elaboration not during runtime.

https://github.com/olofk/edalize/blob/e33ebfb899449ebbac37b77df11427a247a84d71/edalize/tools/xcelium.py#L175

Additionally, the argument definition for the `xrun_options` is also broken because the argument type is `str` when backend is expecting `list`. Ideally, it should use the `shlex.parse` to handle multiple arguments, including whitespace characters. For that, I will create a separate issue.

https://github.com/olofk/edalize/blob/e33ebfb899449ebbac37b77df11427a247a84d71/edalize/tools/xcelium.py#L22

Because the `xcelium` backend is using snapshots in the `run` method, the `gui` option should be passed there.

https://github.com/olofk/edalize/blob/e33ebfb899449ebbac37b77df11427a247a84d71/edalize/tools/xcelium.py#L185

My proposition is to add a new backend option `--gui` that will invoke the GUI:

```python
    def run(self):
        args = ["-R"]

        if self.tool_options.get("gui"):
            args += ["-gui"]
```

tymonx:
Closes https://github.com/olofk/edalize/issues/528

PR adds the `--gui` option for the Xcelium backend that will allow to invoke the Graphical User Interface (GUI) of the Cadence Xcelium simulator.

Usage:

```plaintext
fusesoc run --target <xcelium-flow-target> <core> --gui
```

Added unit test:

```plaintext
pytest -k test_tool_xcelium_gui
```

Source files were also validated with the `pre-commit` (black):

```plaintext
pre-commit run --files ./edalize/tools/xcelium.py ./tests/test_tool_xcelium.py
```

### DIFF
--- a/edalize/tools/xcelium.py
+++ b/edalize/tools/xcelium.py
@@ -20,6 +20,7 @@ class Xcelium(Edatool):
         },
         "timescale": {"type": "str", "desc": "Default timescale for simulation"},
         "xrun_options": {"type": "str", "desc": "Additional run options for xrun"},
+        "gui": {"type": "bool", "desc": "Invoke the Graphical User Interface (GUI)"},
     }
 
     V_SRC_FILE_TYPES = ["verilogSource", "systemVerilogSource"]
@@ -183,6 +184,9 @@ class Xcelium(Edatool):
 
     def run(self):
         args = ["-R"]
+
+        if self.tool_options.get("gui"):
+            args += ["-gui"]
 
         # Set plusargs
         if self.plusarg:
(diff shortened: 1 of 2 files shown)


olofk:
Excellent! Thanks for that.

PR #537, opened 2026-06-24
fix(xcelium): handle xrun_options passed from FuseSoC command line

### LINKED ISSUE #530: bug(xcelium): cannot use the fusesoc run --xrun_options from command line because option is a string type
Edalize is expecting that the `xrun_options` is a list. But the `xrun_options` defined in the `TOOL_OPTIONS` is using the `str` type:

https://github.com/olofk/edalize/blob/e33ebfb899449ebbac37b77df11427a247a84d71/edalize/tools/xcelium.py#L22

Unfortunately the mapping is limited on the FuseSoC side:

https://github.com/olofk/fusesoc/blob/f15e1c8a76815c4f391231dd0e743e2b683c6b45/fusesoc/edalizer.py#L478

Generally this is a common problem for all tools.

Proposition is to add a new small utility function:

```python
import shlex
from typing import Any
from collections.abc import Iterable

def to_args(value: Any) -> list[str]:
    """Convert any provided value to list of strings."""
    if not value:
        return []

    if isinstance(value, str):
        return shlex.parse(value)

    if isinstance(value, Iterable):
        return [str(item) for item in value]

    return to_args(str(value))
```

Then it could be used like this:

```python
args += to_args(self.tool_options.get("xrun_options"))
```

tymonx:
Closes https://github.com/olofk/edalize/issues/530

This PR:

* Fixes passing the `--xrun_options="<args>"` from the `fusesoc` command line interface to the Edalize `xcelium` backend

Correct way to enable handling list for an option: https://github.com/olofk/fusesoc/pull/791#issuecomment-4864705684

### DIFF
--- a/edalize/tools/xcelium.py
+++ b/edalize/tools/xcelium.py
@@ -19,7 +19,11 @@ class Xcelium(Edatool):
             "desc": "Disable 64-bit mode",
         },
         "timescale": {"type": "str", "desc": "Default timescale for simulation"},
-        "xrun_options": {"type": "str", "desc": "Additional run options for xrun"},
+        "xrun_options": {
+            "type": "str",
+            "desc": "Additional run options for xrun",
+            "list": True,
+        },
         "gui": {"type": "bool", "desc": "Invoke the Graphical User Interface (GUI)"},
     }
 


olofk:
Hmm... it's a clever solution, but xrun_options should really be a list and I think we better fix that bug instead. I didn't catch that when I did the initial review.

tymonx:
@olofk Alternatively we could introduce a new type `list` or `args`. But this is not supported on the FuseSoC side:

https://github.com/olofk/fusesoc/blob/f15e1c8a76815c4f391231dd0e743e2b683c6b45/fusesoc/edalizer.py#L335

```python
    def _build_parser(self, backend_class, edam):
        typedict = {
            "bool": {"type": str2bool, "nargs": "?", "const": True},
            "file": {"type": str, "nargs": 1, "action": FileAction},
            "int": {"type": int, "nargs": 1},
            "str": {"type": str, "nargs": 1},
            "real": {"type": float, "nargs": 1},
        }
```

I can create an issue and PR in the FuseSoC project to add the `list` or `args` type and later update this PR.

The `str` type is definitely not clear here because core files are using YAML array with strings and tool backend is expecting list of arguments (aka list of strings).

tymonx:
For now, I have moved this PR to the Draft state. I will create a separate PR in the FuseSoC project https://github.com/olofk/fusesoc/issues/790 to add the `list` type and refactor this PR to use that type instead of the `str` type.

tymonx:
If this will be merged https://github.com/olofk/fusesoc/pull/791 then this PR can be simplified by replacing the `str` type with the `list` type.

tymonx:
@olofk Ready for review and approve. Sorry for the mess with the `list` type. I followed your comment from the FuseSoC project: https://github.com/olofk/fusesoc/pull/791#issuecomment-4864705684

I have added only the `"list": True` to the `xrun_options`.

olofk:
Great! I was hoping something like this would fix the issue. Picked and pushed

PR #544, opened 2026-07-02
flows: properly merge EDAM from multiple dependencies

ThVerg:
In `Edaflow.configure_tools()`, the nested `merge_edam()` is a placeholder that just returns its second argument ("Yeah, I know. It's just a temporary hack"). For a fan-in node — one with more than one dependency — the input EDAM is whichever dependency was merged last, silently dropping the files, hooks and tool options contributed by the other dependencies.

This replaces the hack with a real merge, promoted to a module-level `merge_edam()` so it is unit-testable:

- dicts merge recursively; scalar conflicts resolve to the later dependency
- lists concatenate, skipping entries that already exist in the earlier list — files/hooks inherited from a common ancestor node appear once instead of once per dependency
- neither input EDAM is mutated

Single-dependency nodes receive the same content as before (the merge of `{}` with the dep's EDAM). Unit tests included; full suite passes (207 passed).

### DIFF
--- a/edalize/flows/edaflow.py
+++ b/edalize/flows/edaflow.py
@@ -72,6 +72,29 @@ def merge_dict(d1, d2):
     return d1
 
 
+def merge_edam(a, b):
+    """Merge two EDAM dicts coming from different flow graph dependencies.
+
+    Dicts are merged recursively and scalar values from the second EDAM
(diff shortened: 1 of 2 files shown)


olofk:
Excellent! I've been having a bad conscience about that hack for many years now so happy to see it properly fixed :)

PR #545, opened 2026-07-02
tests: add tests for untested flows and flow-API tools

ThVerg:
This adds tests for the flow-API tools and flows that currently have no test coverage, using the existing golden-file harness (`tests/edalize_tool_common.py` / `tests/edalize_flow_common.py`, references regenerated with `GOLDEN_RUN`).

New flow-API tool tests (`tests/test_tool_*.py` + goldens under `tests/tools/`):
- ghdl (default std detection, `--std` override via analyze_options, verilog mode, toplevel with library)
- sv2v (with and without options)
- vpr (basic, generate_constraints/F4PGA variant, multiple-netlist error)
- icetime (basic, no-input and multiple-input errors)
- sandpipersaas (basic, output_dir/output_file/includes options, file-count and file-type errors)

New flow tests (`tests/test_flow_*.py` + goldens under `tests/flows/`):
- icestorm (default, pnr=none, invalid pnr error)
- vivado (vivado synth and yosys synth)
- vpr
- gowin (basic, missing part error, vlogparam rejection)
- efinity (basic, missing EFINITY_HOME error)

Notes:
- `tests/edalize_flow_common.py` gains an optional `paramtypes` argument (default unchanged) because the gowin flow rejects vlogparam/vlogdefine parameters.
- The sandpipersaas test sets a fixed `work_root` string instead of using the shared fixture, since the tool embeds `work_root` in the generated Makefile (and string-concatenates it, so it must not be a Path).
- Skipped: `flows/f4pga.py`. Its `configure_flow` still returns the legacy list-of-tuples flow description, which the current `Edaflow` base class cannot consume (`AttributeError: 'list' object has no attribute 'get_nodes'`), so it cannot be instantiated without a library fix.

Full suite passes: 231 tests (204 before, 27 added).


olofk:
Great! Much appreciated. I won't pretend that I looked at all the reference files in detail but the tests make sense.

PR #549, opened 2026-07-03
fix(xcelium): handling DPI libraries

### LINKED ISSUE #548: bug(xcelium): DPI libraries are added as a list of list to command arguments
When adding DPI libraries to the `xcelium` backend, they are added to the `xrun` command as `['-sv_lib', '<library>.so']` not as `-sv_lib <library>.so`. This is because created list of command arguments with DPI libraries is a list of list instead of flat list.

Root cause:

https://github.com/olofk/edalize/blob/7e5f16f7a3c29b3ae744e1f19fa6ea67530edd70/edalize/tools/xcelium.py#L104

Current command:

```plaintext
xrun ['-sv_lib', '<library>.so'] ...
```

Expected command:

```plaintext
xrun -sv_lib <library>.so ...
```

Fix:

```python
dpi_lib_cmd.extend(("-sv_lib", lib))
```

tymonx:
Closes https://github.com/olofk/edalize/issues/548

The `xrun` command is just a wrapper to invoke:

* `xmvhdl` to compile VHDL source files
* `xmvlog` to compile Verilog and SystemVerilog source files
* `xmelab` to perform elaboration of HDL module
* `xmsim` to run simulation

Handling DPI libraries is only done by the `xmsim` command (simulation runtime):

```plaintext
xmsim -help | grep -i dpi
        -DPI_STACK_INT_C                    -- Lists internal C stack frames called from import function
        -SV_LIB <arg>                       -- Dynamically load a DPI Library
```

The elaboration stage is only used to generate header files for export/import functions/tasks:

```plaintext
xmelab | grep -i dpi
        -DPI_VOID_TASK                          -- Return value of export and import tasks will be VOID. 
        -DPIHEADER <arg>                        -- generate the header file for export functions/tasks
        -DPIIMPHEADER <arg>                     -- generate the header file for import functions/tasks
        -GCC_VERS <arg>                         -- Pass gcc_vers switch for DPI file compilation
```

DPI libraries `lib<name>.so` can be passed only to the simulation runtime stage:

```plaintext
xmsim -sv_lib <dir>/lib<name>.so
```

Without it, `xsim` (or `xrun -R`) will raise an error `NOFDPI`:

```plaintext
xmsim: *F,NOFDPI (<file>.sv,<line>|<column>): Function <name> not found in any of the shared object specified with -SV_LIB switch. The corresponding import DPI declaration is done in file <file>.sv at line no. <line>.
```

About the raised error:

```plaintext
xmhelp xmsim NOFDPI

xmsim/NOFDPI =
        C function declared in import DPI declaration should be present in
        shared library. Specified function is not available in any of the
        provided shared objects or default "libdpi".
        (Default severity: 'F')

        See more info at: https://support.cadence.com/apex/Coveo_CommunitySearch#ph=Xcelium&t=PlatformProductPage&oMenu=Search&cadproduct=%22Xcelium%22&q=NOFDPI
```

This PR:

* Fixed handling DPI libraries
* DPI libraries are passed to the simulation runtime stage `xrun -R` (aka `xmsim`)
* Removes DPI libraries from elaboration stage. They where ignored by the `xrun` elaboration stage
* DPI libraries were composed as list of list (`dpi_lib_cmd`) and concatenated with xrun arguments generating invalid arguments at the end `['-sv_lib', 'libdpi.so']`
* Adds additional `libdpi.so` file for test fixture (and fixes legacy xcelium tests because of that...)
* Extends `test_tool_xcelium` to check if the `xrun -R` simulation command contains DPI arguments `-sv_dpi libdpi.so`

### DIFF
--- a/edalize/tools/xcelium.py
+++ b/edalize/tools/xcelium.py
@@ -99,9 +99,9 @@ class Xcelium(Edatool):
             self.tcl_cmd += ["-input", tcl]
 
         # Append DPI libraries
-        dpi_lib_cmd = []
+        self.dpi_lib_cmd = []
         for lib in dpi_libraries:
-            dpi_lib_cmd.append(["-sv_lib", lib])
+            self.dpi_lib_cmd.extend(("-sv_lib", lib))
 
         # Append macro definitions
         macro_def_cmd = []
@@ -183,7 +183,6 @@ class Xcelium(Edatool):
 
         self.xrun_f = (
             en_64bit_cmd
-            + dpi_lib_cmd
             + macro_def_cmd
             + vlogparam_cmd
             + incdir_cmd
@@ -198,7 +197,7 @@ class Xcelium(Edatool):
         self.update_config_file("xrun.f", "\n".join(self.xrun_f) + "\n")
 
     def run(self):
-        args = ["-R"] + self.tcl_cmd
+        args = ["-R"] + self.dpi_lib_cmd + self.tcl_cmd
 
         if self.tool_options.get("gui"):
             args += ["-gui"]
--- a/tests/edalize_common.py
+++ b/tests/edalize_common.py
@@ -244,6 +244,7 @@ FILES = [
     {"name": "qsf_constraint_file.qsf", "file_type": "QSF"},
     {"name": "pdc_floorplan_constraint_file.pdc", "file_type": "FPPDC"},
     {"name": "lpf_file.lpf", "file_type": "LPF"},
+    {"name": "libdpi.so", "file_type": "dpiLibrary"},
 ]
 """Files of all supported file types."""
 
--- a/tests/test_tool_xcelium.py
+++ b/tests/test_tool_xcelium.py
@@ -27,6 +27,8 @@ def test_tool_xcelium(tool_fixture):
     assert "-R" in args
     assert "-input" in args
     assert args[args.index("-input") + 1] == "tcl_file.tcl"
+    assert "-sv_lib" in args
+    assert args[args.index("-sv_lib") + 1] == "libdpi.so"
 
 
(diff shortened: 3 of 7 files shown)


olofk:
Good catch. This looks correct to me.

Another thing. I don't have much experience with Xcelium myself. Do you think it's worth calling the individual xm* commands directly rather than xrun for better control over the compile/elab/run cycles? I tend to prefer to do that instead of using wrappers for other tools that I understand a bit better but chose xrun here since it looked like the easiest option to begin with.

tymonx:
> Another thing. I don't have much experience with Xcelium myself. Do you think it's worth calling the individual xm* commands directly rather than xrun for better control over the compile/elab/run cycles? I tend to prefer to do that instead of using wrappers for other tools that I understand a bit better but chose xrun here since it looked like the easiest option to begin with.

@olofk I also prefer that. Benefits:

* Parallel compilation of libraries. Unfortunately, like any other EDA tool, they parse/compile HDL into single shared database. But at least, compiling N files into same library could be done in parallel. When Edalize will add support for more build runners, it will be more beneficial
* Incremental compilation. It is much easier to track files with Makefile/Ninja and recompile library if needed
* Compilation -> elaboration changes could be detected easier. For example, changing source file -> re-compile - re-elaborate. But changing only HDL parameter -> only re-elaborate
* And for some reason, `xrun` is happy to use invalid/incorrect options :roll_eyes: Few PRs already showed this buggy flaw. At least, using `xmvhdl`, `xmvlog`, `xmelab` and `xmsim` commands separately would raise an error or some kind of information about wrong usage

I will create an issue for that. I'm actively using FuseSoC + Edalize + Xcelium (with Xsim and Questa as well).

PR #551, opened 2026-07-05
feat(xsim): add support for the AMD/Xilinx XSim logic simulator

### LINKED ISSUE #166: Add xsim DPI support
Xilinx's Vivado simulator supports DPI, though it doesn't seem to support VPI. The flow seems to be just like the Icarus VPI flow, with the `xsc` tool used to compile the library like `iverilog-vpi` and the `xelab -sv_dpi <lib name>` switch added to point to the libraries.

It seems like adding this using the existing VPI path would be straightforward even if the name isn't quite right. Does this make sense? Have any other tools added DPI support and if so did they reuse this path as well?

tymonx:
Closes https://github.com/olofk/edalize/issues/166

This PR:

* Adds full support for  the AMD/Xilinx XSim logic simulator including DPI

To add assertions in tests for handling DPI libraries, I need this PR to be merged https://github.com/olofk/edalize/pull/549. It adds the `libdpi.so` file fixture. It also fixes tests for the `xcelium` backend. I will add these assertions in separate PR.


olofk:
Fantastic! Tried it out with SERV and it works perfectly fine, including file-specific defines. Didn't test DPI support, but trust you on that part.

olofk:
Hmm... I'm seeing some CI issues related to xsim and libdpi.so. Could you please take a look at that?

tymonx:
> Hmm... I'm seeing some CI issues related to xsim and libdpi.so. Could you please take a look at that?

@olofk  https://github.com/olofk/edalize/pull/560

Something that should not happen :thinking: How this passed anyway :thinking:

PR #554, opened 2026-07-06
feat(xcelium): add support for VHDL generics

tymonx:
This PR adds missing support for VHDL generics.

From the Cadence Xcelium `xrun` command:

```plaintext
xrun -helpall | grep -i -- -generic
  -generic <arg>             Associate value with top-level generic
```

They are already set for top level, the `self.toplevel` prefix is not needed.


olofk:
Makes sense. Thanks

Issue #470, opened 2025-08-12
Paths in Vivado's TCL commands should always contain forward slashes (also on Windows)


idgmicsei:
I am using edalize through fusesoc and trying to set up a project that can be synthesized on both Linux and Windows (11). While the Linux one works, the Windows version fails when Vivado is trying to create the project via the generated `project.tcl` because files are not found. When looking closer into the `project.tcl`, the paths contain forward slashes (`'/'`) from the files as specified in the core file but may also contain backward slashes (`'\'`) when the path is constructed for reading through Vivado.

An example from the `project.tcl` is shown here:

```
...
read_bd {..\..\..\src/bd/design_1/design_1.bd}
...
```

While `UG973` states that backward slashes are (theoretically) allowed as path delimiters on Windows, the above example does not work as the backward slashes seem to be treated as special characters (this is also stated in `UG894`) in TCL. Note that for some other files (e.g. `read_vhdl ...`) it seems to work also with backward slashes.

I have tried to fix this rudimentarily by adding the following line which converts all backward to forward slashes.

```
diff -ruN orig/edalize/tools/vivado.py new/edalize/tools/vivado.py
--- orig/edalize/tools/vivado.py    2025-08-12 15:37:12.356799587 +0200
+++ new/edalize/tools/vivado.py     2025-08-12 15:38:41.593407800 +0200
@@ -111,6 +111,7 @@
         for f in self.files:
             file_type = f.get("file_type", "")
             cmd = ""
+            f["name"] = f["name"].replace('\\', '/')
             if file_type.startswith("verilogSource"):
                 cmd = "read_verilog"
             elif file_type.startswith("systemVerilogSource"):
``` 

While this works for me I don't know if this is the best solution.
I am using `edalize-0.6.1`

PR #555, opened 2026-07-06
Fixes path issues on Windows with Vivado

dozius:
Forces forward slashes. Fixes issue #470.

### DIFF
--- a/edalize/tools/vivado.py
+++ b/edalize/tools/vivado.py
@@ -7,6 +7,7 @@ import os.path
 import platform
 import re
 import subprocess
+from pathlib import Path
 
 from edalize.tools.edatool import Edatool
 from edalize.utils import EdaCommands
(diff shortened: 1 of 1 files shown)


edalize/tools/vivado.py
```diff
@@ -109,6 +109,7 @@ def setup(self, edam):
 
         dep_files = []
         for f in self.files:
+            f["name"] = f["name"].replace("\\", "/")
```
tymonx: Not better to use the [Path.as_posix()](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix) function 🤔


```suggestion
            f["name"] = Path(f["name"]).as_posix()
```

dozius: Sure, that works too. I'd actually prefer pathlib but I thought sticking with what was being done in other places in the codebase would be more consistent. I'll update to use pathlib instead.

olofk:
Thanks. Agree that using pathlib is nicer here.

PR #557, opened 2026-07-20
feat(cocotb): fail sim flow run when one of cocotb tests failed

### LINKED ISSUE #493: bug(sim-flow): failed cocotb or failed simulation runtime is not propagate as an error by sim flow
Some HDL simulators don't return a non-zero return code (!= 0) when something failed during simulation runtime. Including failed cocotb tests or failed cocotb initialization.

[Cocotb Runner](https://github.com/cocotb/cocotb/blob/v2.0.1/src/cocotb_tools/runner.py#L170) class is handling that by inspecting generated JUnit XML file. It will fail if:
- Generated JUnit XML file `results.xml` is not present
- Generated JUnix XML file `results.xml` contains non-zero `failures` or `errors` in `<testsuite>` XML elements

Something similar could be implemented in Edalize.

References:
- [cocotb/src/cocotb_tools/runner.py](https://github.com/cocotb/cocotb/blob/v2.0.1/src/cocotb_tools/runner.py#L566)
- [cocotb/src/cocotb_tools/check_results.py](https://github.com/cocotb/cocotb/blob/v2.0.1/src/cocotb_tools/check_results.py#L13)

This should be implemented in [edalize/flows/sim.py](https://github.com/olofk/edalize/blob/main/edalize/flows/sim.py).

To reproduce it:
* Git clone https://gitlab.com/tymonx/pytest-cocotb-fusesoc repository
* Edit assertion in [tests/hdl/dff/test_dff.py](https://gitlab.com/tymonx/pytest-cocotb-fusesoc/-/blob/d71673440ee3ac1b15c7e674c1bdefa429669416/tests/hdl/dff/test_dff.py#L57) test from `assert dut.q.value == expected_val` to `assert dut.q.value != expected_val`
* Run simulation via fusesoc `pytest -s -v -rA`

Image

The `test_dff` test is invoking `fusesoc run dff` command -> [src/pytest_cocotb/fusesoc/plugin.py](https://gitlab.com/tymonx/pytest-cocotb-fusesoc/-/blob/main/src/pytest_cocotb/fusesoc/plugin.py?ref_type=heads#L30).

```python
    run(["fusesoc", "run", hdl.toplevel], check=True)
```

 It should fail not pass. The `check=True` argument will tell the `subprocess.run` function to raise an exception when invoked command returned a non-zero status code (error).

tymonx:
Fixes https://github.com/olofk/edalize/issues/493

PR is checking for generated JUnit XML file from cocotb. This can be `results.xml` or `${PYTEST_CURRENT_TEST}.result.xml`. If one of cocotb tests failed, it will raise an exception like cocotb runner is doing.

### DIFF
--- a/edalize/flows/sim.py
+++ b/edalize/flows/sim.py
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 import os
+from pathlib import Path
+from xml.etree import ElementTree as ET
 
 from edalize.flows.generic import Generic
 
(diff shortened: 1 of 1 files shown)


olofk:
This is a very nice functionality! The sw architect in me would consider running this as a separate step after the simulation is complete, but we could split it out later on if we want to.

